PR #22, opened 2020-11-04
config: fixes typo in opsgenie hartbeat config

### LINKED ISSUE #8: Opsgenie heartbeat configuration inconsistency
In `sc-config.yaml` under `alerts.opsGenieHeartbeat` there is both `enable` and `enabled`. When running with the production flavor one is set to true and the other to false.

There should only be one option for enabling the heartbeat and it should be enabled by default for production.

viktor-f:
**What this PR does / why we need it**: Part of qa testing for v0.7.0. This PR fixes a typo in the prod flavor regarding opsgenie heartbeat. Also removes some redundant config in dev flavor.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: #8 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/config/config/flavors/prod-sc.yaml
+++ b/config/config/flavors/prod-sc.yaml
@@ -8,7 +8,7 @@ dex:
 alerts:
   alertTo: "opsgenie"
   opsGenieHeartbeat:
-    enable: true
+    enabled: true
     name: #Fill this out
 
 prometheus:
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,6 +1,8 @@
 ### Release notes
 
-- Configuration for the certificate issuers has been changed and requires running the [migration script](migration/v0.6.x-v0.7.x/migrate-issuer-config.sh).
+- Configuration for the certificate issuers has been changed and requires running the [migration script](migration/v0.6.x-v0.7.x/migrate-issuer-config.sh).'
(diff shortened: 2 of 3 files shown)


lentzi90 (commented):
Looking good!
One thing we should maybe add is that if upgrading, you should remove the faulty config from `sc-config.yaml`. At least I think old clusters will have it.

config/config/flavors/prod-sc.yaml
```diff
@@ -8,7 +8,7 @@ dex:
 alerts:
   alertTo: "opsgenie"
   opsGenieHeartbeat:
-    enable: true
+    enabled: true
     name: #Fill this out
```
Ajarmar, line 12: Should this field actually be empty, instead of "set-me"?

Ajarmar (commented):
Looks good, just wondering about one thing (see comment).

config/config/flavors/prod-sc.yaml
lentzi90: Yes if we set it to "set-me" and the operator forgot to change it, the installation would run without a problem. By leaving it empty we make sure that the user notice it since the config validation will fail unless it is set. :slightly_smiling_face:

lentzi90 (approved):
LGTM!

PR #25, opened 2020-11-04
pipeline: RBAC tests failed when usernames/namespaces had quotes

### LINKED ISSUE #17: User RBAC tests fail due to quotes
The following tests give false negatives (i.e. they fail even though everything is fine) if the user admin in the config is quoted:
- get node   not authorized ❌
- get namespace      not authorized ❌
- create deployments in production   not authorized ❌
- delete deployments in production   not authorized ❌
- create deployments in staging      not authorized ❌
- delete deployments in staging      not authorized ❌
- patch configmaps/fluentd-extra-config in fluentd   not authorized ❌
- patch configmaps/fluentd-extra-plugins in fluentd  not authorized ❌

The tests fail if the user is for example `"admin@example.com"` but not for `admin@example.com`. The issue is most likely due to the parsing of the namespaces and users [in this script](https://github.com/elastisys/compliantkubernetes-apps/blob/main/pipeline/test/services/workload-cluster/testUserRbac.sh#L63).

This may not be critical enough to require for v0.7.0, but if we decide to skip it, we should make sure that this known issue is documented!

viktor-f:
**What this PR does / why we need it**: Part of qa tests for v0.7.0. Simple fix for some rbac tests.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #17 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/pipeline/test/services/workload-cluster/testUserRbac.sh
+++ b/pipeline/test/services/workload-cluster/testUserRbac.sh
@@ -59,8 +59,8 @@ echo
 echo "Testing user RBAC"
 echo "====================="
 
-user_namespaces=$(yq r "$CONFIG_FILE" 'user.namespaces' | sed 's/-//' - )
-user_admin_users=$(yq r "$CONFIG_FILE" 'user.adminUsers' | sed 's/-//' - )
+user_namespaces=$(yq r "$CONFIG_FILE" 'user.namespaces[*]')
+user_admin_users=$(yq r "$CONFIG_FILE" 'user.adminUsers[*]')
 
 for user in ${user_admin_users}; do
     testCanUserDo "get" "node" "$user"


pipeline/test/services/workload-cluster/testUserRbac.sh
```diff
@@ -59,8 +59,8 @@ echo
 echo "Testing user RBAC"
 echo "====================="
 
-user_namespaces=$(yq r "$CONFIG_FILE" 'user.namespaces' | sed 's/-//' - )
-user_admin_users=$(yq r "$CONFIG_FILE" 'user.adminUsers' | sed 's/-//' - )
+user_namespaces=$(yq r "$CONFIG_FILE" 'user.namespaces' | sed 's/-//' - | sed 's/"//g' - )
+user_admin_users=$(yq r "$CONFIG_FILE" 'user.adminUsers' | sed 's/-//' - | sed 's/"//g' - )
```
danielharr, line 63: These lines could be simplified a bit, for example
```suggestion
user_admin_users=$(yq r "$CONFIG_FILE" 'user.adminUsers' | tr -d '"-' )
```

danielharr (approved):
Looks good, added a small suggestion but I'm fine with either :slightly_smiling_face:

pipeline/test/services/workload-cluster/testUserRbac.sh
lentzi90: I wonder if we should try something like [this](https://stackoverflow.com/questions/35005893/convert-a-json-array-to-a-bash-array-of-strings) instead? Otherwise it is just a question of time until someone decides to do this:
```yaml
user:
  adminUsers: ["user1", "user2"]
```
It is perfectly fine to do in yaml, but our parsing here will fail. If we instead get the output from yq as json, we could parse the array as in the linked example.

danielharr: That sounds very much better :slightly_smiling_face:

viktor-f: Did some more testing and found that `user.adminUsers[*]` will output the values on separate lines without any `-` or `"` regardless of if it's entered as
```
user:
  adminUsers: ["user1", "user2"]
```
or

```
user:
  adminUsers: 
  - "user1"
  - "user2"
```
So now we should have safe parsing without needing to format it as json first.

danielharr (approved):

lentzi90 (approved):

Issue #32, opened 2020-11-05
Rework version validation


cristiklein:
`./bin/ck8s` performs version validation as follows:

1. It retrieves the script's version by performing `git describe --tags --abbrev=0 HEAD | sed 's/^v//'`
2. It retrieves the configuration version by looking at `global.ck8sVersion`
3. It mandates that the two be equal.

This is problematic for two reasons:
1. `compliantkubernetes-apps` might be checked out shallow or might be included as a submodule, hence its own version may be unreliable.
2. Version mismatch should be allowed according to Semantic Versioning rules.

lentzi90:
I think this is a duplicate of #10 or at least very much related.

cristiklein:
Agreed. I'm closing this one.

Issue #10, opened 2020-11-04
Config validation doesn't check what version is actually used


lentzi90:
Validation of [config](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/common.bash#L71) doesn't really check what version you use. Instead it just picks the latest tag, no matter how far back in history that is. This is very dangerous since it looks like someone running the latest bleeding edge non-release are actually on a supported tag in the config.

It would be better to clearly warn the user when not running supported release. It must still be possible to run non-tagged versions of course for development. In these cases the version could be set to the current commit sha or simply `dev`.

Bonus: Don't just check what commit is checked out, but also if the directory is "clean" (i.e. are there any non-committed changes). If there are, add `+dirty` to the version to indicate this.

Suggested fix: If no tag is available, set the version to `dev` and print a warning for the user.

tordsson:
Duplicated of #189

PR #35, opened 2020-11-06
bin: Assume 'latest' if version is unavailable

cristiklein:

To deploy Compliant Kubernetes, one would generally include this repo as
a submodule, in which case tags might be unavailable. This commit allows
for the config to have the version 'latest', essentially meaning "I, the
operator, take responsibility to ensure the version of the config and
the version of ck8s match".

Does not really fix, but is related to #32 and #10.

**Checklist:**

- [NA] Updated the [public facing documentation](https://github.com/elastisys/compliantkubernetes)
    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -10,6 +10,7 @@
 - Configurable extra role mappings in Elasticsearch
 - Added falco exporter to workload cluster
 - Falco dashboard added to Grafana
+- `any` can be used as configuration version to disabled version check
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


danielharr (commented):
I'm afraid that "latest" could be interpreted as "latest stable", which in this case would not necessarily be true.
Would it make sense to use the version name `dev` instead?

Latest commit on `main` comes with few guarantees (reviewed, but does not need to pass e2e tests), since we merge PRs directly to main at the moment.
I would be more comfortable providing a `latest` version if that at least meant "latest commit to pass e2e tests" :slightly_smiling_face:

cristiklein:
Well, we could also say something like `any`. Maybe that would make it clearer that version check is disabled.

danielharr (approved):
LGTM :slightly_smiling_face:

OlleLarsson (approved):

PR #40, opened 2020-11-09
apps: Don't add falco repo for sc

OlleLarsson:
Just a minor cleanup in the apps helmfile.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/helmfile/01-applications.yaml
+++ b/helmfile/01-applications.yaml
@@ -8,8 +8,6 @@ repositories:
   url: https://kiwigrid.github.io
 - name: vmware-tanzu
   url: https://vmware-tanzu.github.io/helm-charts
-- name: falcosecurity
-  url: https://falcosecurity.github.io/charts
 - name: ingress-nginx
   url: https://kubernetes.github.io/ingress-nginx
 


danielharr (approved):

geoff-elastisys:
I think there is a good argument to deploy falco in the sc.  So maybe rather thanremove the repo we should look at deploying it instead?

danielharr:
> I think there is a good argument to deploy falco in the sc. So maybe rather thanremove the repo we should look at deploying it instead?

Since it's such a small change I'm fine with removing it for now and adding it once we're starting to run falco in the sc.
Especially since we have the repository listed twice at the moment (one for both sc and wc, and one for only wc) :slightly_smiling_face:

geoff-elastisys:
> > I think there is a good argument to deploy falco in the sc. So maybe rather thanremove the repo we should look at deploying it instead?
> 
> Since it's such a small change I'm fine with removing it for now and adding it once we're starting to run falco in the sc.
> Especially since we have the repository listed twice at the moment (one for both sc and wc, and one for only wc)

Fair point

geoff-elastisys (approved):

PR #51, opened 2020-11-10
bin: Warn if config options are missing

### LINKED ISSUE #59: Relax configuration validation
To allow for better configuration options we need to temprarily relax our current configration validation until we have time to improve it.
One solution could be to warn the user that there might be something wrong in their confiuration and give them the option to abort and double check their configuration.

OlleLarsson:
One possible solution on how we can relax the configuration validation.

fixes #59 

I am not sure if it is smart to read user input or not. Might not be so good for the pipeline.

This is what it looks like right now:
```
[ck8s] Validating wc config
[ck8s] WARN: fluentd.nodeSelector."node-role.kubernetes.io/master" is not set in /home/user/.ck8s/cloud_p/env_name/wc-config.yaml
[ck8s] Do you want to exit? (y/n): y
```

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/bin/common.bash
+++ b/bin/common.bash
@@ -6,6 +6,8 @@
 
 : "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
 
+CK8S_AUTO_APPROVE=${CK8S_AUTO_APPROVE:-"false"}
+
 if [ "${SOPS_PGP_FP+x}" != "" ]; then
     echo "SOPS_PGP_FP is deprecated." >&2
     echo "Set CK8S_PGP_UID or CK8S_PGP_FP and run \`ck8s init\` to" \
@@ -88,7 +90,7 @@ validate_version() {
     ck8s_version=$(yq r "${file}" 'global.ck8sVersion')
     if [[ -z "$ck8s_version" ]]; then
         log_error "ERROR: global.ck8sVersion is not set in ${file}"
-        will_exit="true"
+        exit 1
     fi
     if [ "${ck8s_version}" != "any" ] \
         && [ "${version}" != "${ck8s_version}" ]; then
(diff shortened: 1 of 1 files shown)


bin/common.bash
```diff
@@ -208,7 +212,7 @@ config_load() {
       exit 1
     fi
 
-    validate_version "$1"
+    #validate_version "$1"
```
danielharr, line 215: Should this be commented out?

bin/common.bash
```diff
@@ -88,7 +88,6 @@ validate_version() {
     ck8s_version=$(yq r "${file}" 'global.ck8sVersion')
     if [[ -z "$ck8s_version" ]]; then
         log_error "ERROR: global.ck8sVersion is not set in ${file}"
-        will_exit="true"
```
danielharr, line 91: Should the version variable not be required?

danielharr (commented):
Not sure that I'm a fan of the input requirement :thinking: 
Looks good in general though, just a few questions :slightly_smiling_face:

bin/common.bash
OlleLarsson: Nope

bin/common.bash
OlleLarsson: Oh, right, it should be changed to `maybe_exit` instead.

danielharr: Yeah, we should probably try to declare more variables as `local` to reduce their scope :slightly_smiling_face:

bin/common.bash
```diff
@@ -136,18 +136,23 @@ validate_config() {
...
-                will_exit="true"
+                log_warning "WARN: ${opt} is not set in ${file}"
+                maybe_exit="true"
             fi
         done
-        if [[ "${will_exit}" == "true" ]]; then
-            exit 1
+
+        if [[ "${maybe_exit}" == "true" ]]; then
+            echo -n -e "[\e[34mck8s\e[0m] Do you want to exit? (y/n): " 1>&2
```
danielharr, line 151: If we're going to keep the input requirement, maybe use the word "abort" or rephrase to "Continue anyway?" instead of "exit"?

viktor-f (approved):
LGTM

danielharr (approved):
LGTM

PR #55, opened 2020-11-10
Support for CA certificate bundles

danielharr:
This is a proposal for adding support to trust custom CA certificate bundles.
Feedback, thoughts and suggestions are appreciated! :slightly_smiling_face: 

This proposal assumes that the CA bundle is already present in the cluster as a Configmap or Secret in the namespaces where required.
I guess that the CA bundle installation will be specific for each case where this support is needed, and therefore left it as a manual task for now.
There is however a script available for testing purposes which I do not intend to keep for the final solution.

Summarized, this solution does two things:
* Exposes configuration to mount Secrets or Configmaps, depending on application.
* Exposes configuration to trust a CA certificate in the applications that do not necessarily trust all bundles in `/etc/ssl/certs/`

To test this, I configured a compliantkubernetes environment to use the letsencrypt-staging issuer and `global.verifyTls: true`.
Then I verified that the applications did not trust each other until the letsencrypt staging root certificate was mounted:

* Prometheus wc scraper - we have tests for this, which passed after mounting the certificate. Also port-forwarded to instance and checked it's target list
* Fluentd: stopped logging TLS error messages after certificate was mounted and configured in the Helm chart
* Grafana: logging in using dex static login started working after mounting certificate
* Harbor: logging in using dex static login (also tested OIDC) started working after mounting certificate and configuring Helm chart

This PR does not take internal cluster traffic into consideration, with an exception to when the public ingress is used.

Part of "bare metal" support

The config values that are commented out in the compliantkubernetes config files are commented out by intention.
The Harbor chart renders the value even if the string is empty.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.


viktor-f:
Some questions:

Have you tested starting a pod with an image from Harbor? I'm guessing that will not work with bad certificates and if that's the case then it would be nice to see if we can make kubernetes trust it or at least be able to give our customers some instructions on how to configure their pods.

You say that the script is only for testing. But I think that it would be nice with this kind of script that takes a ca-bundle and populates the necessary configmaps/secrets with it. It would also be nice with a script that populates the config options with values that work with the first script.

danielharr:
> Have you tested starting a pod with an image from Harbor? I'm guessing that will not work with bad certificates and if that's the case then it would be nice to see if we can make kubernetes trust it or at least be able to give our customers some instructions on how to configure their pods.

No, I have not. That would require some work in the details for how we install and configure Kubernetes, which I am not up to date with.
It could be an idea to make sure support for that is implemented and test both this and that feature simultaneously :slightly_smiling_face: 

> You say that the script is only for testing. But I think that it would be nice with this kind of script that takes a ca-bundle and populates the necessary configmaps/secrets with it. It would also be nice with a script that populates the config options with values that work with the first script.

I agree, but I'm not sure if it should be kept with this repository or somewhere else :thinking: 
Maybe keep it with some kind of on-boarding kit, which of course could be part of this repository, but it is not something we currently have.

viktor-f:
> No, I have not. That would require some work in the details for how we install and configure Kubernetes, which I am not up to date with.
> It could be an idea to make sure support for that is implemented and test both this and that feature simultaneously slightly_smiling_face 

Your right, it does not really belong here. I think it's fine to skip that for now.
 
> I agree, but I'm not sure if it should be kept with this repository or somewhere else thinking
> Maybe keep it with some kind of on-boarding kit, which of course could be part of this repository, but it is not something we currently have.

I think it could be placed in this repo. The script that adds bundles to configmaps/secrets could be a part of the bootstrap scripts.

danielharr:
> I think it could be placed in this repo. The script that adds bundles to configmaps/secrets could be a part of the bootstrap scripts.

I'm missing a structure for managing external files that should be "installed" to compliantkubernetes though.
I guess we could pass an absolute path to the CA file, but I would prefer something relative to the environment configuration.
Something like `$CK8S_CONFIG_PATH/addons/certs/`? Not optimal, but some interface for stuff that's optional and not core compliantkubernetes :slightly_smiling_face:

viktor-f:
> I'm missing a structure for managing external files that should be "installed" to compliantkubernetes though.
> I guess we could pass an absolute path to the CA file, but I would prefer something relative to the environment configuration.
> Something like `$CK8S_CONFIG_PATH/addons/certs/`? Not optimal, but some interface for stuff that's optional and not core compliantkubernetes slightly_smiling_face

Yes, something like that would be good. A folder in the config path sounds like a good enough solution.

OlleLarsson:
I'm about to start looking at this PR, but I have an immediate question and that is if you have checked opendistro?
I would assume that we'd need to add some OIDC configuration paramater in Kibana at the very least https://docs.search-guard.com/latest/kibana-authentication-openid#configuration for it to trust dex.

helmfile/charts/init-harbor/templates/init-harbor-job.yaml
```diff
@@ -2,6 +2,9 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   name: {{ .Release.Name }}-job
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
```
OlleLarsson, line 6: I can't quite remember what the init job does but have you checked that it handles being run multiple times correctly?

config/config/sc-config.yaml
```diff
@@ -34,6 +34,7 @@ user:
     tolerations: []
     affinity: {}
     nodeSelector: {}
+    extraConfigmapMounts: []
```
OlleLarsson, line 37: I'd actually think it'd be great if these had an example of how to mount the ca-certificate.

OlleLarsson (commented):
Nice work. I've left a couple of comments.

Xartos:
@danielharr Is this still relevant so we should continue on it or should we close this

danielharr:
> @danielharr Is this still relevant so we should continue on it or should we close this

@Xartos I don't think we decided on if we should pursue this or not, and considering the age of it, let's close :slightly_smiling_face:

PR #91, opened 2020-11-20
config: alertmanager namespace is needed in sc config

viktor-f:
**What this PR does / why we need it**: A miss in the config when the alertmanager namespace was parameterized.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -37,6 +37,7 @@ user:
   # Todo remove dependencie on alertmanager from service cluster
   alertmanager:
     enabled: false
+    namespace: monitoring
     ingress:
       enabled: false
 


config/config/sc-config.yaml
```diff
@@ -37,6 +37,7 @@ user:
   # Todo remove dependencie on alertmanager from service cluster
   alertmanager:
     enabled: false
+    namespace: monitoring
```
OlleLarsson, line 40: So this is the namespace where the user-alertmanager runs?

viktor-f: No, I guess that this is where the operator alertmanager runs in the service cluster. I think this was used by the wc alerts that we apply in the service cluster

OlleLarsson: Okay, sure I did not think about that.

OlleLarsson (approved):

danielharr (approved):

PR #96, opened 2020-11-26
Remove post-install script and dependency on infra.json

### LINKED ISSUE #129: [1] Custom NFS solutions
We need to support custom NFS solutions, where the server path may be different from `/nfs` and we may be running on another cloud provider than Exoscale.
To accomplish this we will need to change [when we deploy the nfs-client provisioner](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bootstrap/storageclass/bootstrap.sh#L23) and allow for [other values](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bootstrap/storageclass/helmfile/values/nfs-client-provisioner.yaml.gotmpl#L11) to be passed to the chart.

Definition of done: I can use the `nfs-client` storage class on any supported cloud provider and I can configure both IP and path for it (separately for SC and WC).

ewnetu:
This PR removes `scripts/post-infra-common.sh` and its dependency. The  NFS_SC_SERVER_IP and NFS_WC_SERVER_IP environment variables, which were in  `scripts/post-infra-common.sh`, are now automatically added to `sc-config.yaml` and ` wc-config.yaml` during initialization.

fixes #https://github.com/elastisys/ck8s-apps/issues/241
fixes #129 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.


bin/apps.bash
```diff
@@ -19,15 +19,6 @@ apps_init() {
...
 
-    # TODO: We should try to get rid of the post-infra-common script.
-
-    log_info "Running post infra script"
-    : "${scripts_path:?Missing scripts path}"
-    : "${config[infrastructure_file]:?Missing infrastructure file}"
-    # shellcheck disable=SC1090
-    source "${scripts_path}/post-infra-common.sh" \
-        "${config[infrastructure_file]}"
-}
```
danielharr, line 30: This closing bracket should probably not be removed :slightly_smiling_face:

bin/bootstrap.bash
```diff
@@ -8,20 +8,12 @@ set -eu -o pipefail
...
 # shellcheck disable=SC1090
 source "${here}/common.bash"
-: "${config[infrastructure_file]:?Missing infrastructure file}"
 : "${secrets[kube_config_sc]:?Missing service cluster kubeconfig}"
 : "${secrets[kube_config_wc]:?Missing workload cluster kubeconfig}"
 export bootstrap_path="${here}/../bootstrap"
 export scripts_path
 export common_path="${here}/common.bash"
 
-bootstrap_init() {
```
danielharr, line 18: The references to this function has to be removed as well :slightly_smiling_face:

bin/init.bash
```diff
@@ -82,6 +82,15 @@ generate_base_sc_config() {
     if [[ -f $file ]]; then
         yq merge "$tmpfile" "$file" --inplace -a=overwrite --overwrite --prettyPrint
     fi
+
+    cloud_provider=$(yq r -e "$tmpfile" 'global.cloudProvider')
```
danielharr, line 86: This file has `CK8S_CLOUD_PROVIDER` as a requirement, so you could use that variable instead of creating this new one

bin/init.bash
```diff
@@ -99,7 +108,19 @@ generate_base_wc_config() {
...
+    storageClass=$(yq r -e "$tmpfile" 'global.storageClass')
+    if [ "$cloud_provider" == "exoscale" &&  "storageClass" == "nfs" ]
```
danielharr, line 114: I see that this check is different than for service cluster - should it?
Would it be possible to create a function for setting the nfs ip instead, and call it for both sc and wc? :slightly_smiling_face:

bin/init.bash
```diff
@@ -82,6 +82,15 @@ generate_base_sc_config() {
     if [[ -f $file ]]; then
         yq merge "$tmpfile" "$file" --inplace -a=overwrite --overwrite --prettyPrint
     fi
+
+    cloud_provider=$(yq r -e "$tmpfile" 'global.cloudProvider')
+    if [ "$cloud_provider" == "exoscale" ]
+    then
+      NFS_SC_SERVER_IP=$( jq -r '.workload_cluster.nfs_ip_addresses.nfs.private_ip' < "${config[infrastructure_file]}")
+      export  NFS_SC_SERVER_IP
```
danielharr, line 90: Does this environment variable have to be exported, if we have it in the helmfile values now? :slightly_smiling_face: 
Same for where the wc config is set up.

helmfile/values/prometheus-operator-sc.yaml.gotmpl
```diff
@@ -290,7 +290,7 @@ prometheus:
       scheme: http
       static_configs:
       - targets:
-        - '{{ requiredEnv "NFS_SC_SERVER_IP" }}:9100'
+        - '{{ .Values.global.nfs_server_ip }}:9100'
```
danielharr, line 293: This change (and for wc) should have removed the requirement of exporting the nfs ip environment variable :slightly_smiling_face:

danielharr (commented):
Really nice! :tada: 
I have added some questions :slightly_smiling_face:

bin/init.bash
ewnetu: > Does this environment variable have to be exported, if we have it in the helmfile values now? 
> Same for where the wc config is set up.

There is a reference for the environment variables in `bootstrap/storagclass/helmfile/helmfile.yaml ` as shown below. I couldn't find a better way to it.  
```
{{ if eq .Environment.Name "workload_cluster" }}
      server: {{ requiredEnv "NFS_WC_SERVER_IP" }}
{{ end }}
```

danielharr: Would it be possible to remove the reference in the helmfile by setting that value (`nfs.server`) in the `bootstrap/storageclass/helmfile/values/nfs-client-provisioner.yaml.gotmpl` file instead?

```
# nfs-client-provisioner.yaml.gotmpl
...
nfs:
  path: "/nfs"
  server: "{{ .Values.global.nfs_server_ip }}"
```

bootstrap/storageclass/helmfile/values/nfs-client-provisioner.yaml.gotmpl
```diff
@@ -9,4 +9,9 @@ serviceAccount:
   name: nfs-client-provisioner
 nfs:
   path: "/nfs"
-  #server: Will be set outside
+{{ if eq .Environment.Name "service_cluster" }}
+  server: {{ .Values.global.nfs_server_ip }}
+{{ end }}
+{{ if eq .Environment.Name "workload_cluster" }}
+  server: {{ .Values.global.nfs_server_ip }}
```
danielharr, line 16: Are the if-cases necessary? The value is `server: {{ .Values.global.nfs_server_ip }}` for both, right? :slightly_smiling_face:

bin/init.bash
```diff
@@ -82,7 +95,12 @@ generate_base_sc_config() {
     if [[ -f $file ]]; then
         yq merge "$tmpfile" "$file" --inplace -a=overwrite --overwrite --prettyPrint
     fi
+
     cat "$tmpfile" > "$file"
+
+    add_nfs_server_ip "$file" '.service_cluster.nfs_ip_addresses.nfs.private_ip'
+
+
```
danielharr, line 103: Maybe remove at least the last two blank lines here?

bin/init.bash
```diff
@@ -99,7 +117,13 @@ generate_base_wc_config() {
     if [[ -f $file ]]; then
         yq merge "$tmpfile" "$file" --inplace -a=overwrite --overwrite --prettyPrint
     fi
+
     cat "$tmpfile" > "$file"
+
+      add_nfs_server_ip "$file" '.workload_cluster.nfs_ip_addresses.nfs.private_ip'
+
+
+
```
danielharr, line 126: Maybe remove at least the last 3 blank lines, and reindent the `add_nfs_server_ip` line?

danielharr (commented):
Great work! :smile: 
A bit of nit picking, added some comments.

It looks like the environment variables are set when running `helmfile lint` in `pipelint/lint.bash`. Could you remove remove the env variables there as well?

Xartos:
Correct me if I'm wrong, but isn't the mission of the issue to move all config from infra.json to wc-config.yaml/sc-config.yaml?

danielharr:
> Correct me if I'm wrong, but isn't the mission of the issue to move all config from infra.json to wc-config.yaml/sc-config.yaml?

I understood the issue as reading the config from infra.json at init, but after that only use the value in wc/sc-config.yaml.
Not sure how we would get rid of infra.json entirely, but it would be really nice if we could :slightly_smiling_face:

ewnetu:
> > Correct me if I'm wrong, but isn't the mission of the issue to move all config from infra.json to wc-config.yaml/sc-config.yaml?
> 
> I understood the issue as reading the config from infra.json at init, but after that only use the value in wc/sc-config.yaml.
> Not sure how we would get rid of infra.json entirely, but it would be really nice if we could

Yes, we need to get the ips during init and as of now that is only from  infra.json.

bin/init.bash
```diff
@@ -65,6 +65,19 @@ replace_set_me(){
...
 }
 
+add_nfs_server_ip()
+{
+  file=$1
+  ip=$2
+
+  if [  "$CK8S_CLOUD_PROVIDER" == "exoscale"  ]
+  then
+    NFS_SERVER_IP=$( jq -r $ip < "${config[infrastructure_file]}")
```
danielharr, line 75: Would it make sense to use a more hardcoded path to the infrastructure file here, so we can get rid of `config[infrastructure_file]` from the variables altogether?

Xartos:
> Yes, we need to get the ips during init and as of now that is only from infra.json.

Sure, but if we anyway are going to poke that code I would say that we should move it to the regular config files instead. And then you would only need to add those IPs manually (or via the wrapper repo help scripts) when you deploy to ecoscale

ewnetu:
> > Yes, we need to get the ips during init and as of now that is only from infra.json.
> 
> Sure, but if we anyway are going to poke that code I would say that we should move it to the regular config files instead. And then you would only need to add those IPs manually (or via the wrapper repo help scripts) when you deploy to exoscale

I guess the trade-off here is between setting IPs manually vs automatically. Setting it manually will reduce the code complexity at the expense of manual intervention. Users may get confused where to get the IPs as well. However, I am not sure which one would be best.

Xartos:
> > > Yes, we need to get the ips during init and as of now that is only from infra.json.
> > 
> > 
> > Sure, but if we anyway are going to poke that code I would say that we should move it to the regular config files instead. And then you would only need to add those IPs manually (or via the wrapper repo help scripts) when you deploy to exoscale
> 
> I guess the trade-off here is between setting IPs manually vs automatically. Setting it manually will reduce the code complexity at the expense of manual intervention. Users may get confused where to get the IPs as well. However, I am not sure which one would be best.

Sure, but if we have clear documentation in the ck8s-cluster repo and add maybe add a helper script in this repo (or the wrapper repo) that you can run to populate these numbers I would say that it would reduce confusion alot.
I think I would be very confused if I tried to use a project with one file that is only used sometimes and serves the same purpose as the config files, just that it's positioned at another location.

bin/init.bash
vanneback: My idea is to completely remove infra.json. And when needed by nfs you manually add the ips in to the config. Doesnt that make sense? Is infra.json used anywere else?

bootstrap/storageclass/helmfile/values/nfs-client-provisioner.yaml.gotmpl
```diff
@@ -9,4 +9,4 @@ serviceAccount:
   name: nfs-client-provisioner
 nfs:
   path: "/nfs"
-  #server: Will be set outside
+  server: {{ .Values.global.nfs_server_ip }}
```
vanneback, line 12: Thins should probably be moved to its own topic in the config. So something like values.storageClass.serverIp. This to support other specific config in the storage class aswell. Im not sure about the logic of how to init / validate it though when having different settings depending on what storageClass we are using. Maybe look at the objectStorage solution for insperation

vanneback (changes requested):
You should need to change in the config.yaml files aswell. This shouldnt work without adding it there?
Like my comment below i think it makes sense to have the storageClass as its own topic in the config now rather than having it under "global"

ewnetu:
> You should need to change in the config.yaml files aswell. This shouldnt work without adding it there?
> Like my comment below i think it makes sense to have the storageClass as its own topic in the config now rather than having it under "global"

Yes, it is already added in both `sc-config.yaml` and `wc-config.yaml`. The line  `yq  w  -i "$file" global.nfs_server_ip "$NFS_SERVER_IP"` in init.bash does that.

bootstrap/storageclass/helmfile/values/nfs-client-provisioner.yaml.gotmpl
ewnetu: > Thins should probably be moved to its own topic in the config. So something like values.storageClass.serverIp. This to support other specific config in the storage class aswell. Im not sure about the logic of how to init / validate it though when having different settings depending on what storageClass we are using. Maybe look at the objectStorage solution for insperation

Wouldn't it confuse users if we the same value at different places, specially if it is set manually?

ewnetu:
> > > > Yes, we need to get the ips during init and as of now that is only from infra.json.
> > > 
> > > 
> > > Sure, but if we anyway are going to poke that code I would say that we should move it to the regular config files instead. And then you would only need to add those IPs manually (or via the wrapper repo help scripts) when you deploy to exoscale
> > 
> > 
> > I guess the trade-off here is between setting IPs manually vs automatically. Setting it manually will reduce the code complexity at the expense of manual intervention. Users may get confused where to get the IPs as well. However, I am not sure which one would be best.
> 
> Sure, but if we have clear documentation in the ck8s-cluster repo and add maybe add a helper script in this repo (or the wrapper repo) that you can run to populate these numbers I would say that it would reduce confusion alot.
> I think I would be very confused if I tried to use a project with one file that is only used sometimes and serves the same purpose as the config files, just that it's positioned at another location.

I see your point. Users need somehow to get hold of the nfs IPs, how can we get the IPs if we remove infra.json? Any idea?

Xartos:
> I see your point. Users need somehow to get hold of the nfs IPs, how can we get the IPs if we remove infra.json? Any idea?

For clusters set up with ck8s-cluster you can run `ck8s internal terraform output` and you'll get a bunch of stuff, and there should be two variables `sc_nfs_ips` and `wc_nfs_ips`

bin/init.bash
danielharr: Sure, it makes sense! I just didn't understand that the purpose was to completely remove the file when I started reviewing :slightly_smiling_face:

vanneback:
> > You should need to change in the config.yaml files aswell. This shouldnt work without adding it there?
> > Like my comment below i think it makes sense to have the storageClass as its own topic in the config now rather than having it under "global"
> 
> Yes, it is already added in both `sc-config.yaml` and `wc-config.yaml`. The line `yq w -i "$file" global.nfs_server_ip "$NFS_SERVER_IP"` in init.bash does that.

It has to be added to the files aswell. Not only from init. The files `sc-config.yaml` and `wc-config.yaml` is the only source of truth to what config values should be available for your own config files. Thats why it should always be present in those files even before the init.bash is run. 

I also think now when adding more parameters to the storageClass (like the nfs_ip) it should be its own topic. And then it should validate that if storageclass = NFS then storageClass.nfsIP is set. (this however goes into the other issues about validation aswell). So either
```yaml
global:
  globalStuff:

storageClass: nfs
  nfsServerIp: ip
```
or

```yaml
global:
  storageClass: nfs
    nfsServerIp: ip
```

bin/init.bash
```diff
@@ -65,6 +65,18 @@ replace_set_me(){
...
 }
 
+add_nfs_server_ip()
+{
+  file=$1
+  ip=$2
+
+  if [  "$CK8S_CLOUD_PROVIDER" == "exoscale"  ]
+  then
+    yq  w  -i "$file" global.nfs_server_ip '"set-me"'
```
vanneback, line 75: We are using camel case in the config so please change to `nfsServerIp`

lentzi90 (changes requested):
Nice to get rid of this finally!

Why was the whitelist test removed? I don't understand what it has to do with the rest of this PR.
Please consider using proper commit messages even if this is a draft, and keep the commit history clean. It is much easier to review that way and you will get a faster approval when you don't have to go back and change things later.

bin/init.bash
```diff
@@ -114,28 +116,42 @@ set_storage_class() {
...
 
         baremetal)
-	  storage_class=node-local
-	  es_storage_class=node-local
-	  ;;
-
+          storage_class=node-local
+          es_storage_class=node-local
+          # Correct?
+          yq write --inplace "$file" 'storageClasses.local.enabled' true
```
lentzi90, line 146: Maybe check if it is already set so we don't overwrite it by mistake?

bootstrap/storageclass/helmfile/values/nfs-client-provisioner.yaml.gotmpl
```diff
@@ -1,12 +1,14 @@
-resources: {{- toYaml .Values.nfsProvisioner.resources | nindent 2  }}
+resources:    {{- toYaml .Values.nfsProvisioner.resources | nindent 2  }}
 nodeSelector: {{- toYaml .Values.nfsProvisioner.nodeSelector | nindent 2 }}
-affinity: {{- toYaml .Values.nfsProvisioner.affinity | nindent 2 }}
-tolerations: {{- toYaml .Values.nfsProvisioner.tolerations | nindent 2 }}
+affinity:     {{- toYaml .Values.nfsProvisioner.affinity | nindent 2 }}
+tolerations:  {{- toYaml .Values.nfsProvisioner.tolerations | nindent 2 }}
 
 storageClass:
   defaultClass: true
```
lentzi90, line 7: This should be set based on config!

lentzi90 (commented):
Very nice job!
I agree with the comment about simplifying the bootstrap logic by making helm chart(s) of all storage classes.
We will still need some migration steps for this change, right? At least something about the change in the config for the storage classes.

bootstrap/storageclass/helmfile/values/nfs-client-provisioner.yaml.gotmpl
OlleLarsson: Yes I agree, i decided to skip it since only nfs has this option(as the other classes are just plain manifests). But sure we can do it for nfs right away.

OlleLarsson:
Shall I squash to one commit?

config/config/sc-config.yaml
```diff
@@ -6,9 +6,23 @@ global:
...
   clusterDns: "10.43.0.10"
 
+storageClasses:
+  # Name of the default storageclass.
+  # Normally one of 'nfs-client', 'cinder-storage', 'local-storage', 'ebs-gp2'.
+  default: "set-me"
+
+  nfs:
+    enabled: null # true | false
+    isDefaultClass: true
```
viktor-f, line 18: This feels redundant, se my comment above about how we could remove it.

config/config/wc-config.yaml
```diff
@@ -6,9 +6,23 @@ global:
...
   clusterDns: "10.43.0.10"
 
+storageClasses:
+  # Name of the default storageclass.
+  # Normally one of 'nfs-client', 'cinder-storage', 'local-storage', 'ebs-gp2'.
+  default: "set-me"
+
+  nfs:
+    enabled: null # true | false
+    isDefaultClass: true
```
viktor-f, line 18: Same here

viktor-f (changes requested):
Nice to get rid of the dependancy on `infra.json`. :)

One comment about the config

bootstrap/storageclass/helmfile/values/nfs-client-provisioner.yaml.gotmpl
```diff
@@ -1,12 +1,14 @@
...
+resources:    {{- toYaml .Values.nfsProvisioner.resources | nindent 2  }}
 nodeSelector: {{- toYaml .Values.nfsProvisioner.nodeSelector | nindent 2 }}
-affinity: {{- toYaml .Values.nfsProvisioner.affinity | nindent 2 }}
-tolerations: {{- toYaml .Values.nfsProvisioner.tolerations | nindent 2 }}
+affinity:     {{- toYaml .Values.nfsProvisioner.affinity | nindent 2 }}
+tolerations:  {{- toYaml .Values.nfsProvisioner.tolerations | nindent 2 }}
 
 storageClass:
-  defaultClass: true
+  defaultClass: {{ .Values.storageClasses.nfs.isDefaultClass }}
```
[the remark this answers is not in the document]
OlleLarsson: Perhaps `storageClasses.default` isn't a very good name since this storageclass is used for persistent volumes in our helm charts, and so it's does not have to be a "default" storageclass in kubernetes.
You could imagine a scenario where you'd want to use something like `storageClasses.default: gcp-sd` but still set the nfs storageclass as default in kubernetes.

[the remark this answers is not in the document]
viktor-f: Maybe, if you want to have two variables, one for our helm charts and one to set default in kubernetes, then that's ok with me.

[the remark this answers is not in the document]
OlleLarsson: Yeah, I'm not sure. Maybe we could leave it the way it is right now and schedule a larger discussion on how we want to handle storageClasses going forward?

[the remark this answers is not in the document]
viktor-f: Sounds good. I still think you can remove `storageClasses.nfs.isDefaultClass`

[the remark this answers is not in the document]
OlleLarsson: Okay, I don't agree, but sure.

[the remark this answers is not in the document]
Xartos: Shouldn't the one set as default just be default in kubernetes and then if a chart wants to use a specific storage class you set it to something else than the default?

E.g.
The kubernetes default is `nfs-client` but we want to use `gcp-sd` for all our applications. Then we should set all our applications to choose a specific one (not use the default storageclass).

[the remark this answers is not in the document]
OlleLarsson: I mean Yes.
As we discovered and discussed last week, your not forced to just one default storageclass in kubernetes, so there is nothing inherently with having storageClasses.*.isDefafultClass.
And also this `storageClasses.default` was just a simple solution for us to select a storageclass to use use in some of our helm charts, like kube-prometheus-stack.

viktor-f (approved):
LGTM!

OlleLarsson (approved):

lentzi90 (approved):

PR #113, opened 2020-12-03
influxdb: Fixed error when not using s3 as objectstorage

### LINKED ISSUE #112: [1] Helmfile fails if not using s3 as object storage
When deploying a cluster and not using s3 it now fails.

## How to replicate

* Create a cluster and set `objectStorage.type` to `gcs`
* Apply apps

Xartos:

Bug fix for clusters not working without s3 object storage.

fixes #112 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/helmfile/values/influxdb.yaml.gotmpl
+++ b/helmfile/values/influxdb.yaml.gotmpl
@@ -147,7 +147,9 @@ backupRetention:
   resources: {{- toYaml .Values.influxDB.backupRetention.resources | nindent 4 }}
   schedule: {{ .Values.influxDB.backupRetention.schedule | quote}}
   startingDeadlineSeconds: {{ .Values.influxDB.backupRetention.startingDeadlineSeconds }}
+  {{- if eq .Values.objectStorage.type "s3" }}
   s3:
     credentialsSecret: influxdb-backup-secret #TODO: use influxdb.fullname from _helpers.tpl instead hardcoded "influxdb"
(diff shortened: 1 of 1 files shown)


OlleLarsson (approved):
LGTM
Maybe change commit prefix to `apps sc` instead of influx?
But yeah i sort agree that it might make sense to add commit prefixes for application aswell :)

viktor-f:
I don't see any other data for gcs in the backup retention. Does the retention work on gcs?

Xartos:
> I don't see any other data for gcs in the backup retention. Does the retention work on gcs?

No, I think we said that we should not put any time for this to support GCS since it's basically not used yet. And fix it when we need to.
So there's no support for gcs backup retention. This only fixes so that you can actually apply apps even if you disabled the retention

viktor-f:
> No, I think we said that we should not put any time for this to support GCS since it's basically not used yet. And fix it when we need to.
> So there's no support for gcs backup retention. This only fixes so that you can actually apply apps even if you disabled the retention

Ok, might be worth to have that as a comment somewhere. Otherwise this is ok, will approve for now.

viktor-f (approved):

OlleLarsson:
PRs as documentation :thinking: :smile: ?
https://github.com/elastisys/compliantkubernetes-apps/pull/88#pullrequestreview-539749547

PR #114, opened 2020-12-03
fluentd: Added environment name as a record for all logs

### LINKED ISSUE #93: [2] Add cluster indentifier to logs
For multi-tenancy, we need to be able to associate each log entry with a specific workload cluster.
Part of #58

Xartos:

To support multi cluster we need to be able to differentiate between all logs in elasticsearch

fixes #93 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -39,6 +39,7 @@ The following options has been removed or replaced
 - `elasticsearch.retention.kubernetesAge` replaced by `elasticsearch.retention.kubernetesAgeDays`
 - `elasticsearch.retention.otherSize` replaced by `elasticsearch.retention.otherSizeGB`
 - `elasticsearch.retention.otherAge` replaced by `elasticsearch.retention.otherAgeDays`
+- Removed unused config `global.environmentName` and added `global.clusterName` to migrate there's [this script](migration/v0.7.x-v0.8.x/migrate-config.sh)
 
 ### Added
 
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -54,6 +54,7 @@ The following options has been removed or replaced
 - Backup retention for InfluxDB.
 - Add Okta as option for OIDC provider
 - Dex configuration to accept groups from Okta as an OIDC provider
+- Added record `cluster.name` in all logs to elasticsearch that matches the cluster setting `global.clusterName`
 
 ### Changed
 
--- a/helmfile/values/fluentd-sc.yaml.gotmpl
+++ b/helmfile/values/fluentd-sc.yaml.gotmpl
@@ -146,6 +146,13 @@ extraConfigMaps:
       </metric>
     </filter>
 
+    <filter **>
+      @type record_transformer
+      <record>
+        cluster.name "{{ .Values.global.clusterName }}"
+      </record>
+    </filter>
+
   output.conf: |-
     <match **>
       @id output-forwarding
--- a/bin/user-kubeconfig.bash
+++ b/bin/user-kubeconfig.bash
@@ -20,12 +20,10 @@ get_user_server() {
 
 log_info "Creating kubeconfig for the user"
 
-environment_name=$(yq r "${config[config_file_wc]}" 'global.environmentName')
-cloud_provider=$(yq r "${config[config_file_wc]}" 'global.cloudProvider')
+cluster_name=$(yq r "${config[config_file_wc]}" 'global.clusterName')
 base_domain=$(yq r "${config[config_file_wc]}" 'global.baseDomain')
 
 # Get server and certificate from the admin kubeconfig
-cluster_name="${environment_name}_${cloud_provider}"
 user_server=$(get_user_server)
 user_certificate_authority=/tmp/user-authority.pem
 append_trap "rm ${user_certificate_authority}" EXIT
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -1,7 +1,7 @@
 global:
   ck8sVersion: ${CK8S_VERSION}
   cloudProvider: ${CK8S_CLOUD_PROVIDER}
-  environmentName: ${CK8S_ENVIRONMENT_NAME}
+  clusterName: ${CK8S_ENVIRONMENT_NAME}-sc
   dnsPrefix: ${CK8S_ENVIRONMENT_NAME}
   baseDomain: "set-me"
   opsDomain: "set-me"
--- a/config/config/wc-config.yaml
+++ b/config/config/wc-config.yaml
@@ -1,7 +1,7 @@
 global:
   ck8sVersion: ${CK8S_VERSION}
   cloudProvider: ${CK8S_CLOUD_PROVIDER}
-  environmentName: ${CK8S_ENVIRONMENT_NAME}
+  clusterName: ${CK8S_ENVIRONMENT_NAME}-wc
   dnsPrefix: ${CK8S_ENVIRONMENT_NAME}
   baseDomain: "set-me"
   opsDomain: "set-me"
--- a/helmfile/values/fluentd-user.yaml.gotmpl
+++ b/helmfile/values/fluentd-user.yaml.gotmpl
@@ -159,6 +159,13 @@ extraConfigMaps:
       </metric>
     </filter>
 
+    <filter **>
+      @type record_transformer
+      <record>
+        cluster.name "{{ .Values.global.clusterName }}"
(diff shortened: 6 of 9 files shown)


helmfile/values/fluentd-sc.yaml.gotmpl
```diff
@@ -146,6 +146,13 @@ extraConfigMaps:
       </metric>
     </filter>
 
+    <filter **>
+      @type record_transformer
+      <record>
+        cluster.name "{{ .Values.global.environmentName }}"
```
Xartos, line 152: Do we want to use `.Values.global.environmentName` or should we create a new entry like `.Values.global.clusterName`?

OlleLarsson: I mean leaning towards a clusterName variable instead. 
Can't from the top of my head think of anywhere elase where we are using environment name than in the deploy scripts,  where it seems to be used to derive a cluster name :D 
https://github.com/elastisys/compliantkubernetes-apps/blob/d73f39d3afb022c804b027cfb9e30bb7b603f84e/scripts/deploy-wc.sh#L14

Xartos: Seem to not be used anywhere else than there 
```bash
$ grep -r CLUSTER_NAME
scripts/deploy-wc.sh:export CLUSTER_NAME="${environment}_${cloud_provider}"
```

So I don't really see the benefit from using a separate variable (if we basically don't use this variable anywhere else). I can see that we want to rename `environmentName` to `clusterName` but I don't see the big benefit of having both

OlleLarsson: Agree, rename `environmentName` to `clusterName`

lentzi90 (commented):
This looks nice but I'm still a bit scared that I'm missing something.
To me the environment name was a common name for both/all clusters, whereas a cluster name should be unique for the cluster. So a `global.clusterName` doesn't really make sense to me as it isn't "global". But maybe I'm overthinking this?
To be clear, I think it works the way things are right now, but not when we get to multiple workload clusters or if we want logs also from SC.

Xartos:
> This looks nice but I'm still a bit scared that I'm missing something.
> To me the environment name was a common name for both/all clusters, whereas a cluster name should be unique for the cluster. So a `global.clusterName` doesn't really make sense to me as it isn't "global". But maybe I'm overthinking this?
> To be clear, I think it works the way things are right now, but not when we get to multiple workload clusters or if we want logs also from SC.

Absolutely, maybe I could rephrase it to be more. 
"Removed `environmentName` variable (since we don't use it) and added new variable `clusterName`."

WIP-CHANGELOG.md
```diff
@@ -38,6 +39,7 @@ The following options has been removed or replaced
 - Support for GCS
 - Backup retention for InfluxDB.
 - Add Okta as option for OIDC provider
+- Added record `cluster.name` in all logs to elasticsearch that matches the cluster setting `global.environmentName`
```
viktor-f, line 42:
```suggestion
- Added record `cluster.name` in all logs to elasticsearch that matches the cluster setting `global.clusterName`
```

viktor-f (changes requested):
Changing to `clusterName` sounds good. But should we then also change `CK8S_ENVIRONMENT_NAME` to `CK8S_CLUSTER_NAME` to match? That would of course require changes in a few more places.

Xartos:
> Changing to `clusterName` sounds good. But should we then also change `CK8S_ENVIRONMENT_NAME` to `CK8S_CLUSTER_NAME` to match? That would of course require changes in a few more places.

Good catch! I think that this is what we want to acheive from the multicluster in the end, but for now we init a whole "environment" not cluster by cluster. But maybe we could for now change `clusterName` to default to `${CK8S_ENVIRONMENT_NAME}-<wc|sc>` and when we change to init clusterwise we can rename this variable to `CK8S_CLUSER_NAME`

viktor-f:
> > Changing to `clusterName` sounds good. But should we then also change `CK8S_ENVIRONMENT_NAME` to `CK8S_CLUSTER_NAME` to match? That would of course require changes in a few more places.
> 
> Good catch! I think that this is what we want to acheive from the multicluster in the end, but for now we init a whole "environment" not cluster by cluster. But maybe we could for now change `clusterName` to default to `${CK8S_ENVIRONMENT_NAME}-<wc|sc>` and when we change to init clusterwise we can rename this variable to `CK8S_CLUSER_NAME`

That sounds good :+1:

WIP-CHANGELOG.md
Xartos: Since it's separate commits I thought this would go first and add the fluentd stuff and the next commit changes the name to clusterName

OlleLarsson (approved):

OlleLarsson:
Should probably update our helper script https://github.com/elastisys/ck8s/blob/master/helper/configure-apps.bash#L119

Xartos:
> Should probably update our helper script https://github.com/elastisys/ck8s/blob/master/helper/configure-apps.bash#L119

Already done ;) https://github.com/elastisys/ck8s/pull/665

OlleLarsson:
> > Should probably update our helper script https://github.com/elastisys/ck8s/blob/master/helper/configure-apps.bash#L119
> 
> Already done ;) [elastisys/ck8s#665](https://github.com/elastisys/ck8s/pull/665)

Great :D

WIP-CHANGELOG.md
lentzi90: While it makes sense when reviewing the commits, I think this makes the changelog confusing to read. I would change this line in the wip changelog also as part of the second commit. So you actually rename it everywhere, also in the changelog. :slightly_smiling_face:

Xartos: I changed the order of the two commits. I think it feels better to have them separate since it's two separate fixes. But I agree that with the order changed the WIP-CHANGELOG is more readable

viktor-f (commented):
I think this looks good now, though I have not tested any of it.

One question regarding the log management: with this we would get all (kubernetes) logs in the same index even if we have multiple workload clusters, right? They would be identifiable since they have the `cluster.name` tag, so this is a fine solution. I'm just wondering if it would be even better to have them as separate indices. Just want to check if we have thought about that option.

Xartos:
> I think this looks good now, though I have not tested any of it.
> 
> One question regarding the log management: with this we would get all (kubernetes) logs in the same index even if we have multiple workload clusters, right? They would be identifiable since they have the `cluster.name` tag, so this is a fine solution. I'm just wondering if it would be even better to have them as separate indices. Just want to check if we have thought about that option.

I don't know the pros/cons with that if we are talking "multi-workload-cluster" setup. But to me I think I would like to have the option to look at all logs in all cluster and to filter out if you want to. Maybe I'm not seeing the benefits of having it in its own index. But to me that is more if you want isolation between data.

viktor-f:
> > I think this looks good now, though I have not tested any of it.
> > One question regarding the log management: with this we would get all (kubernetes) logs in the same index even if we have multiple workload clusters, right? They would be identifiable since they have the `cluster.name` tag, so this is a fine solution. I'm just wondering if it would be even better to have them as separate indices. Just want to check if we have thought about that option.
> 
> I don't know the pros/cons with that if we are talking "multi-workload-cluster" setup. But to me I think I would like to have the option to look at all logs in all cluster and to filter out if you want to. Maybe I'm not seeing the benefits of having it in its own index. But to me that is more if you want isolation between data.

Yes it's probably more convenient to have them in the same index. I think that we should probably stick with this for now, mostly wanted to note it as a possibility.
One advantage of separate indices is option to have different retention settings for different clusters. AFAIK you cannot do that with everything in the same index. In the same spirit, if one cluster started spamming logs it would not affect how much logs from other clusters are stored.

Xartos:
> Yes it's probably more convenient to have them in the same index. I think that we should probably stick with this for now, mostly wanted to note it as a possibility.
> One advantage of separate indices is option to have different retention settings for different clusters. AFAIK you cannot do that with everything in the same index. In the same spirit, if one cluster started spamming logs it would not affect how much logs from other clusters are stored.

Right, yea it's definitely an option. But this feels like we might want to create a issue to implement support for this? Since this feels like a bigger task

viktor-f:
> Right, yea it's definitely an option. But this feels like we might want to create a issue to implement support for this? Since this feels like a bigger task

I agree, an investigation issue to see if it's anything we would want.

viktor-f (approved):

PR #115, opened 2020-12-03
apps sc: dex configured to accept OIDC groups from Okta

geoff-elastisys:
**What this PR does / why we need it**:using Okta as an OIDC provider to dex, dex is configured to accept groups

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #444

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -144,6 +144,11 @@ dex:
   tolerations: []
   affinity: {}
   nodeSelector: {}
+  # These two settings are curently only used by okta. If you don't need the groups claims, both options can be set to 'false'.
+  # 'insecureSkipEmailVerified' should only be set to 'true' if the user's okta
+  # configuration does not require a user to verify their identity to okta when their okta account is being created. This is not recommended.
+  insecureSkipEmailVerified: false
+  insecureEnableGroups: true
 
 elasticsearch:
   sso:
--- a/helmfile/values/dex.yaml.gotmpl
+++ b/helmfile/values/dex.yaml.gotmpl
@@ -44,7 +44,14 @@ config:
       redirectURI: https://dex.{{ .Values.global.baseDomain }}/callback
       clientID: {{ .Values.dex.oktaClientID }}
       clientSecret: {{ .Values.dex.oktaClientSecret }}
-      insecureSkipEmailVerified: true
+      insecureSkipEmailVerified: {{ .Values.dex.insecureSkipEmailVerified }}
+      insecureEnableGroups: {{ .Values.dex.insecureEnableGroups }}
+      scopes:
+        - openid
+        - profile
+        - email
+        - groups
+      getUserInfo: true
   {{ end }}
   {{ if eq .Values.dex.oidcProvider "aaa" }}
   - name: AAA
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -38,6 +38,7 @@ The following options has been removed or replaced
 - Support for GCS
 - Backup retention for InfluxDB.
 - Add Okta as option for OIDC provider
+- Dex configuration to accept groups from Okta as an OIDC provider
 
 ### Changed
 


geoff-elastisys:
Tested - see JWT snippets in issue: https://github.com/elastisys/ck8s/issues/444

helmfile/values/dex.yaml.gotmpl
```diff
@@ -45,6 +45,13 @@ config:
       clientID: {{ .Values.dex.oktaClientID }}
       clientSecret: {{ .Values.dex.oktaClientSecret }}
       insecureSkipEmailVerified: true
+      insecureEnableGroups: true
```
lentzi90, line 48: These two options are insecure so I think we should at least make it possible to disable them.

lentzi90 (commented):
Just a note on the insecure options. I think we should expose those options so that it is possible to disable them.

helmfile/values/dex.yaml.gotmpl
geoff-elastisys: insecureSkipEmailVerified we may have to leave as true in the okta set up because dex doesn't have access to that info from okta (as far as I can tell).  Will test the groups flag a bit more and see if we can secure that.  Notes from the dex docs on email verified:

 # Some providers return claims without "email_verified", when they had no usage of emails verification in enrollment process
    # or if they are acting as a proxy for another IDP etc AWS Cognito with an upstream SAML IDP
    # This can be overridden with the below option
    # insecureSkipEmailVerified: true 

Notes on insecureGroups:

 # Groups claims (like the rest of oidc claims through dex) only refresh when the id token is refreshed
    # meaning the regular refresh flow doesn't update the groups claim. As such by default the oidc connector
    # doesn't allow groups claims. If you are okay with having potentially stale group claims you can use
    # this option to enable groups claims through the oidc connector on a per-connector basis.
    # This can be overridden with the below option
    # insecureEnableGroups: true

geoff-elastisys: tested a bit more - without insecureEnableGroups, groups don't get passed by dex to appliaction (they are still sent by okta).  we could still make this an option but essentially if someone wants groups, we need to enable this option

geoff-elastisys: For the email_verified flag, here is the JWT token that okta sends to dex, note email_verified is not present:

{
  "sub": "00u1f7ceswDEKEnme5d6",
  "name": "Geoff Irwin",
  "email": "geoff.irwin@elastisys.com",
  "ver": 1,
  "iss": "https://elastisys-trial.okta.com/oauth2/default",
  "aud": "0oa1f9rmv0txIrlad5d6",
  "iat": 1607069958,
  "exp": 1607073558,
  "jti": "ID.6YzOn3h7wyN6JyU3RCn25USFUyljAnYbysuUyvoCiMg",
  "amr": [
    "pwd"
  ],
  "idp": "00o1c9issxiBuQqYV5d6",
  "nonce": "nonce",
  "preferred_username": "geoff.irwin@elastisys.com",
  "auth_time": 1000,
  "at_hash": "preview_at_hash",
  "groups": [
    "Everyone",
    "grafana_editor"
  ]
}

lentzi90: Thanks for investigating! I think we want to make them configurable even with the limitations when they are disabled. There will always be someone who want to disable them for whatever reason. :slightly_smiling_face:

geoff-elastisys: cool, will do

geoff-elastisys: and done!

config/config/sc-config.yaml
```diff
@@ -144,6 +144,8 @@ dex:
   tolerations: []
   affinity: {}
   nodeSelector: {}
+  insecureSkipEmailVerified: #bool
```
OlleLarsson, line 147: Could you perhaps add a comment that this only affects the okta oidc connector?

geoff-elastisys: will do

config/config/sc-config.yaml
```diff
@@ -144,6 +144,11 @@ dex:
   tolerations: []
   affinity: {}
   nodeSelector: {}
+  # These two settings curently only used by okta.  If a groups claim is not required, both options can be set to false.
+  # Otherwise lets settings as they are to get access to groups.  SkipEmailVerified should only be set to true if the user's okta
+  # configuration does not require a user to verify their identity to okta when their okta account is being created.  This is not recommended.
```
OlleLarsson, line 149:
```suggestion
  # These two settings are curently only used by okta. If you don't need the groups claims, both options can be set to 'false'.
  # 'insecureSkipEmailVerified' should only be set to 'true' if the user's okta
  # configuration does not require a user to verify their identity to okta when their okta account is being created. This is not recommended.
```

lentzi90 (approved):
LGTM!

OlleLarsson (approved):

PR #111, opened 2020-12-03
apps sc: update user grafana chart->6.1.11 and app version-> 7.3.3

geoff-elastisys:
**What this PR does / why we need it**:user grafana updated so as to use JMES roles mapping for group claims

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #110 

**Public facing documentation PR** *(if applicable)*

**Special notes for reviewer**:

**Checklist:**

- [ ] Added relevant notes to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md)
- [ ] Proper commit message prefix on all commits
- [ ] Updated the [public facing documentation](https://github.com/elastisys/compliantkubernetes)
- Is this changeset backwards compatible for existing clusters? Applying:
    - [ ] is completely transparent, will not impact the workload in any way.
    - [ ] requires running a migration script.
    - [ ] will create noticeable cluster degradation.
          E.g. logs or metrics are not being collected or Kubernetes API server
          will not be responding while upgrading.
    - [ ] requires draining and/or replacing nodes.
    - [ ] will change any APIs.
          E.g. removes or changes any CK8S config options or Kubernetes APIs.
    - [ ] will break the cluster.
          I.e. full cluster migration is required.

PR #116, opened 2020-12-03
Geoff grafana user group claims

### LINKED ISSUE #99: [2] Sample roles for grafana
Create two sample role bindings, corresponding to typical access for a developer and admin.
These should then be mapped to OIDC group claims.

The specific names of the group claims that are mapped to roles in harbor should be configurable.

 Aceptance criteria:
- I can login in using my IDP (e.g. google) and either get developer access or admin access based on what roles I have in my IDP.

geoff-elastisys:
**What this PR does / why we need it**:Maps grafana roles to group claims 

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #99 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/helmfile/values/grafana-user.yaml.gotmpl
+++ b/helmfile/values/grafana-user.yaml.gotmpl
@@ -63,13 +63,19 @@ grafana.ini:
     enabled: true
     client_id: grafana
     client_secret: {{ .Values.grafana.clientSecret }}
+    {{ if eq .Values.dex.oidcProvider "google" }}
     scopes: profile email openid
+    {{ end }}
+    {{ if eq .Values.dex.oidcProvider "okta" }}
+    scopes: profile email openid groups
+    {{ end }}
     auth_url: https://dex.{{ .Values.global.baseDomain }}/auth
     token_url: https://dex.{{ .Values.global.baseDomain }}/token
     api_url: https://dex.{{ .Values.global.baseDomain }}/api
     allowed_domains: {{ join " " .Values.dex.allowedDomains }}
     allow_sign_up: true
     tls_skip_verify_insecure: {{ not .Values.global.verifyTls }}
+    role_attribute_path: contains(groups[*], {{ .Values.user.grafana.userGroups.grafanaAdmin }}) && 'Admin' || contains(groups[*], {{ .Values.user.grafana.userGroups.grafanaEditor }}) && 'Editor' || contains(groups[*], {{ .Values.user.grafana.userGroups.grafanaViewer }}) && 'Viewer'
 
 # Velero backup
 
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -57,6 +57,7 @@ The following options has been removed or replaced
 - Add Okta as option for OIDC provider
 - Dex configuration to accept groups from Okta as an OIDC provider
 - Added record `cluster.name` in all logs to elasticsearch that matches the cluster setting `global.clusterName`
+- Role mapping from OIDC groups to roles in user grafana
 
 ### Changed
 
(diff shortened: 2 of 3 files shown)


geoff-elastisys:
see testing notes in issue: https://github.com/elastisys/compliantkubernetes-apps/issues/99

This PR is dependent of two other PR's at the moment so can't be tested independent of those - https://github.com/elastisys/compliantkubernetes-apps/pull/115 and this one https://github.com/elastisys/compliantkubernetes-apps/pull/111

helmfile/values/grafana-user.yaml.gotmpl
```diff
@@ -64,13 +64,14 @@ grafana.ini:
     enabled: true
     client_id: grafana
     client_secret: {{ .Values.grafana.clientSecret }}
-    scopes: profile email openid
+    scopes: profile email openid groups
```
OlleLarsson, line 67: Can this scope be used when using any other OIDC provider than Okta?

helmfile/values/grafana-user.yaml.gotmpl
```diff
@@ -64,13 +64,14 @@ grafana.ini:
...
     auth_url: https://dex.{{ .Values.global.baseDomain }}/auth
     token_url: https://dex.{{ .Values.global.baseDomain }}/token
     api_url: https://dex.{{ .Values.global.baseDomain }}/api
     allowed_domains: {{ join " " .Values.dex.allowedDomains }}
     allow_sign_up: true
     tls_skip_verify_insecure: {{ not .Values.global.verifyTls }}
+    role_attribute_path: contains(groups[*], 'grafana_admin') && 'Admin' || contains(groups[*], 'grafana_editor') && 'Editor' || 'Viewer'
```
OlleLarsson, line 74: So if the group claim is as follows
`"groups": ["grafana_admin"]`
the user would be mapped to the `Admin` role in Grafana?

helmfile/values/grafana-user.yaml.gotmpl
geoff-elastisys: In the ideal world, yes - but some don't do it as scopes but instead as claims.  It shouldn't break any other provider implementation but just not work as expected. So perhaps having the scopes as a conditional dependent on  provider? ie. if statement

helmfile/values/grafana-user.yaml.gotmpl
geoff-elastisys: yes - and all users would have viewer rights.  Can add a viewer group too if we think its needed.

lentzi90: I'm just guessing here so please bear with me, but I would expect that users want to reuse existing groups here and not be forced to use these quite generic group names. For example, there could be a team/group named Vulcan that should have the Editor role and another named Enterprise that should be Admin.

What I'm trying to say is that I think we need to expose this whole line so that users are free to configure it the way they want. Of course we can have this as default, but it must be possible to change it.

lentzi90 (commented):
I think we need to make this more configurable as there probably isn't two users with similar requirements for group names and roles.

geoff-elastisys:
> I think we need to make this more configurable as there probably isn't two users with similar requirements for group names and roles.

Yep, will do, makes perfect sense

helmfile/values/grafana-user.yaml.gotmpl
OlleLarsson: Keep it if it still works with the google oidc connector. Make it configurable if you're feeling ambitious :)

OlleLarsson (approved):

OlleLarsson (approved):

lentzi90 (approved):

PR #117, opened 2020-12-04
Making object storage optional

viktor-f:
**What this PR does / why we need it**: This makes object storage optional. It mostly adds some config options to disable all backups.

I have not fixed the tests yet. They should of course not give any error for a disabled feature.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -11,6 +11,11 @@ global:
   clusterDns: "10.43.0.10"
 
 objectStorage:
+  # Options are "s3", "gcs", or "none"
+  # If "none", remember to disable features that depend on object storage:
+  #   all backups (velero, harbor, influxdb, elasticsearch), sc logs (fluentd)
+  #   Also set harbor persistence to "filesystem" or "swift"
+  # Otherwise configure the features to match this type.
   type: "s3"
   # gcs:
   #   project: "set-me"
@@ -100,12 +105,15 @@ harbor:
         memory: 512Mi
         cpu: 200m
   persistence:
+    # Valid options are "filesystem" (persistent volume), "swift", or "objectStorage" (matching global config)
     type: "set-me"
     disableRedirect: set-me
   oidc:
     #group claim name used by OIDC Provider
     groupClaimName: "set-me"
     scope: openid,email,profile,offline_access,groups
+  backup:
+    enabled: true
 
 prometheus:
   storage:
@@ -279,6 +287,7 @@ elasticsearch:
     additionalPolicies: {}
   # Snapshot and snapshot lifecycle configuration
   snapshot:
+    enabled: true
     min: 7
     max: 14
     ageSeconds: 864000
@@ -312,6 +321,9 @@ elasticsearch:
     tolerations: []
 
 fluentd:
+  # Enable log collection in the service cluster
+  # and store in object storage via fluentd aggregator.
+  enabled: true
   resources:
     limits:
       cpu: 500m
@@ -335,6 +347,7 @@ fluentd:
     affinity: {}
     nodeSelector: {}
 
+# Log retention for service cluster logs stored in object storage.
 logRetention:
   days: 7
 
@@ -495,6 +508,7 @@ ingressNginx:
     nodeSelector: {}
 
 velero:
+  enabled: true
   tolerations: []
   nodeSelector: {}
   resources:
--- a/helmfile/01-applications.yaml
+++ b/helmfile/01-applications.yaml
@@ -92,13 +92,13 @@ releases:
 {{ end }}
 
 
-{{ if .Values.ck8sdash.enabled }}
 - name: ck8sdash
   namespace: ck8sdash
   labels:
     app: ck8sdash
   chart: ./charts/ck8sdash
   version: 0.3.1
+  installed: {{ .Values.ck8sdash.enabled }}
   missingFileHandler: Error
   needs:
   - cert-manager/cert-manager
@@ -110,7 +110,6 @@ releases:
 {{ if eq .Environment.Name "workload_cluster" }}
   - values/ck8sdash-wc.yaml.gotmpl
 {{ end }}
-{{ end }}
 
 # Velero
 - name: velero
@@ -119,6 +118,7 @@ releases:
     app: velero
   chart: vmware-tanzu/velero
   version: 2.8.2
+  installed: {{ .Values.velero.enabled }}
   missingFileHandler: Error
   values:
   - values/velero.yaml.gotmpl
@@ -217,7 +217,6 @@ releases:
   version: 0.1.0
   missingFileHandler: Error
 
-{{ if .Values.user.grafana.enabled }}
 # Grafana instance for user
 - name: user-grafana
   namespace: monitoring
@@ -227,12 +226,12 @@ releases:
     app.kubernetes.io/name: grafana
   chart: grafana/grafana
   version: 6.1.11
+  installed: {{ .Values.user.grafana.enabled }}
   missingFileHandler: Error
   needs:
   - cert-manager/cert-manager
   values:
   - values/grafana-user.yaml.gotmpl
-{{ end }}
 
 
 # prometheus-elasticsearch-exporter
@@ -255,6 +254,7 @@ releases:
     app: elasticsearch-slm
   chart: ./charts/elasticsearch-slm
   version: 0.1.0
+  installed: {{ .Values.elasticsearch.snapshot.enabled }}
   missingFileHandler: Error
   needs:
   - elastic-system/opendistro-es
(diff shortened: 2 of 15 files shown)


helmfile/01-applications.yaml
```diff
@@ -112,6 +112,7 @@ releases:
 {{ end }}
 {{ end }}
 
+{{ if .Values.velero.enabled }}
```
Xartos, line 115: What happens if you have installed it with `velero.enabled=true` and then set it to `false` and run apply again?
Does it uninstall the chart or does nothing happen? 

I know that if you use `condition` for the charts it will just not do anything if you go from enabled to disabled

Xartos (commented):
Really nice! Have you tested what happens if you have backups enabled with object storage set to `none`? Are you able to deploy any of the components that have backups?

viktor-f:
> Really nice! Have you tested what happens if you have backups enabled with object storage set to `none`? Are you able to deploy any of the components that have backups?

If you have object storage as `none` but still have backups enabled, then helmfile will complain that some variables are not set (e.g. s3 endpoint) and throw an error.

Xartos:
> > Really nice! Have you tested what happens if you have backups enabled with object storage set to `none`? Are you able to deploy any of the components that have backups?
> 
> If you have object storage as `none` but still have backups enabled, then helmfile will complain that some variables are not set (e.g. s3 endpoint) and throw an error.

Will it do that for all components individually? Even for our harbor-backup chart?

helmfile/01-applications.yaml
viktor-f: Tried it now and nothing happens if i set `enabled=false` on a cluster that already have velero installed. It would be nice if velero was uninstalled by setting it to false. But I'm not sure how to make that happen.

viktor-f:
> > > Really nice! Have you tested what happens if you have backups enabled with object storage set to `none`? Are you able to deploy any of the components that have backups?
> > 
> > 
> > If you have object storage as `none` but still have backups enabled, then helmfile will complain that some variables are not set (e.g. s3 endpoint) and throw an error.
> 
> Will it do that for all components individually? Even for our harbor-backup chart?

I think so. I might have missed to test some of them.

helmfile/01-applications.yaml
Xartos: I guess you'd need to add `{{ if .Values.velero.enabled }}` around all templates in the chart. But that's not an option for us to do that for all charts. So if there's no other smart way, I would say that this is good as it is.
But then I think you can change the if statement to just a `condition` on the chart instead. If you think that's better looking.

viktor-f:
> > > > Really nice! Have you tested what happens if you have backups enabled with object storage set to `none`? Are you able to deploy any of the components that have backups?
> > > 
> > > 
> > > If you have object storage as `none` but still have backups enabled, then helmfile will complain that some variables are not set (e.g. s3 endpoint) and throw an error.
> > 
> > 
> > Will it do that for all components individually? Even for our harbor-backup chart?
> 
> I think so. I might have missed to test some of them.
> 
> edit: I might be wrong now that I think about it. Will do some testing

Did some new tests and no, this does not work as I thought it did.

So now I'm thinking that we might want to add things like this to our helmfile values:
```
#Example for harbor-backup.yaml.gotmpl
{{ if not (or (eq .Values.objectStorage.type "s3") (eq .Values.objectStorage.type "gcs") ) }}
{{ fail "Harbor backup requires s3 or gcs object storage, see Values.objectStorage.type" }}
{{ end }}
```
Though `fail` does not seem to work (or I'm using the wrong syntax. But do you think that this kind of validation is a good solution?

Xartos:
> Did some new tests and no, this does not work as I thought it did.
> 
> So now I'm thinking that we might want to add things like this to our helmfile values:
> 
> ```
> #Example for harbor-backup.yaml.gotmpl
> {{ if not (or (eq .Values.objectStorage.type "s3") (eq .Values.objectStorage.type "gcs") ) }}
> {{ fail "Harbor backup requires s3 or gcs object storage, see Values.objectStorage.type" }}
> {{ end }}
> ```
> 
> Though `fail` does not seem to work (or I'm using the wrong syntax. But do you think that this kind of validation is a good solution?

Yes, something like that would be nice

helmfile/01-applications.yaml
lentzi90: I think the way to do this is to set `installed: false` or use the `condition` field like @Xartos mentioned.
Example:
```yaml
releases:
- name: velero
  namespace: velero
  # Not sure if condition will work
  condition: velero.enabled
  # If it doesn't this should!
  installed: false
```

helmfile/values/harbor.yaml.gotmpl
```diff
@@ -32,7 +32,11 @@ persistence:
       size: {{ .Values.harbor.redis.persistentVolumeClaim.size }}
   imageChartStorage:
     disableredirect: {{ .Values.harbor.persistence.disableRedirect }}
-    {{ if eq .Values.harbor.persistence.type "swift" }}
+    {{ if eq .Values.harbor.persistence.type "filesystem" }}
+    type: filesystem
+    filesystem:
+      rootdirectory: /storage
```
lentzi90, line 38: How does this work? Will harbor mount this path from the host or something like that? Do we need to add "additional volumes"?

lentzi90 (commented):
Great work!
Added a couple of questions. :slightly_smiling_face:

helmfile/values/harbor.yaml.gotmpl
viktor-f: We actually already have a volume for the registry (https://github.com/elastisys/compliantkubernetes-apps/blob/baaf7893c08afe0058a520f6f3bee8fb439fe551/helmfile/values/harbor.yaml.gotmpl#L25), and this seems to use the volume correctly.

helmfile/01-applications.yaml
viktor-f: `installed: false` seems to work great. It deletes the chart if it was present before and will not complain if some object storage values are not set. Will provide an update soon!

helmfile/values/harbor.yaml.gotmpl
lentzi90: Oh I see. Filesystem apparently means "use a PVC" :smile: well I guess that is fine then :slightly_smiling_face:

viktor-f: Yes, I agree that it's not very intuitive :smile:

lentzi90 (approved):
Nice!
The `installed` field seems like a really good solution!
I feel a bit bad about all the checks for object storage type since it is in a way duplication of code. But I'm not sure how to do it better either.

helmfile/values/harbor.yaml.gotmpl
```diff
@@ -43,22 +50,20 @@ persistence:
...
+    {{ else if eq .Values.harbor.persistence.type "s3" }}
     type: s3
     s3:
       region: {{ .Values.objectStorage.s3.region }}
       regionendpoint: {{ .Values.objectStorage.s3.regionEndpoint }}
       bucket: {{ .Values.objectStorage.buckets.harbor }}
       secretkey: {{ .Values.objectStorage.s3.secretKey }}
       accesskey: {{ .Values.objectStorage.s3.accessKey }}
-    {{ else if eq .Values.objectStorage.type "gcs" }}
+    {{ else if eq .Values.harbor.persistence.type "gcs" }}
```
Xartos, line 61: I think we want to have this set to `objectStorage` or something which took the global object storage settings? Because it's very strange to needing to setup the bucket name/region etc. in the global object storage and then set the type here

Xartos (commented):
Overall great work! Very nice with the validation

helmfile/values/harbor.yaml.gotmpl
viktor-f: I get your point. Though the start of this `if else` chain has `{{ if eq .Values.harbor.persistence.type "filesystem" }}`, that's why I used the same here.

One alternative would be to have `harbor.persistence.type` use either `filesystem` or `objectstorage` and pick the exact type of object storage from the global config. Would that work?

Xartos: The only thing I thought there was that, what if you wanted to have harbor on s3 and all backups on gcs, then that wouldn't be possible.
But I think that the `filesystem`/`objectstorage` is the best for now, and if there's need for it we can always add the possibility to have different objectstorage backends when we actually have a usecase for it.

viktor-f: Ah, I didn't even consider supporting multiple object storages in the cluster. 
As you say I think we will leave that for later. Then I'll make the change suggested.

Xartos (approved):
LGTM!

config/config/sc-config.yaml
```diff
@@ -11,6 +11,11 @@ global:
   clusterDns: "10.43.0.10"
 
 objectStorage:
+  # Options are "s3", "gcs", or "none"
+  # If "none", remember to disable features that depend on object storage:
+  #   all backups (velero, harbor, influxdb, elasticsearch), sc logs (fluentd)
+  #   Also set harbor persistence to "filesystem" or "swift"
```
OlleLarsson, line 17: A bit counter-intuitive imo having `objectStorage: none`, but saying that you can use objectstorage (swift) for harbor persistence :P

viktor-f: I agree, but do you think I should prevent people from using swift while we have `objectStorage: none`?

OlleLarsson: No, not nessecarily as we do not support swift for backups right?

PR #127, opened 2020-12-10
docs: Update readme to clarify objectStorage config

Ajarmar:
**What this PR does / why we need it**: Clarifies what needs to be changed in the config when setting up a new cluster, specifically `objectStorage` values which will cause errors when deploying apps if they are unset.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.


lentzi90 (approved):
Nice!
I think we should strive to have this documented (together with all the other values in those files) in some other place in the future, but right now this should do fine. :slightly_smiling_face:

PR #128, opened 2020-12-10
bin: Include comments in secrets.yaml when initializing

Ajarmar:
**What this PR does / why we need it**: When running `ck8s init`, commented lines describing which `objectStorage` values need to be changed are now included in `secrets.yaml`. This will make it clearer what actually needs to be changed in the configuration when setting up a new cluster.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -63,6 +63,7 @@ One can now for example configure:
 - The wc fluentd tolerations and nodeSelector configuration options are now only specified in the configuration file.
 - Helmfile install error on `user-alertmanager` when `user.alertmanager.enabled: true`.
 - The wrong job name being used for the alertmanager rules in wc when `user.alertmanager.enabled: true`.
+- Commented lines in `secrets.yaml`, showing which `objectStorage` values need to be set, now appear when running `ck8s init`.
 
 ### Removed
 
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -246,26 +246,27 @@ set_harbor_config() {
 # Usage: generate_secrets <config-file>
 generate_secrets() {
     if [[ $# -ne 1 ]]; then
-        log_error "ERROR: number of args in generates_secrets must be 1. #=[$#]"
+        log_error "ERROR: number of args in generate_secrets must be 1. #=[$#]"
         exit 1
     fi
     file=$1
-    if [[ -f $file ]]; then
-        sops_decrypt "$file"
-        yq merge --inplace "$file" "${config_defaults_path}/secrets/sc-secrets.yaml"
-    else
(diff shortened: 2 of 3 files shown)


lentzi90 (commented):
Great! I also want comments to work in these files!
There seem to be something weird going on though when I try this out.

There is an encrypted comment left in the file when decrypting it. :thinking: 
It looks like this:
```yaml
alerts:                                                             
    slack:                                                                                                                              
        apiUrl: somelongsecret
    opsGenie:                                                                                                                           
        apiKey: somelongsecret                                      
        #ENC[AES256_GCM,data:qui5OW8kTnHV0Msi2PrkJ4wrAhkva1+UQSH3aqysG1sHpenhTA==,iv:0Il4Y2Js58++AnDTRT1zMRp1ef7fsb20NDGDlgje6Tk=,tag:KgSHennZohGUBnZiWXBh9Q==,type:comment]
dex: ...
```
I can't figure out why it is there. The object storage comments seem to work fine, so why not this one?

Edit: I found a clue... There is a comment [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/secrets/sc-secrets.yaml#L41) but not [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/secrets/wc-secrets.yaml#L20) and I guess those are merged in some way?

Ajarmar:
> Edit: I found a clue... There is a comment [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/secrets/sc-secrets.yaml#L41) but not [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/secrets/wc-secrets.yaml#L20) and I guess those are merged in some way?

This seems to be the cause. What `generate_secrets` in the init script does here is that it copies over everything from `sc-secrets.yaml` (to get the comments) and then merges in the values from `wc-config.yaml` with yq, but the merge seems to move `dex:` below the comment. sops then can't decrypt the comment because it's in the wrong place. (Moving the comment manually makes it work again.)

I'm not really sure about the easiest solution here (outside of removing the comment or ignoring the problem). We can't copy from both sc-secrets and wc-secrets (need to `yq merge` the second one to avoid duplicates and ensure correct ordering of values), but I'm not sure if it's possible to preserve original ordering of comments when using yq since it just pretends that comments don't exist.

The problem would be fixed if there only was a single secrets file to initialize from (something I discussed briefly with @viktor-f ) but that's a bigger change and probably more suited to being solved like a regular issue.

lentzi90:
Ok. I guess we can remove the comment then and create an issue about implementing a proper fix later.

lentzi90 (approved):
LGTM!

PR #135, opened 2020-12-11
all: Bump kubectl to v1.18.13

lentzi90:
**What this PR does / why we need it**: We now have k8s v1.18 as default in ck8s-cluster, so it makes sense to have the same version here.

**Which issue this PR fixes**: -

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ The apps are installed using a combination of helm charts and manifests with the
 ### Requirements
 
 * A running cluster based on [ck8s-cluster](https://github.com/elastisys/ck8s-cluster)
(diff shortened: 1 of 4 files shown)


OlleLarsson (approved):
Should we use the same minor version as kubeadm `1.18.12`?

viktor-f (approved):

lentzi90:
> Should we use the same minor version as kubeadm `1.18.12`?

Hmm I don't really see the need to be on the exact same patch. It would of course be nice to always have them synced but I think it is better to have the latest than sync them now.

PR #144, opened 2020-12-15
apps: Revert prometheus back to 2.19.1

OlleLarsson:

The new prometheus version does not like running on NFS and it leads to issues with the wal.
```
fs_type=NFS_SUPER_MAGIC msg="This filesystem is not supported and may lead to data corruption and data loss. Please carefully read https://prometheus.io/docs/prometheus/latest/storage/ to learn more about supported filesystems."
...
msg="error tailing WAL" err="failed to find segment for index
```

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -29,7 +29,6 @@ See [migrations docs for elasticsearch-exporter](migration/v0.8.x-v0.9.x/migrate
 - OIDC group claims added to Harbor
 - The options `s3` and `gcs` for `harbor.persistence.type` have been replaced with `objectStorage` and will then match the type set in the global object storage configuration.
 - Bump kubectl to v1.18.13
-- Prometheus upgraded to 2.23.0.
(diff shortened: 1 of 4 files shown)


lentzi90 (approved):
LGTM!

Any suggestion on how to solve this going forward? Should we introduce node local storage for prometheus in places where NFS was previously used? Or maybe just an emptydir?

OlleLarsson:
> LGTM!
> 
> Any suggestion on how to solve this going forward? Should we introduce node local storage for prometheus in places where NFS was previously used? Or maybe just an emptydir?

Yeah, emptyDir seems like a reasonable solution in clusters where only NFS is available. You have to be a bit more aware of what nodes your running on to properly configure retention.
I was hoping that we could replace NFS with rook/ceph but that might be too wishful thinking ;)

Xartos (approved):
LGTM!

PR #151, opened 2021-01-04
apps: Upgrade cert-manager to 1.1.0.

### LINKED ISSUE #95: Upgrade cert-manager
We are currently running v0.14.1 of cert-manager while v1.1.0 is already available at the time of writing this. Needless to say, it would be great for us to be on the GA v1 release and of course to have the latest and greatest.

Cert-manager has excellent [upgrade documentation](https://cert-manager.io/docs/installation/upgrading/), so please check it for details. We will probably need to upgrade one minor version at a time unless we want a breaking change so this issue could be split up to have one for each upgrade if we want to.

OlleLarsson:
To move from alpha/beta to v1. 

fixes #95 

By following the upgrade trail `0.14 -> 0.15 -> 0.16 -> 1.0 -> 1.1` it appears to work just fine for us to upgrade using the [normal upgrade process](https://cert-manager.io/docs/installation/upgrading/) from 0.14 all the way to 1.1.
So, a single `./bin/ck8s apply {sc,wc}` will be enough to upgrade.

There are some caveats with some bugs in kubectl and when using helm to install CRDs, but from what i gathered we are not affected by any of that :)

I had to update the let's encrypt manifests in order to be able apply.
Without the change we'd get 
```
Error: conversion webhook for cert-manager.io/v1alpha2, Kind=Issuer failed: Post https://cert-manager-webhook.cert-manager.svc:443/convert?timeout=30s: service "cert-manager-webhook" not found
```
as we install cert-manager after creating the issuers.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.


lentzi90 (commented):
Super nice!
It is pretty amazing that no special action is needed for this :exploding_head: 
I'm wondering if we should mention in the release notes also that the cert-manager CRDs have been upgraded. It doesn't change anything for the users right now, but eventually I guess the old version will be dropped so it is good to know about.

viktor-f (approved):
LGTM! I agree that it's very nice that we don't have to do anything special to upgrade :slightly_smiling_face:

lentzi90 (commented):
There are some more resources that needs to be updated [here](https://github.com/elastisys/compliantkubernetes-apps/tree/main/helmfile/charts/harbor-certs/templates) and [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/charts/prometheus-instance/templates/issuer.yaml) (although I don't think we really ever use the second).

I think we should add to the release notes also that users should update their cert-manager resources to the v1 API version before the next upgrade, so we don't have to support older versions any longer than necessary.

bootstrap/issuers/helmfile/charts/letsencrypt/templates/letsencrypt.yaml
```diff
@@ -1,6 +1,6 @@
 {{ if .Values.letsencrypt.enabled -}}
 ---
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: cert-manager.io/v1
```
OlleLarsson, line 3: If we do this change in conjucntion with upgrading cert-manager we will have to have a somewhat complicated upgrade path if one uses letsencrypt issuers.
This is because of our workflow:
Bootstrap: upgrade crds -> upgrade letsencrypt issuers which will fail as we have only upgraded the certmanager crds but not the application!
We will get the following error:
```
Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: could not get information about the resource: conversion webhook for cert-manager.io/v1alpha2, Kind=Issuer failed: the server rejected our request for an unknown reason
```

This means that our upgrade path would become
```
# Upgrade crds, wait for it to fail
./bin/ck8s bootstrap {wc.sc}

./bin/ck8s ops helmfile {wc,sc} -l app=cert-manager -i apply

./bin/ck8s bootstrap {wc.sc}
```

OlleLarsson: Maybe I'm just a bit confused,i need to investigate further

lentzi90: I think you are correct that cert-manager must be upgraded before the issuers. So if we want all in one go, it should be
1. Upgrade CRDs
2. Upgrade cert-manager
3. Upgrade issuers/certificates and other resources

I'm fine with leaving the issuers/certificates on v1alpha2 for the next release to keep the migration simple.

OlleLarsson: The thing is that is we leave the issuers at v1alpha2 we will not have any problem when upgrading but problems when installing on a fresh cluster as the it will try to call the cert-manager conversion webhook which has not yet been installed.
So, it's a bit of a chicken and egg problem in either case :smile:

OlleLarsson: Anyways, I tried with the current solution a fresh install in the workload cluster and an upgrade in the serice cluster by upgrading cert-manager before the crds and it seems to work
```
./bin/ck8s ops helmfile sc -l app=cert-manager -i apply
./bin/ck8s apply sc
```

OlleLarsson: As we decided to move the issuers from the bootstrap step to apps this issue is not longer relevant :)

lentzi90 (approved):
LGTM!
Very nicely done! Love the validation before apply!

PR #162, opened 2021-01-12
config: Document settings in wc-config.yaml

OlleLarsson:

Part of
- https://github.com/elastisys/compliantkubernetes/issues/8
- https://github.com/elastisys/ck8s-apps/issues/228

I've used [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml) as reference.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/config/config/wc-config.yaml
+++ b/config/config/wc-config.yaml
@@ -1,31 +1,105 @@
+## Common kubernetes configuration options
+
+## Define resources requests and limits for single Pods.
+## ref: https://kubernetes.io/docs/user-guide/compute-resources/
+## resources: {}
+
+## Node labels for pod assignment
+## ref: https://kubernetes.io/docs/user-guide/node-selection/
+## nodeSelector: {}
+
+## Tolerations for pod assignment
+## ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
+## tolerations: []
+
+## Affinity for pod assignment
+## ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
+## affinity: {}
+
+## Some common options used in various helm charts.
+##
 global:
+  ## Compliantkubernetes-apps version.
+  ## 'any', can be used when running on non-official release.
   ck8sVersion: ${CK8S_VERSION}
+
+  ## Where the cluster is running.
+  ## Supported: exoscale, safespring, citycloud, aws, baremetal
   cloudProvider: ${CK8S_CLOUD_PROVIDER}
+
+  ## The cluster name.
+  ## Used in logs and metrics as to separate these from other clusters.
   clusterName: ${CK8S_ENVIRONMENT_NAME}-wc
+
+  ## Domain intended for ingress usage in the workload cluster
+  ## and to reach user facing services such as kibana, grafana, and harbor.
+  ## E.g. with 'prod.domain.com', kibana is reached via 'kibana.prod.domain.com'.
   baseDomain: "set-me"
+
+  ## Domain intended for ingress usage in the service cluster and to reach
+  ## non-user facing services such as influxdb and elasticsearch.
+  ## E.g. with 'ops.prod.domain.com', elasticsearch is reached via 'elastic.ops.prod.domain.com'.
   opsDomain: "set-me"
+
+  ## Default cert-manager issuer to use for issuing certificates for ingresses.
+  ## Normally one of 'letsencrypt-staging' or 'letsencrypt-prod'.
   issuer: letsencrypt-staging
+
+  ## Verify ingress certificates
   verifyTls: true
+
+  ## IP of the cluster DNS in kubernetes
   clusterDns: "10.43.0.10"
 
+## Configuration of storageclasses.
 storageClasses:
-  # Name of the default storageclass.
-  # Normally one of 'nfs-client', 'cinder-storage', 'local-storage', 'ebs-gp2'.
+  ## Name of the 'default' storageclass in kubernetes.
+  ## Normally one of 'nfs-client', 'cinder-storage', 'local-storage', or 'ebs-gp2'.
   default: "set-me"
 
-  # These are set to null before the first init to be able to set defaults only once
+  ## These enabled flags are 'null' but after first init
+  ## they will be set to eihter 'true' or 'false'.
+  ## On consecutive inits they will not be touched (unless they set to 'null').
+
+  ## Enable deployment of nfsClientProvisioner
+  ## Note, 'nfs-client' is installed as a deafult storagecalss only if 'storageClasses.default: nfs-client'.
   nfs:
     enabled: null # true | false
+
+  ## Enableds installation of 'cinder-storage'  as a defaultstorageclass.
   cinder:
     enabled: null # true | false
+
+  ## Enables deployment of local-volume-provisioner and installation
+  ## of 'local-storage' storageclass.
   local:
     enabled: null # true | false
+
+  ## Enables installation of 'ebs-gp2' as a default storageclass.
   ebs:
     enabled: null # true | false
 
+## Nfs-client-provisioner configuration.
+## Deployment is controlled via 'storageClasses.nfs.enabled'.
+nfsProvisioner:
+  server: "set-me"
+  path: /nfs
+  resources:
+    limits:
+      cpu: 100m
+      memory: 128Mi
+    requests:
+      cpu: 100m
+      memory: 128Mi
+
+  tolerations: []
+  nodeSelector: {}
+  affinity: {}
+
+## Object storage configuration for backups.
 objectStorage:
-  # Options are "s3", "gcs", or "none"
-  # If "none", remember to disable backups (velero)
+  ## Options are 's3', 'gcs', or 'none'
+  ## If 'none', remember to disable backups (velero)
   type: "s3"
   # gcs:
   #   project: "set-me"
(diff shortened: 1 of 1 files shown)


config/config/wc-config.yaml
```diff
@@ -4,99 +4,187 @@ global:
   clusterName: ${CK8S_ENVIRONMENT_NAME}-wc
   baseDomain: "set-me"
   opsDomain: "set-me"
+
+  ## Default cert-manager issuer to use in ingresses
   issuer: letsencrypt-staging
+
+  ## Verify ingress certificates
+  ##
```
viktor-f, line 12: I think you can drop this and all other lines with only `##`. It will at least make the file a bit shorter.
Unless you have some reason for having them there.
```suggestion
  ## Verify ingress certificates
```

config/config/wc-config.yaml
```diff
@@ -156,80 +302,152 @@ ck8sdash:
...
+
+    ## Node labels for pod assignment
+    ## ref: https://kubernetes.io/docs/user-guide/node-selection/
+    ##
     nodeSelector: {}
+
     config:
       useProxyProtocol: "set-me"
+
+    ## Use host ports 80 and 443
```
viktor-f, line 376:
```suggestion
    ## Use host ports 80 or 443
```
You can't use both right?

OlleLarsson: https://github.com/kubernetes/ingress-nginx/blob/master/charts/ingress-nginx/values.yaml#L56-L63

config/config/wc-config.yaml
```diff
@@ -4,99 +4,187 @@ global:
   clusterName: ${CK8S_ENVIRONMENT_NAME}-wc
   baseDomain: "set-me"
   opsDomain: "set-me"
```
viktor-f, line 6: The global variables could also get some simple comments

config/config/wc-config.yaml
```diff
@@ -105,41 +193,96 @@ elasticsearch:
...
+  ## Affinity for pod assignment
+  ## ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
+  ##
   affinity: {}
+
+  ## Extra fluentd config to mount
+  ##
   extraConfigMaps: {}
+
   user:
```
lentzi90, line 233: This one I think it would be good to comment on. Something like
> Configuration for the user-controllable fluentd that ships logs from nodes that the user can run pods on.

config/config/wc-config.yaml
```diff
@@ -105,41 +193,96 @@ elasticsearch:
   clientNode:
     count: 1
 
+## Configuration for fluentd
+##
```
lentzi90, line 197: It would be nice to mention here a bit more about it I think. For example that this is the "system" fluentd that runs on control-plane nodes and other "reserved" nodes that the user cannot run pods on.

config/config/wc-config.yaml
```diff
@@ -4,99 +4,187 @@ global:
...
 
+## Open policy agent configuration
+##
 opa:
   enabled: true
+
+  ## Enable rule that requires pods to come from
+  ## the image registry defined by "URL".
+  ## "enforcement" can assume either "dryrun" or "deny".
+  ##
```
lentzi90, line 161: Very good!

config/config/wc-config.yaml
```diff
@@ -4,99 +4,187 @@ global:
...
     hostPort: "http://kube-prometheus-stack-alertmanager.monitoring:9093"
-    # if type=slack falco.alerts.slackWebhook must be set in the secrets yaml file
+    ## if type=slack falco.alerts.slackWebhook must be set in the secrets yaml file
 
+## Prometheus configuration
+##
 prometheus:
   remoteWrite:
-    # Same as 'influxDB.users.wcWriter'.
+    ## Same as 'influxDB.users.wcWriter'.
```
lentzi90, line 110: This comment is confusing to me. Does it mean that it should be the same as specified for influxdb in sc-config.yaml?
I think it would make more sense to explain what it is: the user used when connecting to the remote write endpoint, by default influxdb in SC.

lentzi90 (commented):
Nice start!

I think we could put a description of "standard" options at the top of the file to avoid repeating them all the time, (e.g. resources, tolerations, enabled). Other options could use some more descriptions instead. I added some comments below on some that I think could have more.

config/config/wc-config.yaml
```diff
@@ -4,7 +4,12 @@ global:
...
   baseDomain: "set-me"
   opsDomain: "set-me"
+
+  ## Default cert-manager issuer to use in ingresses
   issuer: letsencrypt-staging
+
+  ## Verify ingress certificates
+  ##
   verifyTls: true
   clusterDns: "10.43.0.10"
```
Xartos, line 14: How do you decide on what variables to comment? I'm guessing you only comments the lines you feel necessary, but I'm curios if you have any process for deciding.

config/config/wc-config.yaml
```diff
@@ -22,53 +27,95 @@ storageClasses:
   ebs:
     enabled: null # true | false
 
+  storageClass: "set-me"
```
Xartos, line 30: Why is this here?

config/config/wc-config.yaml
```diff
@@ -22,53 +27,95 @@ storageClasses:
...
   type: "s3"
   # gcs:
   #   project: "set-me"
   # s3:
   #   region: "set-me"
   #   regionAddress: "set-me"
   #   regionEndpoint: "set-me"
+
+  ##
+  ## TODO: remove all but for 'velero'.
```
Xartos, line 50: Why remove those?

config/config/wc-config.yaml
```diff
@@ -22,53 +27,95 @@ storageClasses:
...
   buckets:
     harbor: "${CK8S_ENVIRONMENT_NAME}-harbor"
     velero: "${CK8S_ENVIRONMENT_NAME}-velero"
     elasticsearch: "${CK8S_ENVIRONMENT_NAME}-es-backup"
     influxDB: "${CK8S_ENVIRONMENT_NAME}-influxdb"
     scFluentd: "${CK8S_ENVIRONMENT_NAME}-sc-logs"
 
+## User configuration.
 user:
   # This only controls if the namespaces should be created, user RBAC is always created.
```
Xartos, line 61: Add two `#` here as well?

config/config/wc-config.yaml
```diff
@@ -22,53 +27,95 @@ storageClasses:
...
   createNamespaces: true
+  ## List of user namespaces to create.
+  ##
   namespaces:
```
Xartos, line 65: Should we make sure to leave a empty line above all these comments? (seems like that how you've done it for most of the others

config/config/wc-config.yaml
```diff
@@ -22,53 +27,95 @@ storageClasses:
...
+
+  ## Alertmanager configuration.
+  ##
   alertmanager:
     enabled: false
     namespace: monitoring
     ingress:
       enabled: false
 
+
```
Xartos, line 81: Maybe make sure exactly one empty line between

config/config/wc-config.yaml
```diff
@@ -84,41 +131,85 @@ falco:
     tolerations: []
     affinity: {}
     nodeSelector: {}
+    ## if type=slack falco.alerts.slackWebhook must be set in the secrets yaml file
```
Xartos, line 134: Shouldn't this be in the secrets file instead?

config/config/wc-config.yaml
```diff
@@ -84,41 +131,85 @@ falco:
...
+  ## ref: https://kubernetes.io/docs/user-guide/compute-resources/
+  ##
   resources:
     requests:
       memory: 1Gi
       cpu: 300m
     limits:
       memory: 2Gi
-      cpu: "1"
+      cpu: "1" ## Must be a string
```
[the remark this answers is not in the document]
lentzi90: The API-server complains. I'm not sure exactly why though, haven't looked into it, but I'm guessing that this is set on the Prometheus object and that the CRD requires it to be a string for some reason.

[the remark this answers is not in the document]
lentzi90: Actually, it looks like string or integer both should be fine, but no floats though. Maybe that is just in newer versions or something?
https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/crds/crd-prometheuses.yaml#L1987

config/config/wc-config.yaml
OlleLarsson: Because only the velero bucket is used in the workload cluster.

config/config/wc-config.yaml
OlleLarsson: rebase :)

config/config/wc-config.yaml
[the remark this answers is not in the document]
Xartos: I just looked at it. And it seems like the CRD accepts both int and string in the new version (0.45) but I think we use 0.44 which only accepts a string. So that must be it.

config/config/wc-config.yaml
Xartos: Ofcorse, I didn't think about that.

config/config/wc-config.yaml
viktor-f: I misunderstood the comment. I thought it meant that I should set `useHostPort: 80` or `useHostPort: 443`. But here I see that it should be `true` or `false`: https://github.com/elastisys/compliantkubernetes-apps/blob/15038efd79296dde6e996233e88dbfc096dd6a0a/helmfile/values/ingress-nginx.yaml.gotmpl#L35

Maybe change it to this (or something similar):
```suggestion
    ## If `true`, nginx will use host ports 80 and 443
```

lentzi90 (approved):
This is great!

config/config/wc-config.yaml
viktor-f: Could you remove these now instead of just adding a TODO? If they are not used anywhere I don't think it should be a problem.

config/config/wc-config.yaml
```diff
@@ -1,78 +1,158 @@
...
+##
+## tolerations: []
+
+## Affinity for pod assignment
+## ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
+##
+## affinity: {}
+
+## Some common options used in various helm charts.
+##
```
viktor-f, line 24: You seem to have removed the lines with just `##` in the rest of the file, so you could do the same here.
```suggestion
## Define resources requests and limits for single Pods.
## ref: https://kubernetes.io/docs/user-guide/compute-resources/
## resources: {}

## Node labels for pod assignment
## ref: https://kubernetes.io/docs/user-guide/node-selection/
## nodeSelector: {}

## Tolerations for pod assignment
## ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
## tolerations: []

## Affinity for pod assignment
## ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
## affinity: {}

## Some common options used in various helm charts.
```

config/config/wc-config.yaml
```diff
@@ -178,12 +304,19 @@ ck8sdash:
...
+  ## Source IP range to allow.
   whitelistRange:
     global: "0.0.0.0/0"
     ck8sdash: false
 
+## Nfs-client-provisioner configuration.
+## Deployment is controlled via 'storageClasses.nfs.enabled'.
 nfsProvisioner:
   server: ""
   path: /nfs
```
viktor-f, line 322: Perhaps move this block so that it's directly below the storage config.

viktor-f (commented):
Looks very nice! Just some minor things

config/config/wc-config.yaml
OlleLarsson: Ok

viktor-f (approved):
LGTM!

Xartos (commented):
Tiny remarks, otherwise it looks great!

PR #168, opened 2021-01-15
apps: stop on lint errors

### LINKED ISSUE #107: [3] Enforce helm chart linting
We currently do not fail on linting errors in the helm charts. See [this](https://github.com/elastisys/compliantkubernetes-apps/blob/main/.github/workflows/pipeline.yml#L49).
Remove the `continue-on-error` setting to enforce the linter and fix any issues you find.

See also #105. We may want to run `helmfile template` instead or in addition to `helmfile lint` but this is not strictly part of this issue.

Definition of done: The pipeline fails if there are linting/templating errors.

pettersv:
Makes the linter stop on errors, as apposed to continuing, thereby exposing configuration errors. Also fixes some blank/errorneous configuration errors that made the linter unhappy.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #107

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -117,6 +117,12 @@ set_storage_class() {
           storage_class=cinder-storage
           es_storage_class=cinder-storage
 
+          [ "$(yq read "$file" 'storageClasses.nfs.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.nfs.enabled' false
+          [ "$(yq read "$file" 'storageClasses.local.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.local.enabled' false
+          [ "$(yq read "$file" 'storageClasses.ebs.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.ebs.enabled' false
           [ "$(yq read "$file" 'storageClasses.cinder.enabled')" = "null" ] &&
             yq write --inplace "$file" 'storageClasses.cinder.enabled' true
           ;;
@@ -129,21 +135,38 @@ set_storage_class() {
             yq write --inplace "$file" 'storageClasses.nfs.enabled' true
           [ "$(yq read "$file" 'storageClasses.local.enabled')" = "null" ] &&
             yq write --inplace "$file" 'storageClasses.local.enabled' true
+          [ "$(yq read "$file" 'storageClasses.ebs.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.ebs.enabled' false
+          [ "$(yq read "$file" 'storageClasses.cinder.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.cinder.enabled' false
           ;;
 
         aws)
           storage_class=ebs-gp2
           es_storage_class=ebs-gp2
 
+          [ "$(yq read "$file" 'storageClasses.nfs.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.nfs.enabled' false
+          [ "$(yq read "$file" 'storageClasses.local.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.local.enabled' false
           [ "$(yq read "$file" 'storageClasses.ebs.enabled')" = "null" ] &&
             yq write --inplace "$file" 'storageClasses.ebs.enabled' true
+          [ "$(yq read "$file" 'storageClasses.cinder.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.cinder.enabled' false
           ;;
 
         baremetal)
           storage_class=node-local
           es_storage_class=node-local
+
+          [ "$(yq read "$file" 'storageClasses.nfs.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.nfs.enabled' false
           [ "$(yq read "$file" 'storageClasses.local.enabled')" = "null" ] &&
             yq write --inplace "$file" 'storageClasses.local.enabled' true
+          [ "$(yq read "$file" 'storageClasses.ebs.enabled')" = "null" ] &&
+            yq write --inplace "$file" 'storageClasses.ebs.enabled' false
+          [ "$(yq read "$file" 'storageClasses.cinder.enabled')" = "null" ] &&
+           yq write --inplace "$file" 'storageClasses.cinder.enabled' false
           ;;
     esac
 
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -13,6 +13,7 @@ storageClasses:
   # Normally one of 'nfs-client', 'cinder-storage', 'local-storage', 'ebs-gp2'.
   default: "set-me"
 
+  # These are set to null before the first init to be able to set defaults only once
   nfs:
     enabled: null # true | false
   cinder:
--- a/pipeline/lint.bash
+++ b/pipeline/lint.bash
@@ -4,13 +4,40 @@ set -eu -o pipefail
 
 here="$(dirname "$(readlink -f "$0")")"
 
-# TODO: Try to get rid of the NFS server IP.
-# We need to set the NFS server IP on Exoscale for the nfs-provisioner.
-NFS_SC_SERVER_IP=127.0.0.1 helmfile -e service_cluster -f "${here}/../helmfile/" lint
-NFS_WC_SERVER_IP=127.0.0.1 helmfile -e workload_cluster -f "${here}/../helmfile/" lint
+#We should only lint our "own" charts, not the upstream ones
+#TODO: enable linting of ck8sdash and kubeapi-metrics after the PGP issues in the pipeline have been solved (https://github.com/elastisys/compliantkubernetes-apps/issues/184)
 
-helmfile -e service_cluster -f "${here}/../bootstrap/namespaces/helmfile/" lint
-helmfile -e workload_cluster -f "${here}/../bootstrap/namespaces/helmfile/" lint
+charts_ignore_list=(
+  "app!=ck8sdash"
+  "app!=sc-logs-retention"
+  "app!=fluentd-configmap"
+  "app!=ingress-nginx"
+  "app!=cert-manager"
+  "app!=kube-prometheus-stack"
+  "app!=velero"
+  "app!=metrics-server"
+  "app!=dex"
+  "app!=wc-scraper"
+  "app!=prometheus-auth"
+  "app!=wc-reader"
+  "app!=prometheus-wc-reader"
+  "app!=user-grafana"
+  "app!=prometheus-elasticsearch-exporter"
+  "app!=harbor"
+  "app!=influxdb"
+  "app!=fluentd"
+  "app!=fluentd-aggregator"
+  "app!=opendistro"
+  "app!=falco"
+  "app!=falco-exporter"
+  "app!=user-alertmanager"
+  "app!=kubeapi-metrics")
 
-NFS_SC_SERVER_IP=127.0.0.1 helmfile -e service_cluster -f "${here}/../bootstrap/storageclass/helmfile/" lint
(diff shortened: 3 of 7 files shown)


pipeline/lint.bash
```diff
@@ -6,11 +6,11 @@ here="$(dirname "$(readlink -f "$0")")"
 
 # TODO: Try to get rid of the NFS server IP.
 # We need to set the NFS server IP on Exoscale for the nfs-provisioner.
-NFS_SC_SERVER_IP=127.0.0.1 helmfile -e service_cluster -f "${here}/../helmfile/" lint
-NFS_WC_SERVER_IP=127.0.0.1 helmfile -e workload_cluster -f "${here}/../helmfile/" lint
+NFS_SC_SERVER_IP=127.0.0.1 helmfile -e service_cluster --selector chart=blackbox,ck8sdash -f "${here}/../helmfile/" lint
+#NFS_WC_SERVER_IP=127.0.0.1 helmfile -e workload_cluster -f "${here}/../helmfile/" lint
```
OlleLarsson, line 10: You should be able to remove `NFS_{WC,SC}_SERVER_IP`.

pettersv: Removed

bootstrap/namespaces/helmfile/helmfile.yaml
```diff
@@ -6,14 +6,14 @@ helmDefaults:
 environments:
   workload_cluster:
     values:
-      - "{{ requiredEnv "CK8S_CONFIG_PATH" }}/wc-config.yaml"
+      - "/github/workspace/{{ requiredEnv "CK8S_CONFIG_PATH" }}/wc-config.yaml"
```
Xartos, line 9: This will break local runs since this path is (probably) invalid. Please edit CK8S_CONFIG_PATH instead

helmfile/00-psp_rbac.yaml
```diff
@@ -6,14 +6,18 @@ helmDefaults:
 environments:
   workload_cluster:
     values:
-      - "{{ requiredEnv "CK8S_CONFIG_PATH" }}/wc-config.yaml"
+      - "/github/workspace/{{ requiredEnv "CK8S_CONFIG_PATH" }}/wc-config.yaml"
```
Xartos, line 9: Same here.

pipeline/lint.bash
```diff
@@ -4,13 +4,15 @@ set -eu -o pipefail
 
 here="$(dirname "$(readlink -f "$0")")"
 
-# TODO: Try to get rid of the NFS server IP.
-# We need to set the NFS server IP on Exoscale for the nfs-provisioner.
-NFS_SC_SERVER_IP=127.0.0.1 helmfile -e service_cluster -f "${here}/../helmfile/" lint
-NFS_WC_SERVER_IP=127.0.0.1 helmfile -e workload_cluster -f "${here}/../helmfile/" lint
+#We only lint our "own" charts, not the upstream ones
+charts_ignore_list="app!=ck8sdash,app!=sc-logs-retention,app!=fluentd-configmap,app!=ingress-nginx,app!=cert-manager,app!=kube-prometheus-stack,app!=velero,app!=metrics-server,app!=dex,app!=wc-scraper,app!=prometheus-auth,app!=wc-reader,app!=prometheus-wc-reader,app!=user-grafana,app!=prometheus-elasticsearch-exporter,app!=harbor,app!=influxdb,app!=fluentd,app!=fluentd-aggregator,app!=opendistro,app!=falco,app!=falco-exporter,app!=user-alertmanager,app!=kubeapi-metrics"
```
Xartos, line 8: To make it easier to get a grasp of what applications we are ignoring I suggest:
```suggestion
charts_ignore_list=(
  app!=ck8sdash
  app!=sc-logs-retention
  app!=fluentd-configmap
  app!=ingress-nginx
  app!=cert-manager
  app!=kube-prometheus-stack
  app!=velero
  app!=metrics-server
  app!=dex
  app!=wc-scraper
  app!=prometheus-auth
  app!=wc-reader
  app!=prometheus-wc-reader
  app!=user-grafana
  app!=prometheus-elasticsearch-exporter
  app!=harbor
  app!=influxdb
  app!=fluentd
  app!=fluentd-aggregator
  app!=opendistro
  app!=falco
  app!=falco-exporter
  app!=user-alertmanager
  app!=kubeapi-metrics )
```

pipeline/lint.bash
```diff
@@ -4,13 +4,15 @@ set -eu -o pipefail
...
+#TODO: enable linting of ck8sdash and kubeapi-metrics after the PGP issues in the pipeline have been solved
 
-helmfile -e service_cluster -f "${here}/../bootstrap/namespaces/helmfile/" lint
-helmfile -e workload_cluster -f "${here}/../bootstrap/namespaces/helmfile/" lint
+helmfile -e service_cluster -f "${here}/../helmfile/" -l "$charts_ignore_list" lint
```
Xartos, line 11: To get the string we want we would need to change this to:
```suggestion
helmfile -e service_cluster -f "${here}/../helmfile/" -l "$(IFS=',' ; echo "${charts_ignore_list[*]}")" lint
```

Same for the rest of the lines using this variable

Xartos (changes requested):

bootstrap/namespaces/helmfile/helmfile.yaml
pettersv: I was afraid that that would break things :) Will fix.

pipeline/lint.bash
pettersv: Good idea, have changed to this implementation.

config/config/sc-config.yaml
```diff
@@ -14,13 +14,13 @@ storageClasses:
...
+    enabled: false # true | false
   cinder:
-    enabled: null # true | false
+    enabled: false # true | false
   local:
-    enabled: null # true | false
+    enabled: false # true | false
   ebs:
-    enabled: null # true | false
+    enabled: false # true | false
```
OlleLarsson, line 23: Please update https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/init.bash#L109.

OlleLarsson (changes requested):

config/config/sc-config.yaml
pettersv: Done

bin/init.bash
```diff
@@ -117,32 +117,32 @@ set_storage_class() {
           storage_class=cinder-storage
           es_storage_class=cinder-storage
 
-          [ "$(yq read "$file" 'storageClasses.cinder.enabled')" = "null" ] &&
+          [ "$(yq read "$file" 'storageClasses.cinder.enabled')" = "false" ] &&
```
Xartos, line 120: Isn't it better to have this like:
```suggestion
          [ "$(yq read "$file" 'storageClasses.cinder.enabled')" != "true" ] &&
```
Then it doesn't matter what it's set to from the beginning. As long as it's not `true`, change it to `true`

pipeline/lint.bash
```diff
@@ -4,13 +4,42 @@ set -eu -o pipefail
 
 here="$(dirname "$(readlink -f "$0")")"
 
-# TODO: Try to get rid of the NFS server IP.
-# We need to set the NFS server IP on Exoscale for the nfs-provisioner.
-NFS_SC_SERVER_IP=127.0.0.1 helmfile -e service_cluster -f "${here}/../helmfile/" lint
-NFS_WC_SERVER_IP=127.0.0.1 helmfile -e workload_cluster -f "${here}/../helmfile/" lint
+#charts_ignore_list="app!=ck8sdash,app!=sc-logs-retention,app!=fluentd-configmap,app!=ingress-nginx,app!=cert-manager,app!=kube-prometheus-stack,app!=velero,app!=metrics-server,app!=dex,app!=wc-scraper,app!=prometheus-auth,app!=wc-reader,app!=prometheus-wc-reader,app!=user-grafana,app!=prometheus-elasticsearch-exporter,app!=harbor,app!=influxdb,app!=fluentd,app!=fluentd-aggregator,app!=opendistro,app!=falco,app!=falco-exporter,app!=user-alertmanager,app!=kubeapi-metrics"
```
Xartos, line 7: Get rid of this line

pipeline/lint.bash
```diff
@@ -4,13 +4,42 @@ set -eu -o pipefail
...
+  "app!=fluentd-aggregator"
+  "app!=opendistro"
+  "app!=falco"
+  "app!=falco-exporter"
+  "app!=user-alertmanager"
+  "app!=kubeapi-metrics")
 
-NFS_SC_SERVER_IP=127.0.0.1 helmfile -e service_cluster -f "${here}/../bootstrap/storageclass/helmfile/" lint
-NFS_WC_SERVER_IP=127.0.0.1 helmfile -e workload_cluster -f "${here}/../bootstrap/storageclass/helmfile/" lint
+#TODO: enable linting of ck8sdash and kubeapi-metrics after the PGP issues in the pipeline have been solved
```
Xartos, line 36: Add a link to the issue, so it's easy to find

Xartos (changes requested):
Just some minor things

bin/init.bash
pettersv: Yes probably. Makes it a bit more robust.

pipeline/lint.bash
pettersv: Done and done

Xartos (approved):
LGTM!

bin/init.bash
```diff
@@ -117,32 +117,32 @@ set_storage_class() {
...
             yq write --inplace "$file" 'storageClasses.cinder.enabled' true
           ;;
 
         exoscale)
           storage_class=nfs-client
           es_storage_class=local-storage
 
-          [ "$(yq read "$file" 'storageClasses.nfs.enabled')" = "null" ] &&
+          [ "$(yq read "$file" 'storageClasses.nfs.enabled')" != "true" ] &&
             yq write --inplace "$file" 'storageClasses.nfs.enabled' true
```
OlleLarsson, line 129: I'm not a fan of this because we should only set this if it is unset so that a user is free to do what ever with it.
Now if a user isn't happy with `true` and sets it to `false`, running `ck8s init` again will overwrite the users value, which in my opinion isn't desirable.

Whats opinion on this @lentzi90 @Xartos ?

pettersv: The problem is that you cant use "null" as the linter fails on that. It is supposed to be a boolean value. I had it as [ "$(yq read "$file" 'storageClasses.nfs.enabled')" = "false" ] but after a conversation with  @Xartos I changed it to what it is now. Neither way is perfect.

Xartos: Ooh, I totally misinterpreted that, I thought this was a pipeline init thing. But I totally agree with you!
But why do we need this at all? Now that we default it to false instead of null

pettersv: I guess we could just remove it? I'm also not a fan of settings "magically" being changed without the user knowing about it.

OlleLarsson: > Ooh, I totally misinterpreted that, I thought this was a pipeline init thing. But I totally agree with you!
> But why do we need this at all? Now that we default it to false instead of null

We don't but we seem to like to set nice defaults for users depending on the cloud provider that they are running on.

Xartos: Right, is this the reason to why they were defaulted to `null` to indicate that nothing was set. So that it shouldn't reset to default the next time you run `init`

OlleLarsson: Yep. And with my solution proposed in the first comment, no field would be null after init, and they would not be touched upon consecutive inits :)

Xartos: Yes, I think that's a nice solution.
Maybe we can add some comments describing why they are set to `null` to minimize confusion if you're reading the config file directly. Like:
```bash
# These are set to null before the first init to be able to set defaults only once
```

pettersv: Implemented according to @OlleLarsson suggestion

OlleLarsson (approved):

PR #174, opened 2021-01-15
apps sc: disables clair and updates Harbor config with resource requests and more    

viktor-f:
**What this PR does / why we need it**: This turns of clair in Harbor which was using a lot of resources even though trivy was the default scanner. It also adds resource requests to all Harbor pods based on numbers from running clusters.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -68,54 +68,80 @@ user:
 
 harbor:
   enabled: true
+  # The tolerations, affinity, and nodeSelector are applied to all harbor pods.
   tolerations: []
   affinity: {}
   nodeSelector: {}
   chartmuseum:
     persistentVolumeClaim:
       size: 5Gi
+    resources:
+      requests:
+        memory: 16Mi
+        cpu: 5m
   core:
     resources:
       requests:
-        memory: 256Mi
-        cpu: 100m
-      limits:
-        memory: 512Mi
-        cpu: 200m
+        memory: 64Mi
+        cpu: 10m
   database:
     persistentVolumeClaim:
       size: 1Gi
+    resources:
+      requests:
+        memory: 512Mi
+        cpu: 250m
   jobservice:
     persistentVolumeClaim:
       size: 1Gi
     resources:
       requests:
-        memory: 256Mi
-        cpu: 100m
-      limits:
-        memory: 512Mi
-        cpu: 200m
+        memory: 64Mi
+        cpu: 10m
   registry:
     persistentVolumeClaim:
       size: 5Gi
     resources:
       requests:
-        memory: 256Mi
-        cpu: 100m
-      limits:
-        memory: 512Mi
-        cpu: 200m
+        memory: 64Mi
+        cpu: 10m
+    controller:
+      resources:
+        requests:
+          memory: 16Mi
+          cpu: 1m
   redis:
     persistentVolumeClaim:
       size: 1Gi
+    resources:
+      requests:
+        memory: 32Mi
+        cpu: 10m
   notary:
     resources:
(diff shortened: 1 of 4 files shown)


lentzi90 (approved):
LGTM!

helmfile/values/harbor.yaml.gotmpl
```diff
@@ -73,6 +75,9 @@ externalURL: https://harbor.{{ .Values.global.baseDomain }}
 
 harborAdminPassword: {{ .Values.harbor.password }}
 
+clair:
+  enabled: false
```
OlleLarsson, line 79: Did both trivy and clair run before this?

config/config/sc-config.yaml
```diff
@@ -68,54 +68,78 @@ user:
...
+        memory: 16Mi
+        cpu: 5m
+  trivy:
+    persistentVolumeClaim:
+      size: 5Gi
+    resources:
+      requests:
         cpu: 200m
+        memory: 512Mi
+      limits: null #the chart has default limits
```
OlleLarsson, line 142: So does this make it so that the default limits isn't used?

Xartos (commented):
LGTM! Waiting with approving until Olles comments are resolved

helmfile/values/harbor.yaml.gotmpl
viktor-f: Yes

config/config/sc-config.yaml
viktor-f: Yes

Xartos: And why do we not want to use the default limits?

viktor-f: I mostly added this to be in line with the other pods for harbor. There has been some discussions on if you should have limits, at least for CPU. 
But if you want us to have the limits then I'm fine with that as well.

Xartos: Yes for CPU I can understand. But I guess for memory it could make sense. Especially since the upstream chart has it for this but not for the others, then it might be there for a reason.
But I'm also okey with leaving it as is.

helmfile/values/harbor.yaml.gotmpl
```diff
@@ -106,3 +114,41 @@ notary:
 database:
   internal:
     password: {{ .Values.harbor.databasePassword }}
+    resources: {{- toYaml .Values.harbor.database.resources | nindent 6 }}
+    nodeSelector: {{- toYaml .Values.harbor.nodeSelector | nindent 6 }}
+    affinity: {{- toYaml .Values.harbor.affinity | nindent 6 }}
+    tolerations: {{- toYaml .Values.harbor.tolerations | nindent 6 }}
```
OlleLarsson, line 120: Feel free to allign these if you want :)
```suggestion
    resources:    {{- toYaml .Values.harbor.database.resources | nindent 6 }}
    nodeSelector: {{- toYaml .Values.harbor.nodeSelector | nindent 6 }}
    affinity:     {{- toYaml .Values.harbor.affinity | nindent 6 }}
    tolerations:  {{- toYaml .Values.harbor.tolerations | nindent 6 }}
```

OlleLarsson (approved):

config/config/sc-config.yaml
viktor-f: Tried to do some digging in the history of the chart but I could not find any reason why the limits/requests were at those values. 
I guess it could be cleaner just use the limits from the chart. Though they would be slightly "hidden" when looking at these values, since they are just propagated from the char.

PR #175, opened 2021-01-15
Ev/ambassador

vanneback:
**What this PR does / why we need it**: Adds ambassador support to the workload cluster.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes elastisys/compliantkubernetes#36

https://github.com/elastisys/compliantkubernetes/pull/39/

I added it to config for sc-cluster aswell before i noticed the task only applied to the workload-cluster. Maybe i should remove the config from sc-cluster or add full support for it as well?

It will require adding a config option amdassador.enabled to your config.

I will squash commits and add proper message once reviewed.

**Checklist:**

    - [x] will change any APIs.
    - [ ] will break the cluster.


helmfile/values/ambassador.yaml.gotmpl
```diff
@@ -0,0 +1,9 @@
+service:
+  type: ClusterIP
+hostNetwork: true
+dnsPolicy: ClusterFirstWithHostNet
+daemonSet: true
+
+security:
+  podSecurityContext:
+    runAsUser: 0
```
Xartos, line 9: Does it need to do this? I think this was changed for the binding of port 80. But in the end port 8080 is used?

Xartos (commented):
Some comments but overall it looks good

helmfile/values/ambassador.yaml.gotmpl
vanneback: it is still needed for port 443 right? Which we still need to bind.

bootstrap/crds/crds-wc.txt
```diff
@@ -20,3 +20,21 @@ velero/serverstatusrequests.yaml
...
+ambassador/mappings.yaml
+ambassador/modules.yaml
+ambassador/ratelimitservices.yaml
+ambassador/tcpmappings.yaml
+ambassador/tlscontexts.yaml
+ambassador/tracingservices.yaml
+ambassador/project.yaml
+ambassador/projectcontroller.yaml
+ambassador/projectrevision.yaml
+ambassador/ratelimit.yaml
```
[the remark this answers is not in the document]
vanneback: I wrote about it in the description. Im not sure if I should extend it to SC since the task was only for wc? I did add the config option to sc but when i realized i needed to change in all ingresses as well i went back to only wc.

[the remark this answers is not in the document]
Xartos: Aah, I missed that part. But if we are suppose to support it, we should support it for both clusters? Or do we have a usecase for only supporting it in wc?

helmfile/values/ambassador.yaml.gotmpl
Xartos: Didn't it use 8443 for that?

bootstrap/crds/crds-wc.txt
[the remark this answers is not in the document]
tordsson: This issue is scoped only to allowing ck8s end users (wc users) to use Ambassador as ingress (as alternative to nginx). If we want some Ambassador features also in sc is another discussion and issue.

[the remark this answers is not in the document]
Xartos: When we are becoming a OS solution provider, can we think of the sc owner as us? Aren't we also just a customer that uses the project?
I'm just thinking that if a customer wants to run a cluster with ambassador, is it a cluster we manage then? If we are to manage it, can't we decide on what ingress are being used. And if it's a cluster that somebody else is going to manage and they want ambassador on one cluster. I think the chances are big that they want ambassador on the other cluster as well.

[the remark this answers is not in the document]
Xartos: But I guess it doesn't matter if we split it into two steps. But I see little point in doing it since the addition is small to add it in this PR as well

helmfile/values/ambassador.yaml.gotmpl
vanneback: no we still have to create a "module" overwriting the default 8443 with 443.

We discussed using the hostPort in the svc instead but I did not manage to get that working

bootstrap/crds/crds-wc.txt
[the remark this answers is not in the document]
vanneback: I agree with you Fredrik. Not sure i will have the time to fix that as well though. It still requires changing all the charts using ingresses to add an annotation. And testing it of course.
What do you think @tordsson

helmfile/values/ambassador.yaml.gotmpl
Xartos: I'm convinced this should work (since port 8080 worked) and if it can make it not run as root I think it's time well spent. But I guess it depends how much we should polish this.
Is there anyone that actually needs this or is it just so we can claim that we support it?

helmfile/charts/service-cluster-psp-rbac/templates/ambassador-psp-rbac.yaml
```diff
@@ -0,0 +1,33 @@
...
+  kind: Role
+  name: ambassador-privileged
+subjects:
+- kind: ServiceAccount
+  name: ambassador
+  namespace: ambassador
+- kind: ServiceAccount
+  name: default
+  namespace: ambassador
+{{- end }}
```
Xartos, line 33: Not needed since we target wc only

helmfile/charts/service-cluster-psp-rbac/values.yaml
```diff
@@ -1,2 +1,4 @@
 harbor:
   enabled: false
+ambassador:
+  enabled: false
```
Xartos, line 4: Not needed since we target wc only

helmfile/01-applications.yaml
```diff
@@ -58,11 +60,25 @@ releases:
...
+  namespace: ambassador
+  labels:
+    app: ambassador
+  chart: datawire/ambassador
+  version: 6.5.16
+  installed: {{ .Values.ambassador.enabled }}
+  missingFileHandler: Error
+  wait: true
+  values:
+  - values/ambassador.yaml.gotmpl
```
Xartos, line 80: Surround with
`{{ if eq .Environment.Name "workload_cluster" }}`

config/config/sc-config.yaml
```diff
@@ -525,6 +526,9 @@ ingressNginx:
     affinity: {}
     nodeSelector: {}
 
+ambassador:
+  enabled: false
+
```
Xartos, line 531: not needed, WC only

Xartos (changes requested):
Some small things

helmfile/charts/service-cluster-psp-rbac/templates/ambassador-psp-rbac.yaml
vanneback: I added the sc connection aswell.

helmfile/charts/service-cluster-psp-rbac/values.yaml
vanneback: I added sc as well

helmfile/charts/service-cluster-psp-rbac/templates/ambassador-psp-rbac.yaml
Xartos: Right, my bad. I see that sc ingresses have been added.

WIP-CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ See [migrations docs for elasticsearch-exporter](migration/v0.8.x-v0.9.x/migrate
 - Object storage is now optional. There is a new option to set object storage type to `none`. If you disable object storage, then you must also disable any feature that uses object storage (mostly all backups).
 - Two new InfluxDB users to used by prometheus for writing metrics to InfluxDB.
 - Multicluster support for some dashboards in Grafana.
+- Support for ambassador as ingress controller in the workload cluster.
```
Xartos, line 25:
```suggestion
- Support for ambassador as ingress controller in the clusters.
```

OlleLarsson:
@vanneback still relevant?

PR #197, opened 2021-01-21
Added issue templates

Xartos:
To make issues more consistent and easier to write/read I added these templates (github default with minor changes)

part of elastisys/compliantkubernetes-kubespray#13

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,30 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Version (add all relevant versions):**
+ - Compliant kubernetes apps version [e.g. v0.8.0]
+
+**Additional context**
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,30 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
```
viktor-f, line 6: Are we supposed to fill this out or is it just to make it clear that this is a bug report?

If we should fill this out, then I assume that `name` should always have `Bug report` and that `title` will actually describe the issue (e.g. Harbor backup retention is broken). Is that correct?
Is `about` also supposed to just stay as it is, or should we change that to something that describes the bug?

viktor-f (commented):
These seem very nice.

Should we have more templates? Maybe one for investigations.

.github/ISSUE_TEMPLATE/bug_report.md
Xartos: So these values are:
name is the name of the template.
about description of the template.
title is what the title of the issue should be prefilled with
labels is the labels that is automatically assigned to the issue
assignees is who is going to be automatically assigned to the issue
Then the stuff under is what the issue body should be prefilled with.

Does that answer your question? (I didn't really understand your concern)

OlleLarsson (approved):

.github/ISSUE_TEMPLATE/bug_report.md
viktor-f: I mostly just didn't really understand what this block was used for. I though it would also be visible in the issue body. But after checking some other docs I see that it is used to configure the template and will not be added to the issue body (right?).
So then it is good as it is :slightly_smiling_face:

viktor-f (approved):
LGTM!

lentzi90 (approved):

PR #211, opened 2021-01-27
fluentd: Added tail of kubespray default audit log location

Xartos:

Since we want to update the audit path to use kubespray default we need to make sure that both the old and the new way is supported.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/helmfile/values/fluentd-wc.yaml.gotmpl
+++ b/helmfile/values/fluentd-wc.yaml.gotmpl
@@ -50,7 +50,7 @@ extraConfigMaps:
     <source>
       @id kube-audit
       @type tail
-      path /var/log/kube-audit/kube-apiserver.log
+      path /var/log/kube-audit/kube-apiserver.log,/var/log/kubernetes/audit/kube-apiserver-audit.log
       pos_file /var/log/kube-audit/fluentd-kube-apiserver.pos
       tag kubeaudit.*
       read_from_head true
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -5,9 +5,12 @@
 - Several new dashboards for velero, nginx, gatekeeper, uptime of services, and kubernetes status.
 - Metric scraping for nginx, gatekeeper, and velero.
 - Check for Harbor endpoint in the blackbox exporter.
+
 ### Changed
 
 - The falco dashboard has been updated with a new graph, multicluster support, and a link to kibana.
+- Changed path that fluentd looks for kubernetes audit logs to include default path for kubespray.
+
 ### Fixed
 
 - Fixed issue with adding annotation to bootstrap namespace chart
--- a/helmfile/values/fluentd-configmap.yaml.gotmpl
+++ b/helmfile/values/fluentd-configmap.yaml.gotmpl
@@ -136,8 +136,8 @@ forwarder:
       <source>
         @id kube-audit
         @type tail
-        path /var/log/kube-audit/kube-apiserver.log
-        pos_file /var/log/kube-audit/fluentd-kube-apiserver.pos
+        path /var/log/kube-audit/kube-apiserver.log,/var/log/kubernetes/audit/kube-apiserver-audit.log
+        pos_file /var/log/audit/fluentd-kube-apiserver.pos
         tag kubeaudit.*
         read_from_head true
         <parse>
--- a/helmfile/values/fluentd-user.yaml.gotmpl
+++ b/helmfile/values/fluentd-user.yaml.gotmpl
@@ -43,31 +43,6 @@ extraConfigMaps:
       </log>
     </system>
 
-  10-kube-audit.conf: |-
-    <source>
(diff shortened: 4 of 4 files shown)


OlleLarsson:
How about adding multiple paths in the current input [tail plugin](https://docs.fluentd.org/input/tail#path), [fluentd-wc](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/fluentd-wc.yaml.gotmpl#L53), [fluentd-user](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/fluentd-user.yaml.gotmpl#L50), and [fluentd-sc](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/fluentd-configmap.yaml.gotmpl#L139)
```
path /var/log/kube-audit/kube-apiserver.log,/var/log/audit/kube-apiserver-audit.log
```

Just a note, fluentd-user shouldn't really need the kube-audit config as they are not run on master nodes.

Xartos:
> How about adding multiple paths in the current input [tail plugin](https://docs.fluentd.org/input/tail#path), e.g. [fluentd-wc](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/fluentd-wc.yaml.gotmpl#L53), [fluentd-user](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/fluentd-user.yaml.gotmpl#L50), and [fluentd-sc](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/fluentd-configmap.yaml.gotmpl#L139)?
> 
> ```
> path /var/log/kube-audit/kube-apiserver.log,/var/log/audit/kube-apiserver-audit.log
> ```
> 
> Just a note, fluentd-user shouldn't really need the kube-audit config as they are not run on master nodes.

Aah missed that feature! Great!
Right, so that's probably a copy-paste miss I guess then

lentzi90 (approved):
LGTM!
Do you need help testing it?

Xartos:
> LGTM!
> Do you need help testing it?

Yea, that would be great. I'm setting up a kubespray cluster to test this. So if you want to set up or have a "ck8s-cluster cluster" (or I guess a kubespray cluster with the old config) that would be much appreciated

helmfile/values/fluentd-wc.yaml.gotmpl
```diff
@@ -46,31 +46,6 @@ extraConfigMaps:
       </log>
     </system>
 
-  10-kube-audit.conf: |-
```
OlleLarsson, line 49: This should be kept here but removed from fluentd-user :D

OlleLarsson (changes requested):

helmfile/values/fluentd-wc.yaml.gotmpl
Xartos: Ooh, right. Ofcorse

OlleLarsson (approved):

OlleLarsson:
Maybe add a line or two in the wip-changelog about these two changes.

Xartos:
> Maybe add a line or two in the wip-changelog about these two changes.

Yes I'll add it when I've tested it fully

helmfile/values/fluentd-wc.yaml.gotmpl
```diff
@@ -50,7 +50,7 @@ extraConfigMaps:
     <source>
       @id kube-audit
       @type tail
-      path /var/log/kube-audit/kube-apiserver.log
+      path /var/log/kube-audit/kube-apiserver.log,/var/log/audit/kube-apiserver-audit.log
```
lentzi90, line 53: Using kubespray with `audit_log_path: "/var/log/kube-audit/kube-apiserver.log"` makes the API-server Pod mount things like this:
```yaml
volumes:           
  - hostPath:                      
      path: /var/log/kubernetes/audit
...
  volumeMounts:                                                                                                                                                                                                                                                                 
    - mountPath: /var/log/kube-audit                                                                                                                                                                                                                                              
      name: audit-logs
```

So the path *inside the container* is affected, but the path on the host is actually `/var/log/kubernetes/audit/kube-apiserver.log`, which fluentd doesn't read.
After noticing this, I found another variable for kubespray: `audit_log_hostpath: /var/log/kubernetes/audit`. I think this is what we should be looking at and tell fluentd to use. See https://github.com/kubernetes-sigs/kubespray/blob/master/roles/kubernetes/master/defaults/main/main.yml#L68

lentzi90 (changes requested):
Tested and found some issues with mount points/variables.

helmfile/values/fluentd-wc.yaml.gotmpl
Xartos: Updated fluentd path to math the default in kubespray.

lentzi90 (approved):

PR #231, opened 2021-02-10
apps sc: Update opendistro helm chart to 1.12

### LINKED ISSUE #164: [2] Update opendistro to 1.12.0
**Why?**
- To get access to multi role nodes.
- No need for standalone client, master, and data nodes with our application of elasticsearch
  - Makes more sense to either have data/master/client nodes or master/client + data nodes.

**High prio?**
- Yes

**How?**
- Read relase notes
  - Take actions accordingly
- Test the new version out
  - Install in new cluster
  - Upgrade running cluster
  - Confirm functionality
- Update config to allow for all-in-one statefulset or let the user fully decide the topology.
  - Test upgrade path

OlleLarsson:
Upgrades the helm chart which gives us access to creating multirole nodes.

Updates test according to the new settings.

Updates the dev flavor to deploy a **single** multi-role node(kind of just to see if it works well for us or not).

fixes #164 

I've changed the init flow a bit.
The default value for storageclasses for elasticsearch has been changed to `null`, which means that the cluster default will be used. This makes sense imo as it decreases the init-complexity and is in line with most other helm charts.

With the new version there are some repeated deprecation (every 30s) logs in the client nodes
```
[2021-02-10T18:59:20,396][DEPRECATION][o.e.d.c.m.IndexNameExpressionResolver] [opendistro-es-client-85b7d4f6c5-spkvf] this request accesses system indices: [.kibana_1], but in a future major version, direct access to system indices will be prevented by default
```
https://discuss.elastic.co/t/fresh-install-if-elk-7-10-floods-deprecation-log/259972/3
This is not a big deal imo.

The new version of od introduced changes to ism actions which means that the configurer needs additional permissions. This mean that upgrading will require running `securityadmin.sh`. I will add upgrade docs soon.

I'm running a single node in my cluster to see if it works or not
image

With a single node we won't get a "green" cluster state as indicies by default consist of 1 primary and 1 replica shard, and as we only have one data node, the replica shard cannot be assgined anywhere. To me this does not matter as it's just a dev cluster, just thought that iäd mention it.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,4 +1,10 @@
 ### Release notes
+- With the update of the opendistro helm chart you can now decide whether or not you want dedicated deployments for data and client/ingest nodes.
+  By setting `elasticsearch.dataNode.dedicatedPods: false` and `elasticsearch.clientNode.dedicatedPods: false`,
+  The master node statefulset will assume all roles.
+- To get some of the new default values for resource requests on Harbor pods you will first need to remove the resource requests that you have in your Harbor config and then run `ck8s init` to get the new values.
+- Check out the [upgrade guide](migration/v0.9.x-v0.10.x/upgrade-apps.md) for a complete set of instructions needed to upgrade.
+
 
 - Ck8sdash has been deprecated and will be removed when upgrading.
   Some resources like it's namespace will have to be manually removed.
@@ -14,6 +20,11 @@
 
 - The falco dashboard has been updated with a new graph, multicluster support, and a link to kibana.
 - Changed path that fluentd looks for kubernetes audit logs to include default path for kubespray.
+- Opendistro helm chart updated to 1.12.0.
+- Options to disable dedicated deployments for elasticsearch data and client/ingest nodes.
+- By default, no storageclass is specified for elasticsearch, meaning it'll consume whatever is cluster default.
+- Updated elasticsearch config in dev-flavor.
+  Now the deployment consists of a single master/data/client/ingest node.
 
 ### Fixed
 
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,8 +1,7 @@
 ### Release notes
 - With the update of the opendistro helm chart you can now decide whether or not you want dedicated deployments for data and client/ingest nodes.
   By setting `elasticsearch.dataNode.dedicatedPods: false` and `elasticsearch.clientNode.dedicatedPods: false`,
-  The master node statefulset will assume all roles.
-- To get some of the new default values for resource requests on Harbor pods you will first need to remove the resource requests that you have in your Harbor config and then run `ck8s init` to get the new values.
+  the master node statefulset will assume all roles.
 - Check out the [upgrade guide](migration/v0.9.x-v0.10.x/upgrade-apps.md) for a complete set of instructions needed to upgrade.
 
 
--- a/migration/v0.9.x-v0.10.x/upgrade-apps.md
+++ b/migration/v0.9.x-v0.10.x/upgrade-apps.md
@@ -16,30 +16,35 @@ You will need to follow these steps in order to upgrade each Compliant Kubernete
     - `ck8sdash.*`
     - `externalTrafficPolicy.whitelistRange.ck8sdash`
 
+4. Check elasticsearch configuration in sc-config.yaml
+  Make sure that the new `elasticsearch.dataNode.dedicatedPods` and `elasticsearch.clientNode.dedicatedPods` are set to the desired value (true or false) to avoid unintentional changes to the cluster topology.
+
 4. Upgrade workload cluster applications
   ```bash
   ./bin/ck8s apply wc
   ```
 
 5. Upgrade service cluster applications
+  **Note**, during the upgrade, the opendistro roles will be reloaded, so make sure you've got sufficient backups of the ones you've manually added.
+
   ```bash
-
-
   # Upgrade Elasticsearch
   ./bin/ck8s ops helmfile sc -l app=opendistro apply
 
+  # Open another teminal so that you can run kubectl while helmfile is running
+
   # Wait for master pod 0 to be up and for elasticsearch to have started
 
-  # Reload security config
+  # Reload roles.
   ./bin/ck8s ops kubectl sc -n elastic-system exec opendistro-es-master-0 -- chmod +x ./plugins/opendistro_security/tools/securityadmin.sh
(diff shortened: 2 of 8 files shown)


WIP-CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
 ### Release notes
+- With the update of the opendistro helm chart you can now decide whether or not you want dedicated deployments for data and client/ingest nodes.
+  By setting `elasticsearch.dataNode.dedicatedPods: false` and `elasticsearch.clientNode.dedicatedPods: false`,
+  The master node statefulset will assume all roles.
```
lentzi90, line 4:
```suggestion
  the master node statefulset will assume all roles.
```

WIP-CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
...
+- To get some of the new default values for resource requests on Harbor pods you will first need to remove the resource requests that you have in your Harbor config and then run `ck8s init` to get the new values.
```
lentzi90, line 5: Should this be here? It looks unrelated :thinking:

migration/v0.9.x-v0.10.x/upgrade-apps.md
```diff
@@ -0,0 +1,36 @@
+
+# Upgrade v0.9.x to v0.10.0
+
+You will need to follow these steps in order to upgrade each Compliant Kubernetes environment from v0.8.x to v0.9.0. The instructions assume that you work from the root of the `compliantkubernetes-apps` repository, has the latest changes already pulled and configured `CK8S_CONFIG_PATH` correctly.
```
lentzi90, line 4:
```suggestion
You will need to follow these steps in order to upgrade each Compliant Kubernetes environment from v0.9.x to v0.10.0.
The instructions assume that you work from the root of the `compliantkubernetes-apps` repository, has the latest changes already pulled and configured `CK8S_CONFIG_PATH` correctly.
```

lentzi90 (commented):
Awesome!

I like the change to the storage config and I agree that the issues you mentioned seem to be no big deal. To get the cluster to green state, would it be possible to set the number of replicas to 0 or something like that?
One thing I noticed is that we go from 1.10 to 1.12 directly with this change. Is this safe or should we first go to 1.11? I haven't checked how it is for opendistro, but I know some projects don't support skipping over minor versions, while others have "odd versions are dev releases". :smile:

migration/v0.9.x-v0.10.x/upgrade-apps.md
```diff
@@ -0,0 +1,36 @@
...
+   # Wait for master pod 0 to be up and for elasticsearch to have started
+
+   # Reload security config
+   ./bin/ck8s ops kubectl sc -n elastic-system exec opendistro-es-master-0 -- chmod +x ./plugins/opendistro_security/tools/securityadmin.sh
+   ./bin/ck8s ops kubectl sc -n elastic-system exec opendistro-es-master-0 -- ./plugins/opendistro_security/tools/securityadmin.sh \
+       -cd plugins/opendistro_security/securityconfig/ \
+       -icl -nhnv \
+       -cacert config/admin-root-ca.pem \
+       -cert config/admin-crt.pem \
+       -key config/admin-key.pem
```
viktor-f, line 30: This will remove all other (manually added) users like it's done before? Perhaps it could be worth adding a note about that

migration/v0.9.x-v0.10.x/upgrade-apps.md
```diff
@@ -0,0 +1,36 @@
...
+
+5. Upgrade service cluster applications
+  ```bash
+
+   # Upgrade Elasticsearch
+   ./bin/ck8s ops helmfile sc -l app=opendistro apply
+
+   # Wait for master pod 0 to be up and for elasticsearch to have started
+
+   # Reload security config
```
viktor-f, line 23: You need to reload the config in a new terminal right? If so then write that here, I know that I have missed that in the past and waited until the release timed out.

viktor-f (commented):
Looks good :) 

Some minor upgrade comments.

WIP-CHANGELOG.md
viktor-f: Yes, the harbor line is old.

migration/v0.9.x-v0.10.x/upgrade-apps.md
OlleLarsson: I've changed it so that only roles.yml is reloaded instead of everything.
I've also added a note.

migration/v0.9.x-v0.10.x/upgrade-apps.md
OlleLarsson: I've added a note.

Xartos (approved):
LGTM! Great upgrade for all involved!

WIP-CHANGELOG.md
OlleLarsson: Yes, a copy/paster error from my side.

lentzi90 (approved):
LGTM!

viktor-f (approved):
LGTM!

migration/v0.9.x-v0.10.x/upgrade-apps.md
viktor-f: Nice, so rolebindings will stay as well as users?

migration/v0.9.x-v0.10.x/upgrade-apps.md
viktor-f: Thanks :slightly_smiling_face:

migration/v0.9.x-v0.10.x/upgrade-apps.md
OlleLarsson: Yes

OlleLarsson:
> Awesome!
> 
> I like the change to the storage config and I agree that the issues you mentioned seem to be no big deal. To get the cluster to green state, would it be possible to set the number of replicas to 0 or something like that?

Not impossible but it isn't trivial.
It might be possible to create an index template that's applied to all indices that sets `number_of_replicas: 0`.
However, I suspect that there are some problematic (system)indices that will have been created before we can create the required index templates.
It might be possible to configure the aforementioned indices via configuration(elasticsearch.yml) but I haven't looked further into it.

> One thing I noticed is that we go from 1.10 to 1.12 directly with this change. Is this safe or should we first go to 1.11? I haven't checked how it is for opendistro, but I know some projects don't support skipping over minor versions, while others have "odd versions are dev releases".

It should be fine. I've checked the release notes and the upgrade thing on their website and it didn't say anything specific that'd effect us.

OlleLarsson:
Tested upgrading once more, and a fresh install using the defaults from the dev falvor and it seems to work smoothly.

PR #238, opened 2021-02-15
config: Use proxy protocol defaults to false for safespring

### LINKED ISSUE #214: [1] Default for proxy protocol
**Describe the bug**
We don't provide a default value for `useProxyProtocol` even though it should basically always be set to `false` now when using kubespray.

**To Reproduce**
1. Initialize a config repo
2. Note that `ingressNginx.controller.config.useProxyProtocol` is set to `set-me`.

**Expected behavior**
Since `false` is the upstream default and we rarely use anything other, I think the default for us should also be `false`.

 - Compliant kubernetes apps version: v0.9.0

**Additional context**
We previously used the proxy protocol on certain cloud providers when ck8s-cluster set up a HAProxy instance in front of the cluster. This is no longer the case since we are switching to kubespray.

OlleLarsson:
Updates the default value for using the proxy protocol in nginx for safespring.

fixes #214 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -187,14 +187,8 @@ set_nginx_config() {
         exit 1
     fi
     case ${CK8S_CLOUD_PROVIDER} in
-        citycloud | exoscale)
+        citycloud | exoscale | safespring)
           use_proxy_protocol=false
-          use_host_port=true
-          service_enabled=false
-          ;;
-
-        safespring)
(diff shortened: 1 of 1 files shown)


lentzi90 (approved):
LGTM!

viktor-f (approved):

PR #241, opened 2021-02-16
all: Remove ck8sdash

OlleLarsson:

#240 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.


helmfile/01-applications.yaml
```diff
@@ -106,26 +106,6 @@ releases:
...
 {{ end }}
 
-
-- name: ck8sdash
-  namespace: ck8sdash
-  labels:
-    app: ck8sdash
-  chart: ./charts/ck8sdash
-  version: 0.3.1
-  installed: {{ .Values.ck8sdash.enabled }}
```
lentzi90, line 116: Should we keep this and hard code to `installed: false` to make it auto removed from all environments?

lentzi90 (commented):
I see red, I like :slightly_smiling_face:

helmfile/01-applications.yaml
OlleLarsson: Maybe, I wasn't sure whether to do that or write a script that removes everything including the ck8sdash namespace (but maybe that is installed from the rbac chart in most cases).

lentzi90: Right, whatever is easier works for me. :slightly_smiling_face:

viktor-f:
ck8s dash had some credentials in grafana and/or kibana, right? If so then maybe we should add something in the release/upgrade notes about removing them.

helmfile/01-applications.yaml
OlleLarsson: I went with your solution.

OlleLarsson:
> ck8s dash had some credentials in grafana and/or kibana, right? If so then maybe we should add something in the release/upgrade notes about removing them.

Yes, there were some api keys in grafana. I've added removal instructions in the upgrade guide.

lentzi90 (approved):
LGTM!

Have you tested that helmfile can remove ck8sdash also with the chart files already gone? Do you need any help testing?

viktor-f (approved):
LGTM!

OlleLarsson:
> LGTM!
> 
> Have you tested that helmfile can remove ck8sdash also with the chart files already gone? Do you need any help testing?

Yes, I've tested (I suppose it only needs to know the name of the release in order to remove it)

PR #245, opened 2021-02-17
apps sc: fix fluentd configmap gcs keyfile indentation

### LINKED ISSUE #244: Fix fluentd ConfigMap gcs keyfile indentation
**Describe the bug**
[The fluentd configmap has an incorrect indentation value](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/fluentd-configmap.yaml.gotmpl#L68) which causes an error if you try to use gcs as object storage. The indentation value should be 8.

**To Reproduce**
1. Set `objectStorage.type` to `"gcs"` in `sc-config.yaml` in your config path
2. Run bin/ck8s apply sc

**Expected behavior**
Apps are applied without errors.

 - Compliant kubernetes apps v0.9.0

Ajarmar:
**What this PR does / why we need it**: Fixes an indentation error in the fluentd aggregator ConfigMap which caused problems when using gcs as object storage.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #244 

**Special notes for reviewer**: I feel like this is so minor that it didn't warrant adding to the changelog, but tell me if you think I should do that.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/helmfile/values/fluentd-configmap.yaml.gotmpl
+++ b/helmfile/values/fluentd-configmap.yaml.gotmpl
@@ -65,7 +65,7 @@ aggregator:
       </match>
     {{ if eq .Values.objectStorage.type "gcs" -}}
     gcp_credentials.json: |
-      {{ .Values.objectStorage.gcs.keyfileData | nindent 4 }}
+      {{ .Values.objectStorage.gcs.keyfileData | nindent 6 }}
     {{- end }}
 
   plugins:


helmfile/values/fluentd-configmap.yaml.gotmpl
```diff
@@ -65,7 +65,7 @@ aggregator:
       </match>
     {{ if eq .Values.objectStorage.type "gcs" -}}
     gcp_credentials.json: |
-      {{ .Values.objectStorage.gcs.keyfileData | nindent 4 }}
+      {{ .Values.objectStorage.gcs.keyfileData | nindent 8 }}
```
OlleLarsson, line 68: Maybe?
```suggestion
      {{ .Values.objectStorage.gcs.keyfileData | indent 6 }}
```

Ajarmar: I thought I remembered some reason why it had to be 8 rather than 6, but I tried it with 6 now and it works so it would make sense to change that. Using `indent` instead of `nindent` doesn't work for me though, I'm not sure what the difference is between the two?

OlleLarsson: Okay, if it works it's fine with me.
`nindent` prepends a newline to the string, http://masterminds.github.io/sprig/strings.html

Xartos: If you use `indent` it will use your current indentation plus add indentation for the first line. Therefore I think we should use `nindent` to make sure that the whole block is using the same indentation.

But `nindent 6` should be enough, but then again 8 should work as well. (not 4 though)

Xartos (approved):
LGTM!

OlleLarsson (approved):

PR #271, opened 2021-02-23
Validate issuers email

cristiklein:

This PR adds validation for emails set to Issuers. Previously, setting an invalid email address lead to hard-to-find issues, like Certificates not becoming ready, because Issuers are not ready, because Issuer emails are incorrect. This PR makes the `apply` step fail. 

N/A

**Checklist:**

- [NA?] Added relevant notes to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md)
- [NA] Updated the [public facing documentation](https://github.com/elastisys/compliantkubernetes)
    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/helmfile/charts/issuers/templates/NOTES.txt
+++ b/helmfile/charts/issuers/templates/NOTES.txt
@@ -1,0 +1,1 @@
+{{ include "issuers.validateValues" . }}
--- a/helmfile/charts/issuers/templates/_helpers.tpl
+++ b/helmfile/charts/issuers/templates/_helpers.tpl
@@ -11,3 +11,36 @@ Expand the name of the chart.
 {{- define "provisioner.name" -}}
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
+
+{{/*
+Validate values of Issuer Email:
+- must provide a valid email to LetsEncrypt
+*/}}
+{{- define "issuers.validateValues.letsencrypt.prod.email" -}}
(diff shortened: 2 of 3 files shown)


lentzi90 (commented):
Cool!
I'm not sure what to do about the unit tests though... Either we would have to provide a "valid" email by default, which doesn't feel right, or we need to change how we test. :thinking:

OlleLarsson:
Is the only requirement that the mail address follows correct specification or does that mail address actaully need to exist?

lentzi90:
I'm not sure. But I guess the real problem here is that we want to *both* help the user detect the problem by failing if they don't change the value, and run unit tests that detect problems with the charts. With the current way of testing this won't work, as any value that would pass the test would also be possible to install without the user changing it.

Maybe we could have some values for testing that are not just the defaults? Looks like we already [change some of them](https://github.com/elastisys/compliantkubernetes-apps/blob/main/pipeline/init-unit-tests.bash#L17).

cristiklein:
@lentzi90 Excellent! That's the stuff I was looking for.

lentzi90 (approved):
LGTM!
I would not like too much logic in the `init-unit-test.bash` script though so if we need to add similar things for all the charts then I think it would be better to have a prepared config file instead of updating it in this way.

viktor-f (approved):
LGTM

cristiklein:
> LGTM!
> I would not like too much logic in the `init-unit-test.bash` script though so if we need to add similar things for all the charts then I think it would be better to have a prepared config file instead of updating it in this way.

Agreed!

Issue #242, opened 2021-02-16
[5] Spike: Show us how to customize Prometheus alerts and Grafana dashboards using Jsonnet


cristiklein:
**Context**

As part of the discussion on how to customize Prometheus alerts and Grafana dashboards, we stumbled across this [community proposal](https://docs.google.com/document/d/12CThd7eGiJHXd7oe7LPNBjw5Lv0fv-NkgVvW_2WGzAM/edit?usp=sharing). This seems to be [already implemented](https://github.com/prometheus-operator/kube-prometheus/blob/master/docs/developing-prometheus-rules-and-grafana-dashboards.md). The tech stack is based on [jsonnet](https://jsonnet.org/).

**What should be investigated.**

1. Focus on the service cluster (re-use our existing workload cluster setup). Do allow yourself to change the Prometheus remote writer on the workload cluster if it simplifies the investigation.
2. Disable alerts and dashboards in `compliantkubernetes-apps`. In particular, keep cert-manager and ingress-nginx. (Harbor, Elasticsearch and Kibana can be thrown out too, if it simplifies the investigation.)
3. Show us how to set up Prometheus/Grafana with the setup suggested upstream.
4. Show us what this setup includes by default.
5. Show us how to customize this setup, by adding an alert and a dashboard.

**What artifacts should this produce.**

1. Produce something you can demo in front of the whole team during a 50-minutes time-slot, including questions.
2. Produce a tutorial on how to do steps 3-5. (Markdown with copy-paste-able code in the tutorials repo.)
3. Produce a short video (screencast with voice-over) showing 3-5.

cristiklein:
Might be relevant for the investigation:
- https://github.com/bitnami/kube-prod-runtime/blob/master/manifests/components/prometheus.jsonnet
- https://github.com/bitnami-labs/kubernetes-grafana-dashboards

lentzi90:
See the blog post merged here: https://github.com/elastisys/tutorials-and-blogposts/pull/26

Issue #185, opened 2021-01-20
[2] Make (user-)grafana upgrades work


OlleLarsson:
When we are upgrading grafana, the pod in the new replicaset will be stuck in init-phase becuase the old grafana pod is still running and holding the volume making the new pod unable to start.

```
Events:
  Type     Reason              Age    From                     Message
  ----     ------              ----   ----                     -------
  Normal   Scheduled           9m     default-scheduler        Successfully assigned monitoring/user-grafana-584f4cbfbc-m7xtl to release-test-service-cluster-worker-1
  Warning  FailedAttachVolume  9m1s   attachdetach-controller  Multi-Attach error for volume "pvc-568dd1d4-fcea-4a94-b0dc-8e35ecb3e51c" Volume is already used by pod(s) user-grafana-5f45bb8c48-prnqj
  Warning  FailedMount         6m58s  kubelet                  Unable to attach or mount volumes: unmounted volumes=[storage], unattached volumes=[sc-dashboard-provider storage user-grafana-token-8g745 sc-dashboard-volume config]: timed out waiting for the condition
  Warning  FailedMount         4m45s  kubelet                  Unable to attach or mount volumes: unmounted volumes=[storage], unattached volumes=[storage user-grafana-token-8g745 sc-dashboard-volume config sc-dashboard-provider]: timed out waiting for the condition
  Warning  FailedMount         2m27s  kubelet                  Unable to attach or mount volumes: unmounted volumes=[storage], unattached volumes=[sc-dashboard-volume config sc-dashboard-provider storage user-grafana-token-8g745]: timed out waiting for the condition
  Warning  FailedMount         9s     kubelet                  Unable to attach or mount volumes: unmounted volumes=[storage], unattached volumes=[user-grafana-token-8g745 sc-dashboard-volume config sc-dashboard-provider storage]: timed out waiting for the condition
```

DOD: it is possible to upgrade grafana without manual intervention.

tordsson:
Simply solved by not doing rolling upgrades? Can this be upstreamed?

cristiklein:
This might be fixed in the latest version of the Chart: https://github.com/grafana/helm-charts/commit/855fde1551e5f9f9e42940bdf9ddd41574cee643

lentzi90:
[...]

PR #272, opened 2021-02-23
Fix user Grafana Deployment

cristiklein:

Set deploymentStrategy to Recreate for user Grafana, so that upgrades
work. A rollingUpdate strategy fails, due to persistence being set to
ReadWriteOnce.

Treat this as a temporary fix for #185. A permanent fix will likely come
with #242.

**Checklist:**

    - [X] minor downtime of user Grafana is expected.
    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,5 +1,6 @@
 ### Fixed
 - Fixed service cluster log retention using the wrong service account.
+- Fixed upgrade of user Grafana.
 
 ### Changed
 - Bumped `helm` to `v3.5.2`.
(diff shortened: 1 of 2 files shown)


lentzi90 (approved):
LGTM!

OlleLarsson (approved):

viktor-f:
It seems like I'm a bit late. But this will make grafana create a new PV when it creates a new pod, right?
In that case you would lose all user data that has been added. Primarily dashboards that the users have added and any manual changes to permissions that we have done. I don't think that is what we should go for. But I don't really have a better fix for it right now other than that you can delete the old pod during any upgrade, then the new pod can mount the PV.

cristiklein:
@viktor-f The newly created Pod should attach to the existing PV with zero data loss. Why would you suspect that a new PV is created?

viktor-f:
My bad, didn't read properly. Though it was the PVC policy that was changed not deploymentStrategy. So just ignore what I said :smile:

PR #306, opened 2021-03-08
Updated release documentation to reset CHANGELOG before QA

Xartos:

From the last release I updated the documentation to reflect the flow.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/release/README.md
+++ b/release/README.md
@@ -8,40 +8,56 @@ https://semver.org/
 1. To release a major or minor version create a release branch `release-X.Y` from the last minor release tag.
 
     ```bash
-    git checkout vA.B.0
+    git checkout main
     git checkout -b release-X.Y
     git push -u origin release-X.Y
     ```
 
-2. Checkout the branch to cut the release from.
+1. Reset changelog
 
     ```bash
-    git checkout main
-    git pull
-    git checkout -b branch_name
-    ```
-
-3. Run QA checks on `branch_name` and commit any fixes found during QA.
-
-4. Run the release script.
-
-    ```bash
-    ./release/release.sh X.Y.Z
+    git checkout -b reset-changelog-X.Y
+    release/reset-changelog.sh X.Y.0
     ```
 
     The release script will:
     * Append what is in `WIP-CHANGELOG.md` to `CHANGELOG.md`
     * Clear `WIP-CHANGELOG.md`
-    * Create a git commit with message `Release vX.Y.Z`
+    * Create a git commit with message `Reset changelog for release vX.Y.Z`
+
+    Make sure that the changes only include changes mentioned above, and then push.
+
+    ```
+    git diff HEAD~1
+    git push
+    ```
+
+1. Merge `reset-changelog-X.Y` into `release-X.Y` then, into `main` as soon as possible to minimize risk of conflicts
+
+    **NOTE**: The release action will fail since we haven't tagged the release commit.
+    We will do that after QA.
+
+1. Run QA checks on `branch_name` and commit any fixes found during QA into `release-X.Y`.
+
+    **NOTE**: All changes made in QA should be added to `CHANGELOG.md` and **NOT** `WIP-CHANGELOG.md`.
+
+1. Run the release script.
+
+    ```bash
+    ./release/release.sh X.Y.0
+    ```
+
+    The release script will:
+    * Amend last commit and prepend message `Release vX.Y.Z`
     * Create a tag named `vX.Y.Z`
 
-5. Push the tagged commit, create a PR against the release branch and request a review.
+1. Push the tagged commit, create a PR against the release branch and request a review.
 
     ```bash
     git push --atomic origin branch_name vX.Y.Z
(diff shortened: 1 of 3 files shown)


release/README.md
```diff
@@ -8,40 +8,53 @@ https://semver.org/
...
-    git checkout -b branch_name
+    git checkout -b reset-changelog-X.Y
+    release/reset-changelog.sh X.Y.0
     ```
 
-3. Run QA checks on `branch_name` and commit any fixes found during QA.
+    The release script will:
+    * Append what is in `WIP-CHANGELOG.md` to `CHANGELOG.md`
+    * Clear `WIP-CHANGELOG.md`
+    * Create a git commit with message `Release vX.Y.Z`
```
lentzi90, line 26: Won't this look strange if we add fixes after it while doing QA?

Git log:
```
321cba (tag: v1.0) Fix issue found in QA <-- The tagged release
abc123 Release v1.0 <-- from resetting the changelog
```

release/README.md
```diff
@@ -8,40 +8,53 @@ https://semver.org/
 1. To release a major or minor version create a release branch `release-X.Y` from the last minor release tag.
 
     ```bash
-    git checkout vA.B.0
+    git checkout main
```
lentzi90, line 11: Should we really branch from `main`? I think the idea behind branching from the previous release is to be able to determine what changed between them. The release is then a PR that will always tell you exactly what changed and you can check every commit if you'd like. If we start from `main` there is no such PR, just a tag that you can diff with the last tag.

The issue I see with this is that fixes we add during QA could be "doubled", but maybe I'm missing something?
I mean we add it during QA to the release, but then we also merge it to main, so the next time we branch from main that fix will be there again, including the note about the fix that was already in the last release.

I'm confusing myself :sweat_smile: but here is an attempt to describe it:

1. v1.0 is released and we start working on the next version on main, adding changes to the WIP change log.
2. We start preparing for v1.1 by branching from main and clearing the WIP change log.
3. We add a fix to the soon-to-be release that also requires a change to the (WIP?) change log.
4. Release v1.1
5. Merge the fix back to main, adding the change log note
6. Start working on v1.2. Now the change log contains the note about that fix that was already in the last release.

lentzi90 (commented):
Looks quite nice but I'm confusing myself with how fixes during QA would work. See questions below. :smile:

release/README.md
Xartos: Yea, that's true. Maybe we could change the message to something like `Cut for release vX.Y.Z ` and add then add another commit after QA which the tag is set to and that has the message `Release vX.Y.Z`

lentzi90: Sounds good! Maybe rather `Clear WIP-CHANGELOG.md` or `Clear WIP-CHANGELOG.md for vX.Y.Z`  as message?

release/README.md
Xartos: I don't really see the difference other than in one you get the diff in the PR the other you need to run `git diff ${PREV_VERSION}` the history should be exactly the same in the end.

Maybe we should have QA fixes into WIP changelog. But I thought we should add them to the CHANGELOG directly (since they are part of the release. Then when we merge the fixes they should end up directly in the changelog and nothing gets wierd.
Because otherwise, if we have it like we have it now. When we merge into main we will always have a conflict in WIP if there's more work added since the branching and you need to manually figure out what has been added and not. I just feel like this flow would produce less merge conflicts.

lentzi90: > But I thought we should add them to the CHANGELOG directly (since they are part of the release. Then when we merge the fixes they should end up directly in the changelog and nothing gets wierd.

Ah this was the part I was missing and a really nice solution! Could you mention it maybe in the document as well? Otherwise I'm sure I'll add something in the wrong change log. :smile:

release/README.md
Xartos: Apparently I've already done this in the script but just forgot to change here
https://github.com/elastisys/compliantkubernetes-apps/pull/306/files#diff-a36345fe16d00b329e46048d709e1a54a6fea61245d0f56bef4a91b4e053e6ccR68
But I should fix the release script so it adds the release commit.

lentzi90 (approved):
LGTM!

release/README.md
```diff
@@ -8,40 +8,56 @@ https://semver.org/
...
+    git checkout -b reset-changelog-X.Y
+    release/reset-changelog.sh X.Y.0
+    ```
+
+    The release script will:
+    * Append what is in `WIP-CHANGELOG.md` to `CHANGELOG.md`
+    * Clear `WIP-CHANGELOG.md`
+    * Create a git commit with message `Reset changelog for release vX.Y.Z`
+
+    Make sure that the changes only include chanes mentioned above, and then push
```
OlleLarsson, line 28:
```suggestion
    Make sure that the changes only include changes mentioned above, and then push.
```

OlleLarsson (approved):

Issue #207, opened 2021-01-26
[2] Add configuration option to "user api server" status in blackbox exporter


viktor-f:
**Is your feature request related to a problem? Please describe.**
The current configuration assumes that the workload cluster api server is available at `https://kube-apiserver.{{ .Values.global.baseDomain }}:/healthz`. This is a problem primarily if you are running multiple workload clusters, then blackbox exporter will at best monitor one of them. If the apiserver is not connected to this URL, then the metric will say that the apiserver is failing and the uptime dashboard will also show this (even though the apiserver(s) might work as expected).

**Describe the solution you'd like**
Add config that allows us to specify any amount of apiservers that blackbox should check.
Also make sure that the uptime dashboard can show the status of multiple apiservers. 

**Describe alternatives you've considered**
The only other alternative I see is to have an option to disable this metric, so that it at least does not give any misinformation.

**Additional context**
This is how the uptime dashboard looks now in my setup.
![broken-blackbox](https://user-images.githubusercontent.com/58632240/105812403-5fbb7100-5fae-11eb-965b-0adc8bae12bf.png)

**Definition of done:**
There is config that allows us to specify any number of apiservers to check.
The uptime dashboard can show uptime for multiple apiservers.

lentzi90:
As part of this issue, please also make blackbox exporter [verify TLS](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/blackbox.yaml.gotmpl#L23) based on the [global config](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/config/sc-config.yaml#L8).

tordsson:
Will be replaced by: https://github.com/elastisys/ck8s/issues/102

PR #310, opened 2021-03-10
Use ClusterIssuer in SC and disable in WC

### LINKED ISSUE #287: [2] Switch to ClusterIssuers

Are new service cluster looks as follows:

```
$ kubectl get issuers,clusterissuers -o wide -A
NAMESPACE             NAME                                              READY   STATUS                                                 AGE
elastic-system        issuer.cert-manager.io/letsencrypt-prod           True    The ACME account was registered with the ACME server   4m2s
elastic-system        issuer.cert-manager.io/letsencrypt-staging        True    The ACME account was registered with the ACME server   4m2s
elastic-system        issuer.cert-manager.io/opendistro-es-ca           True    Signing CA verified                                    4m
elastic-system        issuer.cert-manager.io/opendistro-es-selfsigned   True                                                           4m
influxdb-prometheus   issuer.cert-manager.io/letsencrypt-prod           True    The ACME account was registered with the ACME server   4m2s
influxdb-prometheus   issuer.cert-manager.io/letsencrypt-staging        True    The ACME account was registered with the ACME server   4m2s
kube-system           issuer.cert-manager.io/letsencrypt-staging        True    The ACME account was registered with the ACME server   4m2s
monitoring            issuer.cert-manager.io/letsencrypt-staging        True    The ACME account was registered with the ACME server   4m2s
```

This feels fragile and wasteful. Can we use two ClusterIssuers instead, `letsencrypt-prod` and `letsencrypt-staging`?

* A working demo with ClusterIssuers or a reason why this is a bad idea.

lentzi90:
See #287 for why we want to use ClusterIssuers. I also disabled the (Cluster)Issuers in WC since we don't really need them there. We only used them with kubeapi-metrics for health checking the API server. But this check [doesn't need a trusted certificate](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/blackbox.yaml.gotmpl#L23).

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #287

A nicer solution would be to make the blackbox exporter use `tls_verify` depending on what issuer we use and add a config option to specify [extra annotations for kubeapi-metrics](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/charts/kubeapi-metrics/values.yaml#L14) so that we can [use an Issuer](https://github.com/elastisys/compliantkubernetes-apps/compare/lj/use-cluster-issuer?expand=1) instead of a ClusterIssuer in WC. This is a bit larger change though and I wasn't sure if it was worth it (especially considering the config overhaul and that blackbox exporter doesn't verify TLS). Should I go for it anyway?

**Checklist:**

    - [x] will change any APIs.
    - [ ] will break the cluster.

**Pipeline config** :heavy_check_mark:

### DIFF
--- a/config/config/wc-config.yaml
+++ b/config/config/wc-config.yaml
@@ -384,9 +384,6 @@ issuers:
   letsencrypt:
     enabled: true
 
-    ## Namespaces in which to install the letsencrypt issuers in.
-    namespaces: []
-
     prod:
       ## Mail through which letsencrypt can contact you.
       email: "set-me"
--- a/config/config/wc-config.yaml
+++ b/config/config/wc-config.yaml
@@ -382,14 +382,13 @@ issuers:
   ## Deploy let's encrypt ACME issuers
   ## "letsencrypt-prod" and "letsencrypt-staging".
   letsencrypt:
-    enabled: true
-
-    prod:
-      ## Mail through which letsencrypt can contact you.
-      email: "set-me"
-    staging:
-      ## Mail through which letsencrypt can contact you.
-      email: "set-me"
+    enabled: false
+    # prod:
+    #   ## Mail through which letsencrypt can contact you.
+    #   email: ""
+    # staging:
+    #   ## Mail through which letsencrypt can contact you.
+    #   email: ""
 
   ## Additional issuers to create.
   ## ref: https://cert-manager.io/docs/configuration/
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,5 +1,15 @@
+### Release notes
+
+- ClusterIssuers are used instead of Issuers.
+  Administrators should be careful regarding the use of ClusterIssuers in workload clusters, since users will be able to use them and may cause rate limits.
+- Check out the [upgrade guide](migration/v0.11.x-v0.12.x/upgrade-apps.md) for a complete set of instructions needed to upgrade.
+
 ### Added
 
 - NetworkPolicy dashboard in Grafana
 - Added a new helm chart `calico-accountant`
 - Clean-up scripts that can remove compliantkubernetes-apps from a cluster
+
+### Changed
+
+- ClusterIssuers are used instead of Issuers
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -298,34 +298,6 @@ generate_secrets() {
     sops_encrypt "$file"
 }
 
-# Usage: set_issuer_namespaces <config_file> service_cluster|workload_cluster
-set_issuer_namespaces() {
-    if [[ $# -ne 2 ]]; then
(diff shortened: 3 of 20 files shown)


config/config/wc-config.yaml
```diff
@@ -382,17 +382,13 @@ issuers:
...
-    prod:
-      ## Mail through which letsencrypt can contact you.
-      email: "set-me"
-    staging:
-      ## Mail through which letsencrypt can contact you.
-      email: "set-me"
+    enabled: false
+    # prod:
+    #   ## Mail through which letsencrypt can contact you.
+    #   email: null
```
OlleLarsson, line 388: Just tested and this did not work as expected. Using `""` instead gives a warning.

OlleLarsson: With 
```ShellSession
issuers:
  letsencrypt:
    prod:
      email: ""
    staging:
      email: null
```

yq returns nothing/null in bash with `email: ""`, whereas it return the string `null` when `email: null` :facepalm:
Leaving the value empty for a field also works `email:` as yq will return empty string.

lentzi90: Oh good to know. I'll change the comment to have `""` instead of `null` then.

OlleLarsson: This bothers me as all of these should be equivalent as according to the [YAML specification](https://yaml.org/spec/1.2/spec.html#id2805071)
```ShellSession
Regular expression         Resolved to tag
null | Null | NULL | ~     tag:yaml.org,2002:null
/* Empty */                tag:yaml.org,2002:null
```
```
parent:
  key1:               # empty so "null", # is a comment!
  key2: ~             # also "null"
  key3: null          # "null" explicitly ))
  key4: !!null "null" # for the funs of "secondary tag handle: !!"
  key5: "null"        # sorry, it is a string or !!str if you like ((
```

OlleLarsson:
This looks ok to me.
I'm just a bit worried that we need to use a clusterissuer for user-alertmanager if that's enabled, but i guess if we explain it to the user that they shouldn't use it for their certificates then i think we'll be fine.

Concerning your question about the blackbox api server healthcheck i'd say that it is alright to skip right now and maybe do it in a separate pr.

lentzi90:
Thanks for the comments!
I created an issue for [Issuer/ClusterIssuer](https://github.com/elastisys/compliantkubernetes-apps/issues/311) and mentioned in [this issue](https://github.com/elastisys/compliantkubernetes-apps/issues/207) for the blackbox exporter that we should also verify TLS based on some variable instead of always skipping it.

OlleLarsson (approved):

viktor-f (approved):

Issue #300, opened 2021-03-03
[3] Make apps less fragile


cristiklein:
**Describe the bug**

Setting up the apps feels fragile. `./bin/ck8s apply sc` does not always return with a fully working service cluster. I bump into issues like:

- Issuers not being ready, before Ingresses are created, leading to TLS problems.
- Kibana ending up not being configured correctly.

**To Reproduce**

1. `./scripts/clean-sc.sh`
2. `./bin/ck8s apply sc`
3. `./bin/ck8s ops kubectl get issuers,certificates,orders,certificaterequests -o wide`
4. Go to `grafana.$DOMAIN`, `kibana.$DOMAIN`

**Expected behavior**

Either everything works, or documentation is provided with how to set up everything layer by layer. Each layer should consist of commands to "apply; wait; test".

**Screenshots**

Darn, should have taken a few. Will provide next time I bump into this.

**Version (add all relevant versions):**
 - Compliant kubernetes apps version 91253f9bd48c20383bd2ea0758b8531d0d64bc19

cristiklein:
Food for thought: Some defaults could be changed to make it easier to set up CK8s:
1. Many things break without a proper S3 access key and secret. Should we rather default to no object storage?
2. Should we provide a "default" email address for LetsEncrypt?

cristiklein:
More thoughts: Is the bootstrap step really necessary, or could we rely on standard Helm 3 mechanism to set up CRDs and namespaces.

cristiklein:
This is the kind of error that @jakubkrzywda gets, likely due to wrong ordering of Releases. A webhook is not fully started, before a concurrent Release gets installed:

```
==> amazon-ebs: UPDATED RELEASES:
==> amazon-ebs: NAME                    CHART                                        VERSION
==> amazon-ebs: wc-reader               charts/prometheus-instance                     0.1.0
==> amazon-ebs: node-local-dns          charts/node-local-dns                          0.1.0
==> amazon-ebs: grafana-ops             charts/grafana-ops                             0.1.0
==> amazon-ebs: blackbox                charts/blackbox                                0.1.0
==> amazon-ebs: wc-reader-alerts        charts/prometheus-alerts                       0.1.0
==> amazon-ebs: sc-alerts               charts/prometheus-alerts                       0.1.0
==> amazon-ebs: influxdb                elastisys/influxdb                            4.8.12
==> amazon-ebs: metrics-server          stable/metrics-server                         2.10.0
[...]

PR #327, opened 2021-03-16
bin: Update config validation

OlleLarsson:
To lessen the likelihood that a user deploys apps with a bad config.

Before our validation gave warning on `""` and empty/missing paths.
Now it has been update to warn on the following
- `"set-me"`
- `set-me`
- and empty/missing path/field.

This means that if there is a field that we want the user to configure it should prefereably be set to `"set-me"` by default.

part of #300 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/bin/common.bash
+++ b/bin/common.bash
@@ -142,8 +142,8 @@ validate_config() {
         # available in $1
         maybe_exit="false"
         for opt in ${options}; do
-            value=$(yq r "${file}" "${opt}")
-            if [[ -z "$value" ]]; then
+            value=$(yq r --unwrapScalar=false "${file}" "${opt}")
+            if [[ -z "${value}" ]] || [[ "${value}" = '"set-me"' ]] || [[ "${value}" = "set-me" ]]; then
                 log_warning "WARN: ${opt} is not set in ${file}"
                 maybe_exit="true"
             fi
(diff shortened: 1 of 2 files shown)


Xartos (approved):
LGTM

bin/common.bash
```diff
@@ -142,8 +142,8 @@ validate_config() {
         # available in $1
         maybe_exit="false"
         for opt in ${options}; do
-            value=$(yq r "${file}" "${opt}")
-            if [[ -z "$value" ]]; then
+            value=$(yq r --unwrapScalar=false "${file}" "${opt}")
+            if [[ -z "${value}" ]] || [[ "${value}" = '"set-me"' ]] || [[ "${value}" = "set-me" ]]; then
```
cristiklein, line 146: Nice!

config/config/sc-config.yaml
```diff
@@ -502,7 +502,7 @@ externalTrafficPolicy:
     prometheusWc: false
 
 nfsProvisioner:
-  server: ""
+  server: "set-me"
```
cristiklein, line 505: I'm a bit unsure about this. This means that I, as a user, I will get an error even if I don't use the NFS storage provider, right? Maybe a validation inside the Helm Chart would be more appropriate?

OlleLarsson: Yes, you'll see a warning.
That is one of the drawbacks with the current validation technque that we are employing, but if you want i can revert back to `""`. Nobody should use an nfs server anymore.

cristiklein: Let's keep `""` for now and pick up the discussion on whether we want to finally move the storage providers out.

cristiklein (approved):

viktor-f (approved):

PR #346, opened 2021-03-23
Updated release documentation

Xartos:

Updated the release documentation because we found some flaws in it.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: None

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.


release/README.md
```diff
@@ -37,52 +37,55 @@ https://semver.org/
...
 
-    The release script will:
-    * Amend last commit and prepend message `Release vX.Y.Z`
-    * Create a tag named `vX.Y.Z`
-
 1. Push the tagged commit, create a PR against the release branch and request a review.
 
     ```bash
-    git push --atomic origin branch_name vX.Y.Z
+    git push --atomic origin QA-X.Y vX.Y.0
```
viktor-f, line 60: Not sure what the difference is, but I used `git push --tags` when I created the 0.12 tag. Would that be better, worse or does it not matter?

viktor-f (commented):
Looks good.
I have one question (see comment).
Also to make things even better to copy-paste, change line 32 to `git push -u origin reset-changelog-X.Y`

release/README.md
Xartos: I think this is a leftover from previous release flow. I think this is because we pushed directly to the release branch and it would start the release workflow directly and we wanted to be sure that the tag and the commit was pushed.
So I guess we can change to `git push --tags QA-X.Y` now

viktor-f (approved):

OlleLarsson (approved):

PR #140, opened 2020-12-15
apps wc: Add admission.gatekeeper.sh/ignore label to kube-system and gatekeeper-system namespaces

jakubkrzywda:
**What this PR does / why we need it**: This PR adds `admission.gatekeeper.sh/ignore="true"` label to `kube-system` and `gatekeeper-system` namespaces.

**Which issue this PR fixes**:
fixes #148 

**Public facing documentation PR** *(if applicable)*

**Special notes for reviewer**:

**Checklist:**

- [ ] Added relevant notes to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md)
- [x] Proper commit message prefix on all commits
- [ ] Updated the [public facing documentation](https://github.com/elastisys/compliantkubernetes)
- Is this changeset backwards compatible for existing clusters? Applying:
    - [x] is completely transparent, will not impact the workload in any way.
    - [ ] requires running a migration script.
    - [ ] will create noticeable cluster degradation.
          E.g. logs or metrics are not being collected or Kubernetes API server
          will not be responding while upgrading.
    - [ ] requires draining and/or replacing nodes.
    - [ ] will change any APIs.
          E.g. removes or changes any CK8S config options or Kubernetes APIs.
    - [ ] will break the cluster.
          I.e. full cluster migration is required.

PR #365, opened 2021-04-07
Petter od authlog

pettersv:
The authlog (/var/log/authlog) contains information about failed access attempts to the nodes. This PR tails the authlog, formats the message and sends it to opendistro.

**NOTE that in order for this to work you need to use version 1.12.1 or later of the elastisys/opendistro-es chart** 

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #140

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/helmfile/values/fluentd-user.yaml.gotmpl
+++ b/helmfile/values/fluentd-user.yaml.gotmpl
@@ -62,6 +62,19 @@ extraConfigMaps:
         </pattern>
       </parse>
     </source>
+    # Tail the authlog
+    <source>
+      @id authlog
+      @type tail
+      path /var/log/auth.log
+      pos_file /var/log/auth.pos
+      tag authlog.*
+      <parse>
+        @type regexp
+        expression /^(?<time>[^ ]* {1,2}[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[a-zA-Z0-9_\/\.\-]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$/
+      </parse>
+    </source>
+
 
     # Detect exceptions in the log output and forward them as one log entry.
     <match raw.kubernetes.**>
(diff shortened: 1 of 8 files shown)


viktor-f:
As far as I can see this looks good. I guess you also need to add a policy for the new index here: https://github.com/elastisys/opendistro-build/tree/main/helm/opendistro-es/files/configurer/ism-policies

helmfile/values/fluentd-user.yaml.gotmpl
```diff
@@ -62,6 +62,19 @@ extraConfigMaps:
         </pattern>
       </parse>
     </source>
+    # Index the authlog
```
OlleLarsson, line 65:
```suggestion
    # Tail authlog
```

helmfile/values/fluentd-user.yaml.gotmpl
```diff
@@ -62,6 +62,19 @@ extraConfigMaps:
...
+    <source>
+      @id authlog
+      @type tail
+      path /var/log/auth.log
+      pos_file /var/log/auth.pos
+      tag authlog.*
```
OlleLarsson, line 71: We could do without the `.*` if we want. All it does is add the add the dotnotation path, e.g. events will be emitted with tag `authlog.var.log.auth.log`.

helmfile/values/fluentd-user.yaml.gotmpl
```diff
@@ -120,6 +133,44 @@ extraConfigMaps:
     @include /etc/fluent/extra-config.d/*.conf
 
   output.conf: |-
+    <match authlog.**>
+       @id elasticsearch_authlog
+       @type elasticsearch
+       @log_level "#{ENV['OUTPUT_LOG_LEVEL']}"
+       include_tag_key true
```
OlleLarsson, line 140: Seems kind of unessecary to include the tag as a field.

OlleLarsson (commented):
Good job.
Does the logs inlcude any timestaps?
I'm wondering if we should use that as [time_key](https://github.com/uken/fluent-plugin-elasticsearch#time_key) or not :thinking:

helmfile/values/fluentd-user.yaml.gotmpl
pettersv: Fixed :)

helmfile/values/fluentd-user.yaml.gotmpl
pettersv: We seem to use dotnotation for the other indicies so I'm leaving it for now.

helmfile/values/fluentd-user.yaml.gotmpl
pettersv: Yes, removed it

OlleLarsson:
Remember to add configuration options for the retention of authlog indices.

helmfile/values/fluentd-wc.yaml.gotmpl
```diff
@@ -65,6 +65,18 @@ extraConfigMaps:
       @type record_transformer
       remove_keys responseObject,requestObject
     </filter>
+    # Index the authlog
```
OlleLarsson, line 68:
```suggestion
    # Tail the authlog
```

OlleLarsson (commented):
Still missing options to configure [retention](https://github.com/elastisys/opendistro-build/blob/main/helm/opendistro-es/values.yaml#L614-L615) for authlog indices.

pettersv:
> Still missing options to configure [retention](https://github.com/elastisys/opendistro-build/blob/main/helm/opendistro-es/values.yaml#L614-L615) for authlog indices.

OK so wherer does this belong? In sc-config?

OlleLarsson (approved):
LGTM.
Have you tried upgrading to see that everything works as expected, that configurer creates the new resources for authlogs etc? I personally think that it will create everything needed but just checking if you've tested.

pettersv:
> LGTM.
> Have you tried upgrading to see that everything works as expected, that configurer creates the new resources for authlogs etc? I personally think that it will create everything needed but just checking if you've tested.

Yes, I have tested it. Although I did not set up a pristine cluster for all the tests, did some manual cleaning to save time.

viktor-f (approved):
Looks good.
Though the retention values seem a bit strange. In particular that the default size is higher that the prod size. So change that if you want, but it's no blocker.

pettersv:
> Looks good.
> Though the retention values seem a bit strange. In particular that the default size is higher that the prod size. So change that if you want, but it's no blocker.

I copied the values from the "other*" index as I have no real clue what it should be. We have to adjust this later once we get the feeling for how large this index grows. Btw the other* index has the same strange setup with larger values in default than prod. I'll create an issue for that.

Issue #355, opened 2021-03-25
Move away from "pulling" charts


pettersv:
**What should be investigated.**

*Task from arch meeting 2021-03-25*
We currently pull the 3:rd party helm chart on apply. If a link is broken or the chart removed upstream the install will fail. Also, it is not possible to "grep" for values in the charts as they are not present in our repo. Investigate if there is another way to do this.

**What artifacts should this produce.**
A proposal and feasibility study of an alternate apporach, perhaps "vendoring" the charts.

cristiklein:
Thanks @pettersv !

PR #384, opened 2021-04-14
Add S3-exporter for metrics about buckets and dashboard for daily tasks

### LINKED ISSUE #373: [3] Convert daily checks into dashboards

In the spirit of drinking our own Champagne and making our work more demonstrable, let's convert some of the daily checks into dashboards.

Grafana dashboards that can be demonstrated by sales and are useful to ops. The ambition should be to eventually make daily scripts unnecessary. (They can still be around for troubleshooting.)

**Additional context**

For inspiration, see how the [Backup dashboard](https://compliantkubernetes.io/ciso-guide/backup/) allows us to prove to ourselves that backups are working.

Also, check how the [Capacity dashboard](https://compliantkubernetes.io/ciso-guide/kubernetes-status/) allows us to prove that all nodes / Pods are up.

**Definition of done:**

* Grafana dashboards to automate daily checks.
* Documentation for these daily checks on compliantkubernetes.io.

lentzi90:
This adds an s3-exporter and service monitor so that we get metrics about our buckets. These metrics can be used to check if backups and retention jobs are working as they should, and keep an eye on the size.

**Which issue this PR fixes**: fixes #373. 

I'm not happy with the templating of the dashboard to add thresholds. It feels bad in many ways so I would suggest that we drop the last commit and go with a dashboard without thresholds for now.

**Checklist:**

    - [x] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -35,6 +35,8 @@ objectStorage:
   # s3:
   #   region: "set-me"
   #   regionEndpoint: "set-me"
+  #   # Generally false when using AWS and Exoscale and true for other providers.
+  #   forcePathStyle: true
   buckets:
     harbor: "${CK8S_ENVIRONMENT_NAME}-harbor"
     velero: "${CK8S_ENVIRONMENT_NAME}-velero"
@@ -597,3 +599,16 @@ calicoAccountant:
 
 clusterAdmin:
   admins: []
+s3Exporter:
+  # Also requries objectStorage.type=s3
+  enabled: true
+  interval: 60m
+  scrapeTimeout: 10m
+  resources:
+    limits: {}
+    requests:
+      cpu: 50m
+      memory: 20Mi
+  tolerations: []
+  affinity: {}
+  nodeSelector: {}
--- a/helmfile/charts/grafana-ops/values.yaml
+++ b/helmfile/charts/grafana-ops/values.yaml
@@ -26,3 +26,5 @@ dashboards:
     enabled: true
   networkpolicy:
     enabled: true
+  daily:
+    enabled: true
--- a/helmfile/charts/s3-exporter/templates/configmap.yaml
+++ b/helmfile/charts/s3-exporter/templates/configmap.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "s3-exporter.fullname" . }}-env
+  labels:
+    {{- include "s3-exporter.labels" . | nindent 4 }}
+data:
+  AWS_REGION: {{ .Values.s3.region }}
+  S3_EXPORTER_S3_ENDPOINT_URL: {{ .Values.s3.regionEndpoint }}
+  S3_EXPORTER_S3_FORCE_PATH_STYLE: {{ .Values.s3.forcePathStyle | quote }}
--- a/helmfile/charts/s3-exporter/templates/servicemonitors.yaml
+++ b/helmfile/charts/s3-exporter/templates/servicemonitors.yaml
@@ -1,0 +1,42 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "s3-exporter.fullname" . }}
+  labels:
+    {{- include "s3-exporter.labels" . | nindent 4 }}
+spec:
+  endpoints:
+  # One endpoint per bucket so that we can pass the name of the bucket as param
+  {{- $interval := .Values.serviceMonitor.interval }}
+  {{- $scrapeTimeout := .Values.serviceMonitor.scrapeTimeout }}
+  {{- range .Values.s3.buckets }}
+  - port: http
+    scheme: http
+    path: "/probe"
+    interval: {{ $interval }}
+    scrapeTimeout: {{ $scrapeTimeout }}
+    params:
+      bucket:
+      - {{ . }}
+      # We could add a prefix here if we are just interested in specific files/folders
+      # prefix: []
+    metricRelabelings:
+      - sourceLabels: [__address__]
+        regex: '^bucket=(.*);prefix=(.*);$'
+        replacement: '${1}'
+        targetLabel: __param_target
+      - sourceLabels: [__address__]
+        regex: '^bucket=(.*);prefix=(.*);$'
+        replacement: '${2}'
+        targetLabel: '__param_prefix'
+      - targetLabel: __address__
+        replacement: 127.0.0.1:9340  # S3 exporter.
+  {{- end }}
+  jobLabel: s3_exporter
+  selector:
+    matchLabels:
+      {{- include "s3-exporter.selectorLabels" . | nindent 6 }}
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace }}
--- a/helmfile/charts/s3-exporter/values.yaml
+++ b/helmfile/charts/s3-exporter/values.yaml
@@ -1,0 +1,73 @@
+# Default values for s3-exporter.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+s3:
+  region: ""
+  regionEndpoint: ""
+  forcePathStyle: true
+  accessKey: ""
+  secretKey: ""
+  buckets: []
+    # - my-bucket
+    # - second-bucket
+
+serviceMonitor:
+  interval: 10m
+  scrapeTimeout: 60s
+
+replicaCount: 1
+
+image:
+  repository: elastisys/s3_exporter
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: ""
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+podAnnotations: {}
+
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext:
+  capabilities:
+    drop:
+    - ALL
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  runAsUser: 100
+
+service:
+  type: ClusterIP
+  port: 9340
+
+resources: {}
+  # We usually recommend not to specify default resources and to leave this as a conscious
+  # choice for the user. This also increases chances charts run on environments with little
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+nodeSelector: {}
+
+tolerations: []
+
+affinity: {}
--- a/helmfile/values/grafana-ops.yaml.gotmpl
+++ b/helmfile/values/grafana-ops.yaml.gotmpl
@@ -20,3 +20,5 @@ dashboards:
     enabled: true
   networkpolicy:
     enabled: true
+  daily:
+    enabled: true
--- a/helmfile/values/s3-exporter.yaml.gotmpl
+++ b/helmfile/values/s3-exporter.yaml.gotmpl
@@ -1,0 +1,16 @@
+s3:
+  region: {{ .Values.objectStorage.s3.region | quote }}
+  regionEndpoint: {{ .Values.objectStorage.s3.regionEndpoint | quote }}
+  accessKey: {{ .Values.objectStorage.s3.accessKey | quote }}
+  secretKey: {{ .Values.objectStorage.s3.secretKey | quote }}
+  buckets: {{ values .Values.objectStorage.buckets | toYaml | nindent 2 }}
+  forcePathStyle: {{ .Values.objectStorage.s3.forcePathStyle }}
+
+serviceMonitor:
+  interval: {{ .Values.s3Exporter.interval }}
+  scrapeTimeout: {{ .Values.s3Exporter.scrapeTimeout }}
+
+resources: {{- toYaml .Values.s3Exporter.resources | nindent 2  }}
+nodeSelector: {{- toYaml .Values.s3Exporter.nodeSelector | nindent 2  }}
+tolerations: {{- toYaml .Values.s3Exporter.tolerations | nindent 2  }}
+affinity: {{- toYaml .Values.s3Exporter.affinity | nindent 2  }}
--- a/pipeline/config/exoscale/sc-config.yaml
+++ b/pipeline/config/exoscale/sc-config.yaml
@@ -575,3 +575,15 @@ calicoAccountant:
   enabled: true
 clusterAdmin:
   admins: []
+s3Exporter:
+  enabled: true
+  interval: 120s
+  scrapeTimeout: 30s
+  resources:
+    limits: {}
+    requests:
+      cpu: 50m
+      memory: 20Mi
+  tolerations: []
+  affinity: {}
+  nodeSelector: {}
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -13,6 +13,7 @@
 
 - Authlog now indexed by elasticsearch
 - Added a ClusterRoleBinding for using an OIDC-based cluster admin kubeconfig and a script for generating such a kubeconfig (see `bin/ck8s kubeconfig admin`)
+- S3-exporter for collecting metrics about S3 buckets.
 
 ### Removed
 
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -14,6 +14,7 @@
 - Authlog now indexed by elasticsearch
 - Added a ClusterRoleBinding for using an OIDC-based cluster admin kubeconfig and a script for generating such a kubeconfig (see `bin/ck8s kubeconfig admin`)
 - S3-exporter for collecting metrics about S3 buckets.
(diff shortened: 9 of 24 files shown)


helmfile/charts/s3-exporter/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+s3:
+  region: ""
+  regionEndpoint: ""
+  accessKey: ""
+  secretKey: ""
+  buckets: []
+    # - my-bucket
+    # - second-bucket
+
+serviceMonitor:
```
Xartos, line 14: Maybe add a `seviceMonitor.enabled` here? Although for our use case I guess we never want do disable the service monitor.

On that note, if this we should host this chart. Doesn't it fit better if we create and upload it to https://github.com/elastisys/chart-repo?

helmfile/charts/s3-exporter/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+podAnnotations: {}
+
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext:
+  capabilities:
+    drop:
+    - ALL
```
Xartos, line 47: Isn't this by default?

helmfile/charts/s3-exporter/templates/servicemonitors.yaml
```diff
@@ -0,0 +1,42 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "s3-exporter.fullname" . }}
+  labels:
+    {{- include "s3-exporter.labels" . | nindent 4 }}
+spec:
+  endpoints:
+  # One endpoint per bucket so that we can pass the name of the bucket as param
```
Xartos, line 10: Why is this necessary?

pipeline/config/exoscale/sc-config.yaml
```diff
@@ -575,3 +575,15 @@ metricsServer:
   enabled: true
 calicoAccountant:
   enabled: true
+s3Exporter:
```
Xartos, line 578: Should we add image tag here?

config/config/sc-config.yaml
```diff
@@ -597,3 +597,16 @@ metricsServer:
 
 calicoAccountant:
   enabled: true
+
+s3Exporter:
```
Xartos, line 601: Should we add image tag here?

Xartos (commented):
Great stuff! some comments though

helmfile/charts/s3-exporter/values.yaml
lentzi90: Good point! I didn't think about it actually, but I could definitely add it there. This chart is pretty generic so I guess it could be useful to separate it from ck8s On the other hand, we have https://github.com/elastisys/compliantkubernetes-apps/issues/355 that seems to be about just vendoring the charts so I'm not sure. But I do like the idea of having the charts in the chart-repo so I'll add a PR there and remove it from here unless I hear objections :slightly_smiling_face:

helmfile/charts/s3-exporter/templates/servicemonitors.yaml
Xartos: I saw that this is a requirement in the repo. So I guess this is the way it is.

helmfile/charts/s3-exporter/values.yaml
lentzi90: I don't think so but I'm not sure.
```console
❯ k explain deployment.spec.template.spec.containers.securityContext.capabilities
KIND:     Deployment
VERSION:  apps/v1

RESOURCE: capabilities <Object>

DESCRIPTION:
     The capabilities to add/drop when running containers. Defaults to the
     default set of capabilities granted by the container runtime.

```

helmfile/charts/s3-exporter/templates/servicemonitors.yaml
lentzi90: The s3_exporter works by "probes", so we need to tell Prometheus what to probe for. With curl, you do it like this:
```
curl localhost:9340/probe?bucket=some-bucket&prefix=some-folder/some-file.txt
```
We want to probe all our buckets so we need to somehow tell Prometheus to scrape the s3-exporter for all of them. Specifying multiple endpoints was the best solution I could come up with. The blackbox exporter [is very similar](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/charts/blackbox/templates/servicemonitors.yaml), but there we do one whole ServiceMonitor per target instead of just an endpoint. Could probaby rework it.

config/config/sc-config.yaml
lentzi90: Hmm, not sure about this. We don't seem to expose tags for anything else?

helmfile/charts/s3-exporter/values.yaml
Xartos: Right, yea then I guess it makes sense.

config/config/sc-config.yaml
Xartos: Right, but then I think that we should have a tag set for the chart values file. It doesn't feel very reassuring to always use latest

lentzi90: It uses the charts app version by default so not the latest. :slightly_smiling_face: 
```
image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```

OlleLarsson:
Have you tested the exporter against different S3 providers?
I think I would be good to confirm wteather or not it uses path style  or dns access by default, so testing against safespring/citycloud (path access) and exoscale (dns access) would be nice 🙂

lentzi90:
> Have you tested the exporter against different S3 providers?
> I think I would be good to confirm wteather or not it uses path style or dns access by default, so testing against safespring/citycloud (path access) and exoscale (dns access) would be nice

True, I have just tested with Exoscale yet. I'm also afraid there is no way to make this work with GCP... But then again there are already some other things that only work with S3 so I guess it is ok?
Will check Safespring soon :slightly_smiling_face:

OlleLarsson:
> True, I have just tested with Exoscale yet. I'm also afraid there is no way to make this work with GCP... But then again there are already some other things that only work with S3 so I guess it is ok?
> Will check Safespring soon

Sounds good.
GCP is a completely different topic that I would say is out of scope.

config/config/sc-config.yaml
Xartos: Ooh, I missed that. Yea, that's fine then

helmfile/charts/s3-exporter/values.yaml
OlleLarsson: I think we can keep that chart in this repo for the tmie being :)

lentzi90: I looked into this a bit and realized that the chart-repo is for publishing the charts, not for "development", right? So I'll keep it here for now but we can still publish it if we want to. Later we could create a separate repo for the chart also if we want to decouple it.

lentzi90:
I have now tested (and fixed) this on Safespring also and added the daily dashboard.

I'm not happy with the templating of the dashboard to add thresholds. It feels bad in many ways so I would suggest that we drop the last commit and go with a dashboard without thresholds for now. It will be less useful of course, but maintaining the templated json is very painful. If you have any suggestions for how to improve the situation I'm all ears!

helmfile/charts/grafana-ops/values.yaml
```diff
@@ -26,3 +26,32 @@ dashboards:
...
+      serviceClusterKB: 4500000
+  thresholds:
+    # Thresholds for the number of elasticsearch snapshots
+    elasticsearchSnapshots:
+      critical: 20
+      warning: 16
+    # How old should the oldest snapshot get before it indicates an issue?
+    # Daily backups + retention of max 14 snapshots means maximum 15 days.
+    # With time based retention set to 10 days, it should be no more than 11.
+    elasticsearchOldestSnapshotSeconds:
```
OlleLarsson, line 45: It depends, we might have a bug where a snapshot is taken every 10'th minute meaning that snapshots will most likely be removed due to `nr of snapshots > max snapshots` so no snapshot will ever be "old".
But then again the converse can also happen - too few snapshots are taken/failing, meaning that no snapshot will be removed due to `nr of snapshots < min snapshots`.

I guess in essence I don't know there answer to your question.

helmfile/charts/s3-exporter/templates/configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "s3-exporter.fullname" . }}-env
+  labels:
+    {{- include "s3-exporter.labels" . | nindent 4 }}
+data:
+  AWS_REGION: {{ .Values.s3.region }}
+  S3_EXPORTER_S3_ENDPOINT_URL: {{ .Values.s3.regionEndpoint }}
+  S3_EXPORTER_S3_FORCE_PATH_STYLE: {{ .Values.s3.forcePathStyle | quote }}
```
OlleLarsson, line 10: Can we shorten this down?
```suggestion
  S3_ENDPOINT_URL: {{ .Values.s3.regionEndpoint }}
  S3_FORCE_PATH_STYLE: {{ .Values.s3.forcePathStyle | quote }}
```

helmfile/values/s3-exporter.yaml.gotmpl
```diff
@@ -0,0 +1,15 @@
+s3:
```
OlleLarsson, line 1: I think `forcePathStyle` is missing from here, or should we just always have it set to `true`?

pipeline/config/exoscale/sc-config.yaml
```diff
@@ -575,3 +575,15 @@ metricsServer:
...
+  enabled: true
+  interval: 120s
```
OlleLarsson, line 580: Do you think that there are any downsides to having an even higher interval, let's say every 10'th minute?

OlleLarsson (commented):
Looks good to me, I've left a few comments.
I don't really have much to say about how the thresholds are set in the grafana dashboard.

helmfile/charts/grafana-ops/values.yaml
```diff
@@ -26,3 +26,32 @@ dashboards:
...
+daily:
+  limits:
+    influxdb:
+      workloadClusterKB: 4500000
+      serviceClusterKB: 4500000
+  thresholds:
+    # Thresholds for the number of elasticsearch snapshots
+    elasticsearchSnapshots:
+      critical: 20
+      warning: 16
```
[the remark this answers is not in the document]
lentzi90: Right, this is a good point. I was thinking that too few would be detected in the "last modified object" pane, but not sure. Do you think it will be enough? I mean if there are no snapshots made then the last modified object will just get older and older and trigger its own threshold.

helmfile/charts/grafana-ops/values.yaml
lentzi90: Yeah this is complicated. I was first thinking that it would be great to calculate this number based on the values we already have. But I think that would be way too hard, and basically require us to parse the cron-schedule. So this is probably the best we can do. Let the user set the time. :slightly_smiling_face:

helmfile/charts/s3-exporter/templates/configmap.yaml
lentzi90: No they come straight from the [s3-exporter](https://github.com/ribbybibby/s3_exporter#flags). The other option would be to set them as flags, but I think it is better like this.

helmfile/values/s3-exporter.yaml.gotmpl
lentzi90: Good point, I can add it :+1:

helmfile/charts/s3-exporter/templates/configmap.yaml
OlleLarsson: Okay.

helmfile/charts/grafana-ops/values.yaml
OlleLarsson: Agree

pipeline/config/exoscale/sc-config.yaml
lentzi90: I was also thinking about this. I tried it with 5 minutes and it worked fine, except that Grafana was showing gaps in the graphs. It is probably possible to tweak the dashboard to "know" how sparse the points are, but then this would also need templating if we want to change the interval. Maybe it would be a good idea to lock the interval and configure the dashboard to work with large value?

I should also mention that this is very relevant for sc-logs. It takes a long time to get metrics for this bucket since it contains so many objects. This means that the s3-exporter can start timing out if it takes too long, (and we obviously cannot have the timeout longer than the interval).

helmfile/values/grafana-ops.yaml.gotmpl
```diff
@@ -20,3 +20,28 @@ dashboards:
     enabled: true
   networkpolicy:
     enabled: true
+  daily:
+    enabled: true
+
+daily:
```
viktor-f, line 26: Shouldn't this be under `dashboards.daily`? That's at least how it is for `gatekeeper` and `falco`.

viktor-f (commented):
I think it's ok with this type of templating, it's at least better than nothing.

helmfile/charts/grafana-ops/values.yaml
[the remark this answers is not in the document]
viktor-f: Together with "last modified object" I think it is probably enough to look at "too many" here.

pipeline/config/exoscale/sc-config.yaml
viktor-f: If you can make the dashboard work with it, then a longer static interval sounds good.

helmfile/values/grafana-ops.yaml.gotmpl
```diff
@@ -20,3 +20,26 @@ dashboards:
...
+      # Thresholds for how long ago the last modification should be before firing
+      bucketObjectModifiedSeconds:
+        warning: 86400 # 24h
+        critical: 172800 # 48h
+      s3QuotaBytes:
+        warning: 4000000000 # 4GB
+        critical: 5000000000 # 5GB
+      influxdbLeftKB:
+        warning: 2000000 # 2GB
+        critical: 1000000 # 1GB
```
viktor-f, line 45: Shouldn't these be read from config?

lentzi90:
Alright, after a discussion on slack I have now removed the templating from the dashboard. It was way too cumbersome to work with and made the whole dashboard much more complicated. Everything from thresholds to queries and "data density" had to be templated to make sure it would work in different situations.

So now there is no templating at all, no thresholds, limits or other "dynamic" features. The dashboard just shows the data and doesn't try to interpret it to tell you if the some value is too high or low. Some assumptions were still needed for data density, i.e. how much time is there between the data points. This is how it works:

1. It can take a long time to scrape metrics from a bucket if there are lots of objects in it. The situation doesn't change that fast either so it is not valuable for us to have data for ever 30 seconds or similar. For these reasons, the S3-exporter is configured to scrape the buckets with a 1h interval. This value can be changed.
2. Grafana automatically picks an interval between data points based on the time range and size of the screen. If there are missing data points for example because the sample rate is much lower than what Grafana chose, there will be "holes" in the graph. We can tell Grafana how to treat these null values (zero, "connected" or missing), but it can be quite confusing to look at and won't work with multiple time series (see example below).

   Grafana cannot know about how dense or sparse the data is, so we would have to tell it how to work around it. We don't want to template the dashboard though, so it must be hard coded. I have chosen to hard code the dashboard to show only one data point per hour for the object storage metrics. This should be sparse enough to not cause trouble but still useful to look at.

## Data density and phase

There are a number of things you can do to make Grafana work nicely with sparse data. Not everything will work in all situations though, so I'll try to describe with an example how and why I have done it.

### What does the data look like?

Prometheus scrapes it from the S3-exporter. The metrics are per bucket, so we need to scrape for each bucket we want data from (very similar to blackbox exporter). Prometheus is configured with an interval and timeout to use, but *does not scrape all buckets simultaneously*. This means that even though we set the same interval for all of them, the metrics will most likely be staggered. With 1h interval, Prometheus may decide to start scraping bucket 1 at 12.00, bucket 2 at 12.10 and bucket 3 at 12.30. Bucket 1 would be scraped again at 13.00, bucket 2 at 13.10 and bucket 3 at 13.30. So the interval is correct but we cannot assume that the data points will line up.

prometheus-sparse-data

### What can we do to make Grafana understand the data?

The most obvious thing to do is probably adjusting the "Min interval":
> A lower limit for the interval. Recommended to be set to write frequency, for example 1m if your data is written every minute. Default value can be set in data source settings for most data sources.

This will make sure that Grafana doesn't try to find data points more dense than this interval. Great!
Unfortunately though, it doesn't help when the data is staggered. Grafana would for example find the data for bucket 1 and maybe bucket 2 if it is close enough, but bucket 3 is completely out of phase so Grafana would never find it. This means the dashboards would be completely useless. If we calculate the total size of the buckets, there is no guarantee that all buckets are actually included! If we plot buckets separately, we may only see half of them. Compare this to the above.

grafana-sparse-data

To solve this, we need to write the queries so that they gather data over a certain time and then make that into a single data point. So instead of finding 1 data point and be happy about it, Grafana would query over 1h to find all the data points there, then take the average, max or min value of each series and preset as the single data point for the interval. This way the data from all buckets will be found, assuming that Prometheus scrapes at least once per hour.

Example query:
```
sum(max_over_time(s3_objects_size_sum_bytes[1h]))
```
Explanation:
- `s3_objects_size_sum_bytes` = metric showing the total size of a bucket (so one value for each bucket)
- `max_over_time(<series>[1h])` = for each series, go through a 1h interval and find the maximum value. This gives us the maximum size of each bucket during the 1h around the point we are interested in.
- `sum(<series>)` = aggregate over all the buckets to get the total size.

OlleLarsson (approved):
LGTM.
Well done @lentzi90 I can tell you put a lot of effort into this👏🏻 🙌🏻

viktor-f (approved):
Very well done!

PR #388, opened 2021-04-15
all: add RBAC and script for OIDC-based admin kubeconfig

### LINKED ISSUE #323: [3] Use individual credentials for admin access to Kubernetes API
Cluster administrators should use individual credentials when accessing the Kubernetes API, instead of relying on a shared admin kubeconfig. Otherwise it is impossible to know who accessed the cluster.

The necessary RBAC to grant all cluster admins access should be created as part of bootstrapping the cluster. After this the admin kubeconfig should not be used anymore for other than break glass situations. To ensure this, it would make sense to purge it from the local machine. In a break glass situation, it can simply be retrieved from the control-plane nodes.

In short:
1. Get admin kubeconfig
2. Install Dex
3. Create RBAC to bind cluster admins to the `cluster-admin` ClusterRole at the cluster level.
4. Generate a kubeconfig for individual access (see https://github.com/elastisys/compliantkubernetes-apps/issues/221 )
5. Purge the admin kubeconfig

An alternative would be to generate individual certificates and kubeconfigs for all admins using the CA used by the cluster [like this](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#normal-user). The certificates would need to be rotated every now and then and there is no way to revoke access without replacing all the certificates, so this solution would add a bit of operational overhead.

**Additional context**
This issues is in a way part of https://github.com/elastisys/ck8s/issues/50 and https://github.com/elastisys/ck8s/issues/442.
Note that Dex must be installed before we can switch to OIDC authentication, so it must be part of the bootstrap.
It may make sense to move the bootstrap to compliantkubernetes-kubespray and handle everything in this repository using OIDC auth. Since Dex becomes crucial for accessing the cluster with this change we should also add another replica to make it a bit more resilient.

**Definition of done:**
Cluster administrators use individual credentials when accessing Kubernetes API, for example by using OIDC authentication. Access to the shared admin kubeconfig should require ssh to the control-plane to retrieve it (so that the user is identified through ssh).

Ajarmar:
**What this PR does / why we need it**: Adds a ClusterRoleBinding to the `cluster-admin` ClusterRole, and a script (as part of `bin/ck8s`) that can be used to generate a cluster admin kubeconfig that requires authentication to be used.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #323 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.


bin/admin-kubeconfig.bash
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
```
Xartos, line 1: Why must this be a separate file from `user-kubeconfig`? Shouldn't the config look the same in the end?

bin/ck8s
```diff
@@ -28,6 +28,7 @@ usage() {
     echo "  ops helmfile <wc|sc>                        run helmfile as cluster admin" 1>&2
     echo "  s3cmd [cmd]                                 run s3cmd" 1>&2
     echo "  user-kubeconfig                             generate user kubeconfig, stored at CK8S_CONFIG_PATH/user/"
+    echo "  admin-kubeconfig                            generate cluster-admin kubeconfig, stored at CK8S_CONFIG_PATH/.state/"
```
Xartos, line 31: Is it necessary to store it in `.state`? Isn't that why we want it in the first place, to not need to save the kubeconfig in the repo.

I guess maybe this is a start, and a separate issue would be to remove the assumption of having the kubeconfig in this location and rely on `KUBECONFIG` instead.

helmfile/charts/cluster-admin-rbac/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Kubernetes
```
Xartos, line 3:
```suggestion
description: A Helm chart for cluster-admin RBAC
```

helmfile/charts/cluster-admin-rbac/values.yaml
```diff
@@ -0,0 +1 @@
+users: []
```
Xartos, line 1: Just out of curiosity, what happens if you deploy with this empty? Is that still valid?

Xartos (commented):
I think it looks good! Some comments.

And also, you haven't updated the config for the pipeline

bin/admin-kubeconfig.bash
Ajarmar: The resulting kubeconfig is almost exactly the same, with a few minor differences (such as calling the user `admin@$cluster_name` instead of `user@$cluster_name` and giving the kubeconfig a different filename). The most significant difference between the two scripts is that `admin-kubeconfig.bash` should be able to be used on any cluster whereas `user-kubeconfig.bash` only applies to the workload cluster, so `admin-kubeconfig.bash` needs some arguments specifying the cluster while `user-kubeconfig.bash` does not. I guess a solution to reducing the duplicated code would be to consolidate the shared code in `common.bash`.

bin/ck8s
Ajarmar: I interpreted [this comment from Cristian](https://github.com/elastisys/compliantkubernetes-apps/issues/323#issuecomment-801784144) as "put it in .state for now and fix it later when we move towards actually implementing the ADR fully", so that would be a separate issue like you said.

helmfile/01-applications.yaml
```diff
@@ -162,6 +162,23 @@ releases:
...
+  namespace: kube-system
+  labels:
+    app: cluster-admin-rbac
+  chart: ./charts/cluster-admin-rbac
+  version: 0.1.0
+  missingFileHandler: Error
+  {{ if .Values.opa.enabled }}
+  needs:
+  - gatekeeper-system/gatekeeper-constraints
+  {{ end }}
```
viktor-f, line 177:
```suggestion
  {{ if eq .Environment.Name "workload_cluster" }}
  {{ if .Values.opa.enabled }}
  needs:
  - gatekeeper-system/gatekeeper-constraints
  {{ end }}
  {{ end }}
```
This seems to be required in the service cluster

bin/ck8s
viktor-f:
```suggestion
    echo "  admin-kubeconfig <wc|sc>                    generate cluster-admin kubeconfig, stored at CK8S_CONFIG_PATH/.state/"
```

helmfile/01-applications.yaml
OlleLarsson: Why does the chart need gatekeeper-constraints?

bin/admin-kubeconfig.bash
Xartos: Right, well then I would rather like a `kubeconfig.bash` file that have these small difference by giving it an argument or something like that. I feel like `common.bash` only should have stuff that is common for all scripts.

helmfile/01-applications.yaml
Ajarmar: I wasn't sure about this - I just included it since `user-rbac` includes it. Does `user-rbac` only need it because it creates some extra namespaces?

bin/admin-kubeconfig.bash
Ajarmar: Sure, that makes sense. I'd be fine with doing it that way.

helmfile/01-applications.yaml
viktor-f: Right, I didn't even consider if it was needed. Yes user-rbac should only include it because of the namespaces it creates. So you should be able to delete it here.

Xartos (approved):
LGTM!

helmfile/01-applications.yaml
viktor-f:
```suggestion
```
This whole part can be removed.

viktor-f:
Added one clarification on the helmfile part.
Also the tests are failing (make sure you have the pre-commit hook installed locally).

Otherwise this looks good!

helmfile/charts/cluster-admin-rbac/values.yaml
Ajarmar: Tried it out and yes, it is still valid. It will give you a `ClusterRoleBinding` with no subjects.

helmfile/01-applications.yaml
Ajarmar: My bad, completely forgot about this.

viktor-f (approved):
LGTM

Issue #391, opened 2021-04-16
cert-manager webhook failure


cristiklein:
**Describe the bug**

We seem to have a race condition with the cert-manager webhook. Writing it here so we know about it.

**To Reproduce**
1. Create a brand new vanilla Kubernetes cluster, or re-use an old one with `./scripts/clean-sc.sh`.
2. Create a new configuration (e.g., `./bin/ck8s init`, edit relevant fields).
3. Apply `./bin/ck8s apply sc`.

**Expected behavior**

Apply should work successfully.

**Actual behavior**

```
FAILED RELEASES:
NAME
issuers
opendistro-es
in ./01-applications.yaml: 2 errors:
err 0: failed processing release issuers: command "/home/cklein/bin/helm" exited with non-zero status:

PATH:
  /home/cklein/bin/helm

ARGS:
  0: helm (4 bytes)
  1: upgrade (7 bytes)
  2: --install (9 bytes)
  3: --reset-values (14 bytes)
  4: issuers (7 bytes)
  5: charts/issuers (14 bytes)
  6: --version (9 bytes)
  7: 0.1.0 (5 bytes)
  8: --timeout (9 bytes)
  9: 600s (4 bytes)
  10: --namespace (11 bytes)
  11: cert-manager (12 bytes)
  12: --values (8 bytes)
  13: /tmp/helmfile395826681/cert-manager-issuers-values-55f8745cd5 (61 bytes)
  14: --history-max (13 bytes)
  15: 10 (2 bytes)

ERROR:
  exit status 1

EXIT STATUS
  1

STDERR:
  Error: Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": x509: certificate signed by unknown authority

COMBINED OUTPUT:
  Release "issuers" does not exist. Installing it now.
  Error: Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": x509: certificate signed by unknown authority
err 1: failed processing release opendistro-es: command "/home/cklein/bin/helm" exited with non-zero status:

PATH:
  /home/cklein/bin/helm

ARGS:
  0: helm (4 bytes)
  1: upgrade (7 bytes)
  2: --install (9 bytes)
  3: --reset-values (14 bytes)
  4: opendistro-es (13 bytes)
[...]

PR #392, opened 2021-04-19
Bootstrap rework PoC

### LINKED ISSUE #362: [3] Investigate if Helm 3 got better with handlings CRDs and namespaces

The bootstrap step exists (mainly) to set up CRDs and namespace, since Helm 2 was not great at that. However, Helm 3 can now create namespaces and handles CRDs better. Helmfile now supports `createNamespace`, `disableValidationOnInstall`, which should also make using CRDs smoother.

Can we get rid of CRDs and namespaces in the bootstrap step? How much time would that save? How many lines of code would that remove?

* A branch which shows how to get rid of CRDs and namespaces.
* Measurements of how much it takes to execute `./bin/ck8s apply sc|wc`, after performing `./scripts/clean-*.sh`.
* If above does not work, please document why it does not work (detailed error message).

**Implementation hint**

We might need to be more "aggressive" on `wait` and `needs`. E.g., `cert-manager` needs to have the Pod handing its ValidatingWebhook running *before* installing the `issuers` Chart.

OlleLarsson:
This pr shows a poc of how it might be possible to rework the appplication installation by removing namespaces and crd installation from our "bootstrap" step.

**Which issue this PR fixes**
fixes #362 

CRDs:
- While helm allows us to install CRDs, it does nothing in terms of lifecycle management - aka upgrading and removing them.
  So, we'll need to devise some other way of handling that.
- We can no loger run `template --validate` in `bin/ck8s apply` for obvious reasons - the CRDs dosen't exist.
  Use `apply {wc,sc} --skip-template-validate` to skip templating.
- We need to disable validation `disableValidationOnInstall: true` on all releases that'll install a cr. 

Namespaces:
- I had to remove 00-rbac installation as that one installs roles in namespace that have yet to be created.
  I tried to enable psps in all charts that allowed it.
  - I had to upgrade harbor in order for it to start.
  - Influxdb does not work without access to priviledged psp atm, I just removed the cronjob sidecar to deploy influxdb.
- I did not bother to create a chart or any other means of labling namespaces like our `namespaces` bootstrap chart did - so no namespace is labled with `owner: operator` and the namespaces `kube-system` and `gatekeeper-system` are not labled with`admission.gatekeeper.sh/ignore: "true"`.
- Fluentd secrets needed to be moved to after the helmfile apply step in the deploy-sc/wc script.
  This of course means that we might have some problem in the pipeline where fluentd is crashlooping because it could not mount the secrets immediately.
- Gatekeeper is another tricky topic as it installs an validation admission webhook that essentially blocks creation of namespaces until the gatekeeper pods are ready to serve the hook.
  I had to make sure that gatekeeper is installed last, thus mitigating the problem that it does not allow for any concurrecnt chart installations that creates a new namespace.
- I have essentially done a complete rework of the DAG in helmfile.
  It was pretty strange that a lot of releases depended on cert-manager even though they did not create any cert-manager cr.
  It would probably to be good to have the DAG on paper for easy visualization.
  That's essntially what i did on paper in order for this to work and visualize the dependencies.

Cleanup:
- The cleanup scripts will need to be changed.
  It does not uninstall crds.

The afformentioned issue mentioned something about timing the `bin/ck8s apply` step. I have not done that. However, I very much doubt that difference would be anything else than in the order of seconds.

Alertmanager (user):
- I did somehting for it so that it does not install psps for it if it is disabled.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.


Xartos (commented):
`+225 −48,708` I'd say that's a pretty good!
Can't see anything negatives with this, as long as it works that is

viktor-f:
Very interesting!

Does `helmfile diff` still work with this? You only mentioned `helmfile template --validate`

Labeling namespaces might be something that we want to keep, but it should be easy to do with a chart as you imply.

A DAG of the helmfile would be nice if you could get it in a good format.

OlleLarsson:
> Does `helmfile diff` still work with this? You only mentioned `helmfile template --validate`

It should still work, see https://github.com/roboll/helmfile/pull/1618 for some explanation of --disableValidationOnInstall.

I mean after CRDs have been installed it should be fine to run `template`.

viktor-f:
> It should still work, see [roboll/helmfile#1618](https://github.com/roboll/helmfile/pull/1618) for some explanation of --disableValidationOnInstall.
> 
> I mean after CRDs have been installed it should be fine to run `template`.

Right, so `helmfile diff` does not work before the first installation, but afterwards it does work. That seems fair.

lentzi90:
> While helm allows us to install CRDs, it does nothing in terms of lifecycle management - aka upgrading and removing them.
> So, we'll need to devise some other way of handling that.
> We need to disable validation disableValidationOnInstall: true on all releases that'll install a cr.

I think this is enough reason to keep CRDs separate from the rest. We will anyway need a way to upgrade/remove CRDs so why not install them the same way? I.e. just like we do today.

The same is also true for namespaces I'd say. We want to label them, maybe even add annotations in the future, so why make it more complicated by mixing two ways to manage them? With this we would first need to install the relevant helm release before we even have a namespace to label.

cristiklein:
Do we still have time as part of this issue?

We just discussed #391 and it seems to me that we are:
1. Installing too many releases in parallel. Those that "change Kubernetes" by installing webhooks should not run in parallel to anything else.
2. The bootstrap steps feels like an out-of-helmfile solution to serialization. The Helmfile documentation suggests (and @viktor-f confirms) that separate Helmfile-s are serialized.

Could we instead do the following instead?
```
00-namespaces.yaml => installs and labels namespaces
01-gatekeeper.yaml => installs Gatekeeper
02-prometheus-operator.yaml
10-cert-manager.yaml
11-nginx-ingress.yaml
50-everything-else.yaml
```

OlleLarsson:
> Do we still have time as part of this issue?
No.

cristiklein:
Thanks @OlleLarsson. I created a follow-up issue.

lentzi90:
> Could we instead do the following instead?

I think this is a good idea. It makes it easier to understand what is going on with less massive files.
However, I'm not sure it solves the issue with webhooks. AFAIK helm is quite limited in what it can wait for so even when it thinks it is done with one helmfile it is no guarantee that the webhooks are ready. If the Pods become "ready" but it still needs time to set up the webhook then this issue can still happen.

cristiklein:
@OlleLarsson Can we close this, thanks to #446 ?

PR #393, opened 2021-04-19
apps wc: Only install alertmgr rbac if enabled

OlleLarsson:
Makes it so that rbac for user alertmanager is only installed when it is actually enabled.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/helmfile/values/user-rbac.yaml.gotmpl
+++ b/helmfile/values/user-rbac.yaml.gotmpl
@@ -2,4 +2,6 @@ namespaces: {{ toYaml .Values.user.namespaces | nindent 2 }}
 users: {{ toYaml .Values.user.adminUsers | nindent 2 }}
 createNamespaces: {{ .Values.user.createNamespaces }}
 enableFalcoViewer: {{ toYaml .Values.falco.enabled | nindent 2 }}
-alertmanagerNamespace: {{ .Values.user.alertmanager.namespace }}
+alertmanager:
+  enabled: {{ .Values.user.alertmanager.enabled }}
+  namespace: {{ .Values.user.alertmanager.namespace }}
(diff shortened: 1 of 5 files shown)


helmfile/values/user-rbac.yaml.gotmpl
```diff
@@ -2,4 +2,6 @@ namespaces: {{ toYaml .Values.user.namespaces | nindent 2 }}
 users: {{ toYaml .Values.user.adminUsers | nindent 2 }}
 createNamespaces: {{ .Values.user.createNamespaces }}
 enableFalcoViewer: {{ toYaml .Values.falco.enabled | nindent 2 }}
-alertmanagerNamespace: {{ .Values.user.alertmanager.namespace }}
+alertmanager:
+  enabled: {{ .Values.user.alertmanager.enabled }}
+  namespace: {{ .Values.user.alertmanager.namespace }}
```
lentzi90, line 7: We need a migration for this, right? Translating the old `alermanagerNamespace` to `alertmanager.namespace`.

lentzi90 (commented):
Looks good, but will need a little migration

helmfile/values/user-rbac.yaml.gotmpl
OlleLarsson: I don't think so, the namespace is still assumed from the same template variable: `{{ .Values.user.alertmanager.namespace }}`.

lentzi90: Ah sorry, I read that a bit too fast. :sweat_smile:

lentzi90 (approved):

viktor-f (approved):

PR #395, opened 2021-04-20
all: remove default codeowner

Ajarmar:
**What this PR does / why we need it**: Removes the default codeowner, which used a deprecated team.

**Special notes for reviewer**: Should the CODEOWNERS file just be removed entirely, or is there still use for it?

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,9 +3,6 @@
 # Note: if many patterns match the same file, the latest pattern
 # will override the previous ones.
 
-# Default for files with no matching pattern later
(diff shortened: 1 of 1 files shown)


OlleLarsson (approved):
OK.
We should probably clean up and remove non dev-ops people from being codeowners as well, but this is ok as a start.

lentzi90 (approved):
It's a start, but we need to go through the rest of the list as well. That can be done next sprint.

PR #405, opened 2021-04-22
Force elasticsearch-slm cronjob to use integers

### LINKED ISSUE #339: Elasticsearch-slm job doesn't fail on error
**Describe the bug**
Issue #337 was discovered only after the elasticsearch-slm jobs started to hit their deadline. Before that they have run for several days just looping through all snapshots and erroring on the malformed `MAX_AGE_SECONDS`. The Job should fail if there is an [error in the script](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/charts/elasticsearch-slm/scripts/slm-retention.bash).

**To Reproduce**
1. Create an environment with elasticsearch and elasticshearch-slm
2. Set the `MAX_AGE_SECONDS` to something weird, like `2.592e+06`
3. Trigger an SLM job. You can do this with `./bin/ck8s ops kubectl sc -n elastic-system create job manual-slm --from=cronjob/elasticsearch-slm`
4. Check what happens with the Job. Does it complete "successfully"?

**Expected behavior**
The Job should not be counted as successfully completed if the script has errors.

 - Compliant kubernetes apps version v0.9.0

**Additional context**
I'm not really sure why the script doesn't exit with an error since it looks like we specify the necessary parameters [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/charts/elasticsearch-slm/scripts/slm-retention.bash#L13)

Xartos:

Bug fix

fixes #339 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.


lentzi90 (commented):
Seems reasonable.

I stumbled across [this](https://stackoverflow.com/questions/13103701/how-to-trap-errors-inside-the-if-statement) way of checking the exit code for the test. Do you think that would be useful also?
Maybe we could do something like this?
```bash
[ "12346" -gt "$(LC_ALL=C printf '%.0f\n' 123345.0)" ];
if [[ $? -eq 0 ]];
...
```

Xartos:
> Maybe we could do something like this?

I still think that we should convert the variable in the beginning since it's used in multiple places. I guess we can move out the test but I feel like we should do that for all the if statements then?

lentzi90:
Yes that sounds good. Not sure if it is useful to do all if statements like this though. If we do it in this file, we should also do it everywhere else. :slightly_smiling_face:

OlleLarsson (approved):

lentzi90 (approved):

PR #425, opened 2021-04-30
bin: Fix check for bash array in team.bash

lentzi90:
The check was for a normal bash environment variable, but it is actually an array.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: none

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Fixed
 
 - When using harbor together with rook there is a potential bug that appears if the database pod is killed and restarted on a new node. This is fixed by upgrading the Harbor helm chart to version 1.6.1.
+- The command `team-add` for adding new PGP fingerprints no longer crashes when validating some environment variables.
 
 ### Added
 
--- a/bin/team.bash
+++ b/bin/team.bash
@@ -75,7 +75,7 @@ sops_remove_pgp() {
 # Update all secrets with the public keys from the fingerprints in the SOPS
 # config file.
 sops_update_keys() {
-    : "${secrets:?Missing secrets}"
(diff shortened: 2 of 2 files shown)


viktor-f (approved):

OlleLarsson (approved):
I'm not 100% sure what this changes.
Do you have any examples of what didn't work that now does?

lentzi90:
Basically the bash expansion doesn't work with an array so I got `Missing secrets` when I tried running `team-add`. I came up with this that seems to work. Here is an example of how it works:

```console
❯ declare -A secrets
❯ secrets["a"]="something"
❯ secrets["b"]="somethingelse"
❯ echo "secrets: ${secrets[@]}"
❯ : "${secrets:?Missing secrets}"
bash: secrets: Missing secrets
❯ : "${secrets[@]:?Missing secrets array}"
❯ # Try an empty array to see if we catch it
❯ declare -A empty_array
❯ echo "empty_array: ${empty_array[@]}"
empty_array:
❯ : "${empty_array:?Missing empty_array}"
bash: empty_array: Missing empty_array
❯ : "${empty_array[@]:?Missing empty_array array}"
bash: empty_array[@]: Missing empty_array array
```

simonklb (approved):
I don't remember adding this? Why is the check necessary? Something must have gone horribly wrong if the secrets variable is unset. :smile: In any case, LGTM!

Really curious how this hasn't been hit before though? Is the command not being used?

lentzi90:
I guess it is not being used enough at least :joy: And honestly, I don't get why the check is necessary. :slightly_frowning_face:

PR #426, opened 2021-05-03
bin: Make S3 entry script more verbose

### LINKED ISSUE #424: Make S3 scripts safer
The [S3 scripts](https://github.com/elastisys/compliantkubernetes-apps/blob/main/scripts/S3/entry.sh) currently don't do much to ensure that you "know what you are doing". They don't explain what credentials are used, what endpoint it is trying to use or what buckets it will create/destroy. This means it is quite easy to do mistakes and increase the insecurity.

The scripts should ask for confirmation and explain what it is about to do.

None

**Definition of done:**
The s3 scripts explain asks for confirmation at least before deleting things and also lists what buckets it is about to touch.

OlleLarsson:
To make the S3 scripts safer.
This just adds some more output to the entry script:
- Print buckets the will be operated on
- Print which s3cfg config is used
- Print command executed - this really contains the above information as well

fixes #424 

Printing region or endpoint is not very useful imo as that would require the script to parse the s3cfg file.
I find it more sensical to really just present that "Hey, this is the s3cfg file you passed", if you need more information then a simple cat should do :)

I did not create a Yes/no promt for creating buckets as that is not really an intrusive operation.

Feedback appreciated :) 

**Example run**:
image

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/scripts/S3/entry.sh
+++ b/scripts/S3/entry.sh
@@ -10,44 +10,57 @@ CK8S_AUTO_APPROVE=${CK8S_AUTO_APPROVE:-"false"}
 
 here="$(dirname "$(readlink -f "$0")")"
 
+log_info() {
+    echo -e "[\e[34mck8s\e[0m] ${*}" 1>&2
+}
+
+log_error() {
+    echo -e "[\e[31mck8s\e[0m] ${*}" 1>&2
+}
+
 objectstorage_type_sc=$(yq r "${CK8S_CONFIG_PATH}/sc-config.yaml" 'objectStorage.type')
 objectstorage_type_wc=$(yq r "${CK8S_CONFIG_PATH}/wc-config.yaml" 'objectStorage.type')
 
-[ "$objectstorage_type_sc" != "s3" ] && echo "S3 is not enabled in service cluster" 1>&2
-[ "$objectstorage_type_wc" != "s3" ] && echo "S3 is not enabled in workload cluster" 1>&2
+[ "$objectstorage_type_sc" != "s3" ] && log_info "S3 is not enabled in service cluster"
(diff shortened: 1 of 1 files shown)


lentzi90 (approved):
Excellent, thank you very much!

scripts/S3/entry.sh
```diff
@@ -10,44 +10,57 @@ CK8S_AUTO_APPROVE=${CK8S_AUTO_APPROVE:-"false"}
...
     action="$1"
     cmd="${here}/manager.sh --$1 $buckets"
+    log_info "Using s3cmd config file: ~/.s3cfg"
 fi
 
 if [ "$action" = "delete" ] && ! ${CK8S_AUTO_APPROVE}; then
-    echo -n -e "Are you sure you want to delete all buckets? (y/n): " 1>&2
+    echo -n -e "[\e[34mck8s\e[0m] Are you sure you want to delete all buckets? (y/n): " 1>&2
     read -r reply
     if [[ "${reply}" == "n" ]]; then
```
Xartos, line 59: This is not part of this change. But I would much rather have a `if [ $reply != "[yY]" ]; then exit;`
Because this script would continue if I misstyped and entered an "m" instead.

Xartos (changes requested):
Great job. Saw a small thing outside of this issue that you could fix while your at it

Xartos (approved):

simonklb (approved):
The changes looks good!

If you have the time and it seems like a small change it would be nice if you could only delete the buckets for one cluster at the time. Totally optional though!

viktor-f (approved):

PR #431, opened 2021-05-05
apps sc & wc: improve dry-run and deploy scripts

### LINKED ISSUE #382: [1] Improve diff in dryrun script
In the dryrun.sh script we currently use `helmfile diff`. This will not check agianst what the actual state of the cluster is. Change this to use `helmfile template | kubectl diff` instead.

In the apply.sh script, we should use `helmfile sync` instead of `helmfile apply`

robinelastisys:
The dry-run.sh and deploy scripts we currently use will not check against the actual state of the cluster. We are now adding commando flags to enable the option of using sync if a user want to.

**Which issue this PR fixes** 
fixes #382

**Message to reviewers**
Im not sure if this is the right way to do it. This was the closest i could figure out. Feel free to point me out in the right direction if needed.

**Checklist:**

- [x] Some test is done to check if the problem still persists.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -7,6 +7,7 @@
 
 - The sc-logs-retention cronjob now runs without error even if no backups were found for automatic removal
 - Harbor Swift authentication configuration options has moved from `citycloud` to `harbor.persistence.swift`.
+- The dry-run and apply command now have the options to check against the state of the cluster while ran by using the flags "--sync" and "--kubectl".
 
 ### Fixed
 
--- a/bin/apps.bash
+++ b/bin/apps.bash
@@ -22,19 +22,17 @@ apps_init() {
 
 apps_run_sc() {
     log_info "Applying applications in service cluster"
-
     (
         with_kubeconfig "${secrets[kube_config_sc]}" \
-            CONFIG_PATH="${CK8S_CONFIG_PATH}" "${scripts_path}/deploy-sc.sh"
+            CONFIG_PATH="${CK8S_CONFIG_PATH}" "${scripts_path}/deploy-sc.sh" "${1:-""}"
     )
 }
 
 apps_run_wc() {
     log_info "Applying applications in workload cluster"
-
     (
         with_kubeconfig "${secrets[kube_config_wc]}" \
-            CONFIG_PATH="${CK8S_CONFIG_PATH}" "${scripts_path}/deploy-wc.sh"
+            CONFIG_PATH="${CK8S_CONFIG_PATH}" "${scripts_path}/deploy-wc.sh" "${1:-""}"
     )
 }
 
@@ -73,16 +71,18 @@ apps_sc() {
     # but feels good enough until we figure out something smarter.
     #
     #[ "$1" != "--skip-template-validate" ] && template_validate_sc
-    apps_run_sc
+
+    apps_run_sc "${2:-""}"
 
     log_info "Applications applied successfully!"
 }
 
 apps_wc() {
     apps_init
-    # See rationale above.
+    # See rationale above
     #[ "$1" != "--skip-template-validate" ] && template_validate_wc
-    apps_run_wc
+
+    apps_run_wc "${2:-""}"
 
     log_info "Applications applied successfully!"
 }
@@ -90,14 +90,12 @@ apps_wc() {
 #
 # ENTRYPOINT
 #
-
-
 if [[ $1 == "wc" ]]; then
     config_load "$1"
-    apps_wc "$2"
+    apps_wc "$2" "$3"
 elif [[ $1 == "sc" ]]; then
     config_load "$1"
-    apps_sc "$2"
+    apps_sc "$2" "$3"
 else
     echo "ERROR:  [$1] is an invalid argument:"
     echo "usage:   ck8s apps <wc|sc>. "
--- a/bin/ck8s
+++ b/bin/ck8s
@@ -10,30 +10,42 @@ source "${here}/common.bash"
 
 usage() {
     echo "COMMANDS:" 1>&2
-    echo "  init                                            initialize the config path" 1>&2
-    echo "  bootstrap <wc|sc>                               bootstrap the cluster" 1>&2
-    echo "  apps <wc|sc> [--skip-template-validate]         deploy the applications" 1>&2
-    echo "  apply <wc|sc> [--skip-template-validate]        bootstrap and apps" 1>&2
-    echo "  test <wc|sc>                                    test the applications" 1>&2
-    echo "  dry-run <wc|sc>                                 runs helmfile diff" 1>&2
-    echo "  team add-pgp <fp>                               add a new PGP key to secrets" 1>&2
-    echo "  team remove-pgp <fp>                            remove a PGP key from secrets and rotate the data encryption key" 1>&2
+    echo "  init                                              initialize the config path" 1>&2
+    echo "  bootstrap <wc|sc>                                 bootstrap the cluster" 1>&2
+    echo "  apps <wc|sc> [--sync] [--skip-template-validate]  deploy the applications" 1>&2
+    echo "  apply <wc|sc> [--sync] [--skip-template-validate] bootstrap and apps" 1>&2
+    echo "  test <wc|sc>                                      test the applications" 1>&2
+    echo "  dry-run <wc|sc> [--kubectl]                       runs helmfile diff" 1>&2
+    echo "  team add-pgp <fp>                                 add a new PGP key to secrets" 1>&2
+    echo "  team remove-pgp <fp>                              remove a PGP key from secrets and rotate the data encryption key" 1>&2
     # TODO: We might want to make this command less visible once we have proper
     #       support for OIDC logins.
-    echo "  ops kubectl <wc|sc>                             run kubectl as cluster admin" 1>&2
-    echo "  ops helm <wc|sc>                                run helm as cluster admin" 1>&2
+    echo "  ops kubectl <wc|sc>                               run kubectl as cluster admin" 1>&2
+    echo "  ops helm <wc|sc>                                  run helm as cluster admin" 1>&2
     # TODO: We might want to make this command less visible once we feel
     #       confident that the apply command and migrations are good enough
     #       that direct Helmfile access is not necessary.
-    echo "  ops helmfile <wc|sc>                            run helmfile as cluster admin" 1>&2
-    echo "  s3cmd [cmd]                                     run s3cmd" 1>&2
-    echo "  kubeconfig <user|admin <wc|sc> [cluster_name]>  generate user/admin kubeconfig, stored at CK8S_CONFIG_PATH/user or CK8S_CONFIG_PATH/.state"
-    echo "  completion bash                                 output shell completion code for bash" 1>&2
-    echo "  validate <wc|sc>                                validates config files" 1>&2
-    echo "  providers                                       lists supported cloud providers" 1>&2
-    echo "  flavors                                         lists supported configuration flavors" 1>&2
+    echo "  ops helmfile <wc|sc>                              run helmfile as cluster admin" 1>&2
+    echo "  s3cmd [cmd]                                       run s3cmd" 1>&2
+    echo "  kubeconfig <user|admin <wc|sc> [cluster_name]>    generate user/admin kubeconfig, stored at CK8S_CONFIG_PATH/user or CK8S_CONFIG_PATH/.state"
+    echo "  completion bash                                   output shell completion code for bash" 1>&2
+    echo "  validate <wc|sc>                                  validates config files" 1>&2
+    echo "  providers                                         lists supported cloud providers" 1>&2
+    echo "  flavors                                           lists supported configuration flavors" 1>&2
     exit 1
 }
+
+SYNC=""
+SKIP=""
+KUBECTL=""
+
+for arg in "$@"; do
+  case "$arg" in
+    "--skip-template-validate") SKIP="--skip-template-validate" ;;
+    "--sync") SYNC="sync" ;;
+    "--kubectl") KUBECTL="kubectl" ;;
+  esac
+done
 
 case "${1}" in
     init)
@@ -45,12 +57,12 @@ case "${1}" in
         ;;
     apps)
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
-        "${here}/apps.bash" "${2}" "${3}"
+        "${here}/apps.bash" "${2}" "${SKIP}" "${SYNC}"
         ;;
     apply)
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
         "${here}/bootstrap.bash" "${2}"
-        "${here}/apps.bash" "${2}" "${3}"
+        "${here}/apps.bash" "${2}" "${SKIP}" "${SYNC}"
         ;;
     test)
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
@@ -58,7 +70,7 @@ case "${1}" in
         ;;
     dry-run)
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
-        "${here}/dry-run.bash" "${2}"
+        "${here}/dry-run.bash" "${2}" "${KUBECTL}"
         ;;
     team)
         case "${2}" in
--- a/bin/dry-run.bash
+++ b/bin/dry-run.bash
@@ -21,12 +21,19 @@ config_load "$1"
 if [[ $1 == "sc" ]]; then
   log_info "Running helmfile diff on the service cluster"
 
-  "${here}/ops.bash" helmfile sc diff
-
+  if [ ${#} -eq 2 ] && [ "$2" = "kubectl" ]; then
+    "${here}/ops.bash" helmfile sc template | "${here}/ops.bash" kubectl sc diff -f -
+  else
+    "${here}/ops.bash" helmfile sc diff
+  fi
 elif [[ $1 == "wc" ]]; then
   log_info "Running helmfile diff on the workload cluster"
 
-  "${here}/ops.bash" helmfile wc diff
+  if [ ${#} -eq 2 ] && [ "$2" = "kubectl" ]; then
+    "${here}/ops.bash" helmfile wc template | "${here}/ops.bash" kubectl wc diff -f -
+  else
+    "${here}/ops.bash" helmfile wc diff
+  fi
 else
   log_error "ERROR: unsupported option for dry-run. Supported options are <wc|sc>"
   exit 1
--- a/scripts/deploy-sc.sh
+++ b/scripts/deploy-sc.sh
@@ -10,30 +10,33 @@ source "${SCRIPTS_PATH}/../bin/common.bash"
 : "${secrets[secrets_file]:?Missing secrets}"
 
 alertTo=$(yq r -e "${config[config_file_sc]}" 'alerts.alertTo')
-if [[ "$alertTo" != "slack" && "$alertTo" != "null" && "$alertTo" != "opsgenie" ]]
-then
+if [[ "$alertTo" != "slack" && "$alertTo" != "null" && "$alertTo" != "opsgenie" ]]; then
     log_error "ERROR: alerts.alertTo must be set to one of slack, opsgenie or null."
     exit 1
 fi
-
 
 INTERACTIVE=${1:-""}
 
 objectStoreProvider=$(yq r -e "${config[config_file_sc]}" objectStorage.type)
 if [[ ${objectStoreProvider} == "s3" ]]; then
-  echo "Creating fluentd secrets" >&2
-  s3_access_key=$(sops_exec_file "${secrets[secrets_file]}" 'yq r -e {} objectStorage.s3.accessKey')
-  s3_secret_key=$(sops_exec_file "${secrets[secrets_file]}" 'yq r -e {} objectStorage.s3.secretKey')
-  kubectl create secret generic s3-credentials -n fluentd \
-      --from-literal=s3_access_key="${s3_access_key}" \
-      --from-literal=s3_secret_key="${s3_secret_key}" \
-      --dry-run=client -o yaml | kubectl apply -f -
+    echo "Creating fluentd secrets" >&2
+    s3_access_key=$(sops_exec_file "${secrets[secrets_file]}" 'yq r -e {} objectStorage.s3.accessKey')
+    s3_secret_key=$(sops_exec_file "${secrets[secrets_file]}" 'yq r -e {} objectStorage.s3.secretKey')
+    kubectl create secret generic s3-credentials -n fluentd \
+        --from-literal=s3_access_key="${s3_access_key}" \
+        --from-literal=s3_secret_key="${s3_secret_key}" \
+        --dry-run=client -o yaml | kubectl apply -f -
 fi
 
 echo "Installing helm charts" >&2
 cd "${SCRIPTS_PATH}/../helmfile"
 declare -a helmfile_opt_flags
 [[ -n "$INTERACTIVE" ]] && helmfile_opt_flags+=("$INTERACTIVE")
-helmfile -f . -e service_cluster "${helmfile_opt_flags[@]}" apply --suppress-diff
+
+if [ ${#} -eq 1 ] && [ "$1" = "sync" ]; then
+    helmfile -f . -e service_cluster sync
+else
+    helmfile -f . -e service_cluster apply --suppress-diff
+fi
 
 echo "Deploy sc completed!" >&2
--- a/scripts/deploy-wc.sh
+++ b/scripts/deploy-wc.sh
@@ -26,14 +26,19 @@ kubectl -n fluentd create secret generic elasticsearch \
 # We use `create` here instead of `apply` to avoid overwriting any changes the
 # user may have done.
 kubectl create -f "${SCRIPTS_PATH}/../manifests/examples/fluentd/fluentd-extra-config.yaml" \
-    2> /dev/null || echo "fluentd-extra-config configmap already in place. Ignoring."
+    2>/dev/null || echo "fluentd-extra-config configmap already in place. Ignoring."
 kubectl create -f "${SCRIPTS_PATH}/../manifests/examples/fluentd/fluentd-extra-plugins.yaml" \
-    2> /dev/null || echo "fluentd-extra-plugins configmap already in place. Ignoring." >&2
+    2>/dev/null || echo "fluentd-extra-plugins configmap already in place. Ignoring." >&2
 
 echo "Installing helm charts" >&2
 cd "${SCRIPTS_PATH}/../helmfile"
 declare -a helmfile_opt_flags
 [[ -n "$INTERACTIVE" ]] && helmfile_opt_flags+=("$INTERACTIVE")
-helmfile -f . -e workload_cluster "${helmfile_opt_flags[@]}" apply --suppress-diff
+
+if [ ${#} -eq 1 ] && [ "$1" = "sync" ]; then
+    helmfile -f . -e workload_cluster sync
+else
+    helmfile -f . -e workload_cluster apply --suppress-diff
+fi
 
 echo "Deploy wc completed!" >&2


bin/dry-run.bash
```diff
@@ -21,13 +21,13 @@ if [[ $1 == "sc" ]]; then
   log_info "Running helmfile diff on the service cluster"
   : "${config[config_file_sc]:?Missing service cluster config file}"
 
-  "${here}/ops.bash" helmfile sc diff
+  "${here}/ops.bash" helmfile sc template | kubectl sc diff
```
simonklb, line 24: Could you give an example of what this solves?

scripts/deploy-sc.sh
```diff
@@ -34,6 +34,6 @@ echo "Installing helm charts" >&2
 cd "${SCRIPTS_PATH}/../helmfile"
 declare -a helmfile_opt_flags
 [[ -n "$INTERACTIVE" ]] && helmfile_opt_flags+=("$INTERACTIVE")
-helmfile -f . -e service_cluster "${helmfile_opt_flags[@]}" apply --suppress-diff
+helmfile -f . -e service_cluster "${helmfile_opt_flags[@]}" sync --suppress-diff
```
simonklb, line 37: Why are you doing this change?

WIP-CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
...
+
+- The dry-run and deploy scripts now checks against the state of the cluster while ran.
+
+### Fixed
+
+
+### Added
+
+
+### Removed
```
simonklb, line 11: No need to add empty headings.

simonklb (changes requested):
Mostly looking for explanations. Otherwise this looks fine!

bin/dry-run.bash
robinelastisys: Actually not sure. @pettersv Do you have a good example?

scripts/deploy-sc.sh
robinelastisys: This is the description of the commands:

    sync      sync all resources from state file (repos, releases and chart deps)
    apply     apply all resources from state file only when there are changes

So, i think it would be good for the sake of recovering a state? Not sure i have done changes in the right place. It was the only place i could find in where it would effect the "apply" command as the issuer wanted. Couldn't find a "helmfile apply" statement in either apps.sh or bootstrap.sh .

simonklb: I'm not sure if this actually will change something in practice, but if it does it's definitely worth mentioning in the changelog as this is changing the apply command, not dry-run.

OlleLarsson: The reason why this was even brought up in the first place is that helm-diff does not check the acutally state in kubernetes, and thus if somebody has made any manual changes to let's say a deployment, helm-diff won't notice that and thus not upgrade the release. sync, will always upgrade the releases to a new version thus syncing the state on every single apply.

Personally, now that we are aware of the fact of this difference between apply and sync. Whether or not we want to run sync only and never apply, I can't say.

bin/dry-run.bash
OlleLarsson, line 24:
```suggestion
  "${here}/ops.bash" helmfile sc template | kubectl diff
```

OlleLarsson, line 24: Have you actually tried running this?
```suggestion
  "${here}/ops.bash" helmfile sc template | kubectl diff
```

bin/dry-run.bash
```diff
@@ -21,13 +21,13 @@ if [[ $1 == "sc" ]]; then
...
 
 elif [[ $1 == "wc" ]]; then
   log_info "Running helmfile diff on the workload cluster"
   : "${config[config_file_wc]:?Missing workload cluster config file}"
 
-  "${here}/ops.bash" helmfile wc diff
+  "${here}/ops.bash" helmfile wc template | kubectl wc diff
```
OlleLarsson, line 30: Same

bin/dry-run.bash
viktor-f: As noted in the issue `helm diff` does not check what's actually running in the cluster, but it was suggested that piping the template to `kubectl diff` could actually provide us with this. So if there was any manual changes we could see how that would impact this upgrade.

simonklb: Gotcha, I guess one should refrain from doing manual changes instead of going the route of adding things to git first etc, but I see the point and it's a good one!

Are we sure that doing `kubectl diff` instead of `helm diff` won't lose any information? It's hard to visualize, but could there be something that helm tracks that is not visible by just looking at the "plain k8s state"?

scripts/deploy-sc.sh
simonklb: Worth investigating before merging? :smile:

bin/dry-run.bash
viktor-f: Yes I think that everyone agrees that we should stay away from manual changes. But it will happen sometimes for different reasons.

At least I am not sure, so it would be very good to test (@robinelastisys have you been testing these changes, and if so then how have you tested it?). But I think you would get all relevant information.

scripts/deploy-sc.sh
Xartos: I think we said that we should use sync since this is the same as running `helm upgrade --install` (instead of `helm apply` for when apply is used) which is how you would do it if you were using helm.

simonklb: `helm apply` isn't a thing I think? At least not on my end:
```
% helm apply
Error: unknown command "apply" for "helm"
Run 'helm --help' for usage.
% helm version --short
v3.2.4+g0ad800e
```

simonklb: Just did a quick search since I wasn't totally aware of the commands and I think the top answer [here](https://stackoverflow.com/questions/59703760/helmfile-sync-vs-helmfile-apply) explained it pretty clearly.

To me it feels like we should not run `helmfile sync` since it's going to basically upgrade the chars even though there are no changes made. Or am I missing something here?

Xartos: Oh, yea I misread the documentation there.
But I think the reason to why we want this is (as olle said) diff will not notice stuff that's manually edited. And since that might be the case that we need to do that (for an emergency fix or something) we should not allow this manual change to hang around for a couple of upgrades until the release are actually redeployed and breaks everything again (maybe months later). So in my mind it's better that we notice this change quicker when we might have a chance of remembering why. But there might be some issues with upgrading the release that I don't see?

simonklb: Gotcha! I'm a bit worried that manual changes that have been done would be destroyed by a sync though? For example, what happens if you have made manual changes to some labels and it runs an upgrade of the chart release? It's all of these unknowns that I would be a bit cautious introducing before we actually know more.

If anyone here actually have made some effort into testing this and researching sync vs apply implications please add it to the changelog so that we don't introduce this without anyone knowing what is going on.

robinelastisys:
We had a meeting with the outcome:

The changes could be activated by using a command flag

robinelastisys:
Now i have presented a start of a solution. Not the most beautiful but it works. "apply-sync" and "dry-run-sync".

bin/apps.bash
```diff
@@ -23,23 +23,45 @@ apps_init() {
...
+        )
+    else
+        (
+            : "${scripts_path:?Missing scripts path}"
+            : "${secrets[kube_config_sc]:?Missing service cluster kubeconfig}"
+            with_kubeconfig "${secrets[kube_config_sc]}" \
+                CONFIG_PATH="${CK8S_CONFIG_PATH}" "${scripts_path}/deploy-sc.sh"
+        )
+
+    fi
```
Xartos, line 42: You should be able to simplify this to something like this, right?

```suggestion
        (
            : "${scripts_path:?Missing scripts path}"
            : "${secrets[kube_config_sc]:?Missing service cluster kubeconfig}"
            with_kubeconfig "${secrets[kube_config_sc]}" \
                CONFIG_PATH="${CK8S_CONFIG_PATH}" "${scripts_path}/deploy-sc.sh" "${1:-""}"
        )
```

bin/apps.bash
```diff
@@ -23,23 +23,45 @@ apps_init() {
...
+        )
+    else
+        (
+            : "${scripts_path:?Missing scripts path}"
+            : "${secrets[kube_config_wc]:?Missing workload cluster kubeconfig}"
+            with_kubeconfig "${secrets[kube_config_wc]}" \
+                CONFIG_PATH="${CK8S_CONFIG_PATH}" "${scripts_path}/deploy-wc.sh"
+        )
+
+    fi
```
Xartos, line 64: Same here

bin/apps.bash
```diff
@@ -75,33 +97,65 @@ apps_sc() {
...
+
+    if [ ${#} -eq 2 ]; then
+
+        #[ "$1" != "--skip-template-validate" ] && template_validate_sc
+        apps_run_sc "$2"
+    else
+        #[ "$1" != "--skip-template-validate" ] && template_validate_sc
+        apps_run_sc
 
+    fi
```
Xartos, line 110: Same here

bin/apps.bash
```diff
@@ -75,33 +97,65 @@ apps_sc() {
...
+
+    if [ ${#} -eq 2 ]; then
 
+        #[ "$1" != "--skip-template-validate" ] && template_validate_wc
+        apps_run_wc "$2"
+    else
+        #[ "$1" != "--skip-template-validate" ] && template_validate_wc
+        apps_run_wc
+
+    fi
```
Xartos, line 136: Same here

bin/apps.bash
```diff
@@ -75,33 +97,65 @@ apps_sc() {
...
+
 elif [[ $1 == "sc" ]]; then
     config_load "$1"
-    apps_sc "$2"
+    if [ ${#} -eq 3 ]; then
+        apps_sc "$2" "$3"
+    else
+        apps_sc "$2"
+    fi
```
Xartos, line 158: And here

bin/dry-run.bash
```diff
@@ -21,14 +21,28 @@ if [[ $1 == "sc" ]]; then
   log_info "Running helmfile diff on the service cluster"
   : "${config[config_file_sc]:?Missing service cluster config file}"
 
-  "${here}/ops.bash" helmfile sc diff
+  if [ ${#} -eq 2 ]; then
```
Xartos, line 24: I would like that you actually verify that the variable is set to "sync" and not just assuming that it's set correctly.

This applies to all places where this assumption is done.

bin/dry-run.bash
```diff
@@ -21,14 +21,28 @@ if [[ $1 == "sc" ]]; then
...
 
-elif [[ $1 == "wc" ]]; then
+    "${here}/ops.bash" helmfile sc template | "${here}/ops.bash" kubectl sc diff -f -
+  else
+    "${here}/ops.bash" helmfile sc diff
+
+  fi
+elif
+  [[ $1 == "wc" ]]
+then
```
Xartos, line 33:
```suggestion
elif [[ $1 == "wc" ]]; then
```

Xartos (changes requested):
Would it be possible to instead of introducing a new command, having a flag instread? Like
```console
$ bin/ck8s apply --sync
```
I guess something like [getops](https://www.golinuxcloud.com/bash-getopts/) would be something but that doesn't support long arguments (only single letter). If it's not feasable I'm okey with this approach as well.

I would also like to rename the `dry-run-sync` to `dry-run-kubectl` since it has nothing to do with sync.
There's also some formatting issues I have commented on

robinelastisys:
So, now i have done a lot of changes.

1. I changed from apply-sync and dry-run-sync to --sync and --kubectl.
2. I changed the if-cases to go by "${1:-""}" instead of if-case series.
3. I did some  string checks at the entry point of apps.bash and in deploy-wc & sc for further development if/when more flags come.

scripts/deploy-wc.sh
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 SCRIPTS_PATH="$(dirname "$(readlink -f "$0")")"
 # shellcheck disable=SC1090 # Can't follow non-constant source.
 source "${SCRIPTS_PATH}/../bin/common.bash"
-
+s
```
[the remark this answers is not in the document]
robinelastisys: dammit haha :smile:

bin/ck8s
```diff
@@ -13,9 +13,9 @@ usage() {
     echo "  init                                            initialize the config path" 1>&2
     echo "  bootstrap <wc|sc>                               bootstrap the cluster" 1>&2
     echo "  apps <wc|sc> [--skip-template-validate]         deploy the applications" 1>&2
-    echo "  apply <wc|sc> [--skip-template-validate]        bootstrap and apps" 1>&2
+    echo "  apply <wc|sc> [--sync] [--skip-template-validate] bootstrap and apps" 1>&2
```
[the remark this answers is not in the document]
robinelastisys: I thought about that me also.. its a tricky one as getopts aren't created to handle longer flags like that :thinking:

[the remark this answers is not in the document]
simonklb: @robinelastisys usually long flags/options is done by iterating over the arguments, would that work here? Something like this, for example:
```
for arg in "$@"; do
  case "$arg" in
    "--sync") sync=1 ;;
    "--skip-template-validate") skip_template_validate=1 ;;
  esac
done
```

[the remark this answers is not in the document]
robinelastisys: Hmmmmm.. interesting idea! Should work somehow indeed!  :thinking:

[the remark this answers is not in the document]
robinelastisys: So, now i have implemented a version in where the input parameters are parsed and minor bugs are fixed.

bin/apps.bash
```diff
@@ -75,34 +71,75 @@ apps_sc() {
...
+    #    #[ "$1" != "--skip-template-validate" ] && template_validate_sc
+
     #[ "$1" != "--skip-template-validate" ] && template_validate_sc
-    apps_run_sc
+
+    if [ "$#" -gt 0 ] && [ "${*: -1}" = "sync" ]; then
+        apps_run_sc "${*: -1}"
+    else
+        apps_run_sc
+    fi
```
Xartos, line 82: Wouldn't this work?

```suggestion
        apps_run_sc "${*: -1}"
```

bin/apps.bash
```diff
@@ -75,34 +71,75 @@ apps_sc() {
...
+    #    #[ "$1" != "--skip-template-validate" ] && template_validate_sc
+
     #[ "$1" != "--skip-template-validate" ] && template_validate_wc
-    apps_run_wc
+
+    if [ "$#" -gt 0 ] && [ "${*: -1}" = "sync" ]; then
+        apps_run_wc "${*: -1}"
+    else
+        apps_run_wc
+    fi
```
Xartos, line 107: Same here

bin/apps.bash
```diff
@@ -75,34 +71,75 @@ apps_sc() {
     # While it would be nice to have some template validation before `helmfile apply`,
     # at least Helmfile does "just in time" template validation. Not as nice,
     # but feels good enough until we figure out something smarter.
-    #
+    #    #[ "$1" != "--skip-template-validate" ] && template_validate_sc
+
     #[ "$1" != "--skip-template-validate" ] && template_validate_sc
-    apps_run_sc
+
+    if [ "$#" -gt 0 ] && [ "${*: -1}" = "sync" ]; then
```
Xartos, line 78: Since we can be sure that "sync" are on position 3 I would rather have `$3` instead of this. 
Less confusion.

bin/apps.bash
```diff
@@ -23,23 +23,19 @@ apps_init() {
...
-    (
-        : "${scripts_path:?Missing scripts path}"
-        : "${secrets[kube_config_sc]:?Missing service cluster kubeconfig}"
-        with_kubeconfig "${secrets[kube_config_sc]}" \
-            CONFIG_PATH="${CK8S_CONFIG_PATH}" "${scripts_path}/deploy-sc.sh"
-    )
+    : "${scripts_path:?Missing scripts path}"
+    : "${secrets[kube_config_sc]:?Missing service cluster kubeconfig}"
+    with_kubeconfig "${secrets[kube_config_sc]}" \
+        CONFIG_PATH="${CK8S_CONFIG_PATH}" "${scripts_path}/deploy-sc.sh" "${1:-""}"
```
Xartos, line 29: Why removing the parenthesis?

Xartos (changes requested):
Just some minor details

robinelastisys:
Done! Looks a lot smoother now indeed! :-) :+1:

bin/ck8s
```diff
@@ -43,21 +56,21 @@ case "${1}" in
...
         ;;
     apply)
-        [[ "${2}" =~ ^(wc|sc)$ ]] || usage
-        "${here}/bootstrap.bash" "${2}"
-        "${here}/apps.bash" "${2}" "${3}"
-        ;;
+    [[ "${2}" =~ ^(wc|sc)$ ]] || usage
+    "${here}/bootstrap.bash" "${2}"
+    "${here}/apps.bash" "${2}" "$SKIP" "$SYNC"
+    ;;
```
Xartos, line 65: Intentation
```suggestion
        [[ "${2}" =~ ^(wc|sc)$ ]] || usage
        "${here}/bootstrap.bash" "${2}"
        "${here}/apps.bash" "${2}" "$SKIP" "$SYNC"
        ;;
```

bin/ck8s
```diff
@@ -43,21 +56,21 @@ case "${1}" in
...
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
         "${here}/test.bash" "${2}"
         ;;
     dry-run)
-        [[ "${2}" =~ ^(wc|sc)$ ]] || usage
-        "${here}/dry-run.bash" "${2}"
-        ;;
+    [[ "${2}" =~ ^(wc|sc)$ ]] || usage
+        "${here}/dry-run.bash" "${2}" $KUBECTL
+    ;;
```
Xartos, line 73: Indentation
```suggestion
        [[ "${2}" =~ ^(wc|sc)$ ]] || usage
        "${here}/dry-run.bash" "${2}" $KUBECTL
        ;;
```

bin/apps.bash
```diff
@@ -75,34 +73,50 @@ apps_sc() {
     # While it would be nice to have some template validation before `helmfile apply`,
     # at least Helmfile does "just in time" template validation. Not as nice,
     # but feels good enough until we figure out something smarter.
-    #
+    #    #[ "$1" != "--skip-template-validate" ] && template_validate_sc
+
```
Xartos, line 77: This can be removed
```suggestion
    #
```

bin/apps.bash
```diff
@@ -75,34 +73,50 @@ apps_sc() {
...
+    # The first few Charts install CRDs, which will make template validation
+    # fail. CRDs are "changes" to the Kubernetes API, hence validation against
+    # the Kubernetes API cannot be done. OTOH, manually adding the CRDs during
+    # bootstrap is error-prone and adds maintenance burden.
+    #
+    # While it would be nice to have some template validation before `helmfile apply`,
+    # at least Helmfile does "just in time" template validation. Not as nice,
+    # but feels good enough until we figure out something smarter.
+    #    #[ "$1" != "--skip-template-validate" ] && template_validate_sc
+
```
Xartos, line 98: Reverted to
```suggestion
    # See rationale above.
```

bin/apps.bash
```diff
@@ -75,34 +73,50 @@ apps_sc() {
...
     log_info "Applications applied successfully!"
 }
 
 #
 # ENTRYPOINT
 #
-
-
 if [[ $1 == "wc" ]]; then
+
```
Xartos, line 110: New line not needed
```suggestion
```

bin/apps.bash
```diff
@@ -75,34 +73,50 @@ apps_sc() {
...
     config_load "$1"
-    apps_wc "$2"
+    apps_wc "$2" "$3"
+
```
Xartos, line 113: New line not needed
```suggestion
```

bin/apps.bash
```diff
@@ -75,34 +73,50 @@ apps_sc() {
...
 elif [[ $1 == "sc" ]]; then
+
```
Xartos, line 115: New line not needed
```suggestion
```

bin/apps.bash
```diff
@@ -75,34 +73,50 @@ apps_sc() {
...
     config_load "$1"
-    apps_sc "$2"
+    apps_sc "$2" "$3"
 else
+
```
Xartos, line 119: New line not needed
```suggestion
```

bin/ck8s
```diff
@@ -33,6 +33,19 @@ usage() {
     exit 1
 }
 
+
```
Xartos, line 36: New line not needed
```suggestion
```

bin/dry-run.bash
```diff
@@ -21,14 +21,26 @@ if [[ $1 == "sc" ]]; then
...
   : "${config[config_file_sc]:?Missing service cluster config file}"
 
-  "${here}/ops.bash" helmfile sc diff
-
-elif [[ $1 == "wc" ]]; then
+  if [ ${#} -eq 2 ] && [ "$2" = "kubectl" ]; then
+    "${here}/ops.bash" helmfile sc template | "${here}/ops.bash" kubectl sc diff -f -
+  else
+    "${here}/ops.bash" helmfile sc diff
+
```
Xartos, line 28: New line not needed
```suggestion
```

bin/dry-run.bash
```diff
@@ -21,14 +21,26 @@ if [[ $1 == "sc" ]]; then
...
+  [[ $1 == "wc" ]]; then
   log_info "Running helmfile diff on the workload cluster"
   : "${config[config_file_wc]:?Missing workload cluster config file}"
 
-  "${here}/ops.bash" helmfile wc diff
+  if [ ${#} -eq 2 ] && [ "$2" = "kubectl" ]; then
+    "${here}/ops.bash" helmfile wc template | "${here}/ops.bash" kubectl wc diff -f -
+  else
+    "${here}/ops.bash" helmfile wc diff
+
```
Xartos, line 39: New line not needed
```suggestion
```

bin/dry-run.bash
```diff
@@ -21,14 +21,26 @@ if [[ $1 == "sc" ]]; then
...
+  fi
+
 else
+
```
Xartos, line 43: New line not needed
```suggestion
```

scripts/clean-wc.sh
```diff
@@ -57,7 +57,7 @@ PROM_CRDS=$(
         kubectl wc api-resources \
         --api-group=monitoring.coreos.com \
         -o name
-    )
+)
```
Xartos, line 60: Revert to
```suggestion
    )
```

scripts/deploy-sc.sh
```diff
@@ -10,30 +10,34 @@ source "${SCRIPTS_PATH}/../bin/common.bash"
...
 cd "${SCRIPTS_PATH}/../helmfile"
 declare -a helmfile_opt_flags
 [[ -n "$INTERACTIVE" ]] && helmfile_opt_flags+=("$INTERACTIVE")
-helmfile -f . -e service_cluster "${helmfile_opt_flags[@]}" apply --suppress-diff
+
+if [ ${#} -eq 1 ] && [ "$1" = "sync" ]; then
+    helmfile -f . -e service_cluster sync
+else
+    helmfile -f . -e service_cluster apply --suppress-diff
+
```
Xartos, line 40: New line not needed
```suggestion
```

Xartos (changes requested):
I only have some formatting issues now but otherwise this looks great!

robinelastisys:
Done! :smile:

bin/ck8s
```diff
@@ -33,6 +33,18 @@ usage() {
...
+SYNC=""
+SKIP=""
+KUBECTL=""
+
+for arg in "$@"; do
+  case "$arg" in
+    "--skip-template-validate") SKIP="--skip-template-validate" ;;
+    "--sync") SYNC="sync" ;;
+    "--kubectl") KUBECTL="kubectl" ;;
+esac
```
Xartos, line 45: Indentation
```suggestion
  esac
```

bin/dry-run.bash
```diff
@@ -21,13 +21,21 @@ if [[ $1 == "sc" ]]; then
...
-  "${here}/ops.bash" helmfile sc diff
-
-elif [[ $1 == "wc" ]]; then
+  if [ ${#} -eq 2 ] && [ "$2" = "kubectl" ]; then
+    "${here}/ops.bash" helmfile sc template | "${here}/ops.bash" kubectl sc diff -f -
+  else
+    "${here}/ops.bash" helmfile sc diff
+  fi
+elif
+  [[ $1 == "wc" ]]; then
```
Xartos, line 30: This doesn't need to be split to multiple lines
```suggestion
elif [[ $1 == "wc" ]]; then
```

bin/apps.bash
```diff
@@ -76,32 +74,39 @@ apps_sc() {
...
+    #
+    # The first few Charts install CRDs, which will make template validation
+    # fail. CRDs are "changes" to the Kubernetes API, hence validation against
+    # the Kubernetes API cannot be done. OTOH, manually adding the CRDs during
+    # bootstrap is error-prone and adds maintenance burden.
+    #
+    # While it would be nice to have some template validation before `helmfile apply`,
+    # at least Helmfile does "just in time" template validation. Not as nice,
+    # but feels good enough until we figure out something smarter.
+    #
```
Xartos, line 94: This could still be reverted to
```suggestion
    # See rationale above
```

To minimize duplications

bin/ck8s
```diff
@@ -43,20 +55,20 @@ case "${1}" in
         ;;
     apps)
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
-        "${here}/apps.bash" "${2}" "${3}"
+        "${here}/apps.bash" "${2}" $SKIP
```
Xartos, line 58: We should try to use {} and "" for all variables
```suggestion
        "${here}/apps.bash" "${2}" "${SKIP}"
```

bin/ck8s
```diff
@@ -43,20 +55,20 @@ case "${1}" in
...
         ;;
     apply)
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
         "${here}/bootstrap.bash" "${2}"
-        "${here}/apps.bash" "${2}" "${3}"
+        "${here}/apps.bash" "${2}" "$SKIP" "$SYNC"
```
Xartos, line 63: same here

bin/ck8s
```diff
@@ -43,20 +55,20 @@ case "${1}" in
...
         ;;
     test)
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
         "${here}/test.bash" "${2}"
         ;;
     dry-run)
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
-        "${here}/dry-run.bash" "${2}"
+        "${here}/dry-run.bash" "${2}" $KUBECTL
```
Xartos, line 71: same here

Xartos (changes requested):
Just the last few things :sweat_smile:

WIP-CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 - The sc-logs-retention cronjob now runs without error even if no backups were found for automatic removal
 - Harbor Swift authentication configuration options has moved from `citycloud` to `harbor.persistence.swift`.
+- The dry-run and apply command now checks against the state of the cluster while ran.
```
simonklb, line 10: Update this so that it reads as optional now since the functionality is behind optional flags.

bin/ck8s
```diff
@@ -13,9 +13,9 @@ usage() {
     echo "  init                                            initialize the config path" 1>&2
     echo "  bootstrap <wc|sc>                               bootstrap the cluster" 1>&2
     echo "  apps <wc|sc> [--skip-template-validate]         deploy the applications" 1>&2
-    echo "  apply <wc|sc> [--skip-template-validate]        bootstrap and apps" 1>&2
+    echo "  apply <wc|sc> [--sync] [--skip-template-validate] bootstrap and apps" 1>&2
     echo "  test <wc|sc>                                    test the applications" 1>&2
```
viktor-f, line 17:
```suggestion
    echo "  apply <wc|sc> [--sync] [--skip-template-validate] bootstrap and apps" 1>&2
    echo "  test <wc|sc>                                      test the applications" 1>&2
```
Nit: fix the indentation for the whole `usage` text (I can't mark all lines in this comment) so that it matches `bootstrap and apps`

bin/ck8s
```diff
@@ -43,20 +55,20 @@ case "${1}" in
         ;;
     apps)
         [[ "${2}" =~ ^(wc|sc)$ ]] || usage
-        "${here}/apps.bash" "${2}" "${3}"
+        "${here}/apps.bash" "${2}" "${SKIP}"
```
viktor-f, line 58: Shouldn't sync be available for this command as well?
```suggestion
        "${here}/apps.bash" "${2}" "${SKIP}" "${SYNC}"
```

bin/ck8s
```diff
@@ -13,9 +13,9 @@ usage() {
     echo "  init                                            initialize the config path" 1>&2
     echo "  bootstrap <wc|sc>                               bootstrap the cluster" 1>&2
     echo "  apps <wc|sc> [--skip-template-validate]         deploy the applications" 1>&2
```
viktor-f, line 15: Shouldn't sync be available for this command as well?
```suggestion
    echo "  apps <wc|sc> [--sync] [--skip-template-validate]  deploy the applications" 1>&2
```

OlleLarsson:
Please resolve some of the comments that's been fixed. A bit cumbersome to overview the pr ;)

bin/apps.bash
```diff
@@ -76,32 +74,30 @@ apps_sc() {
     # at least Helmfile does "just in time" template validation. Not as nice,
     # but feels good enough until we figure out something smarter.
     #
-    #[ "$1" != "--skip-template-validate" ] && template_validate_sc
```
simonklb, line 79: Keep this comment, it was commented out because we currently can't use it but not necessarily something we want to have disabled forever.

bin/apps.bash
```diff
@@ -76,32 +74,30 @@ apps_sc() {
...
+
+    apps_run_sc "${2:-""}"
 
     log_info "Applications applied successfully!"
 }
 
 apps_wc() {
     apps_init
-    # See rationale above.
-    #[ "$1" != "--skip-template-validate" ] && template_validate_wc
```
simonklb, line 88: Keep this comment, it was commented out because we currently can't use it but not necessarily something we want to have disabled forever.

simonklb (changes requested):

viktor-f (approved):
LGTM, well done!

simonklb (approved):
LGTM! Nice :)

Xartos (approved):
LGTM! Great work

PR #432, opened 2021-05-05
QA: change updateStrategy for harbor deployments

OlleLarsson:
In order to be able to update harbor smoothly we need to use the recreate updatestrategy as we don't have rwm volumes.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/helmfile/values/harbor.yaml.gotmpl
+++ b/helmfile/values/harbor.yaml.gotmpl
@@ -152,3 +152,6 @@ redis:
     nodeSelector: {{- toYaml .Values.harbor.nodeSelector | nindent 6 }}
     affinity:     {{- toYaml .Values.harbor.affinity | nindent 6 }}
     tolerations:  {{- toYaml .Values.harbor.tolerations | nindent 6 }}
+
+updateStrategy:
+  type: Recreate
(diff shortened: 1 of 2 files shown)


helmfile/values/harbor.yaml.gotmpl
```diff
@@ -152,3 +152,6 @@ redis:
     nodeSelector: {{- toYaml .Values.harbor.nodeSelector | nindent 6 }}
     affinity:     {{- toYaml .Values.harbor.affinity | nindent 6 }}
     tolerations:  {{- toYaml .Values.harbor.tolerations | nindent 6 }}
+
+updateStrategy:
+  type: Recreate
```
lentzi90, line 157: Does this apply to all components or just some? How about the database for example?

lentzi90 (approved):
Makes sense!

helmfile/values/harbor.yaml.gotmpl
OlleLarsson: Only the kubernetes deployments (not statefulsets), https://github.com/goharbor/harbor-helm/blob/master/values.yaml#L329-L330

PR #435, opened 2021-05-06
Rook monitoring

### LINKED ISSUE #381: [5] Monitoring Rook and Ceph
We use Rook and Ceph to provide persistent storage in environments where this is not available from the cloud provider. However, we gather no metrics about the Ceph cluster or Rook outside of the generic data we have on everything that runs in the cluster. There is nothing about the Ceph cluster health, remaining disk space or such. No alerts either.
In order to keep the storage clusters happy, we need insights into how they are performing and a way to detect issues early.

Configure Rook/Ceph to expose metrics and tell Prometheus to scrape them. Set up alerts for the most obvious things that we need to react to and maybe add a dashboard as well. [This should be a good place to start.](https://github.com/rook/rook/blob/master/Documentation/ceph-monitoring.md)

- Rely on existing, generic, data. With this it is possible to tell if Pods are crashing, how much resources they use and if disks on the Nodes are getting full. But these things can be hard to interpret and/or come too late. Available capacity on the Node disks does not translate directly to available capacity in Ceph, for example, since the data is replicated.

**Definition of done:**
Compliant Kubernetes has basic alerts for Ceph cluster health and a dashboard with an overview.

lentzi90:
This adds ServiceMonitors, PrometheusRules (alerts) and Grafana dashboards for monitoring Rook-Ceph. It is based on the docs [here](https://rook.io/docs/rook/v1.6/ceph-monitoring.html) which uses the resources [here](https://github.com/rook/rook/tree/master/cluster/examples/kubernetes/ceph/monitoring).

**Which issue this PR fixes**: fixes #381

The current monitoring configuration is starting to show its limits unfortunately. We run Rook both in SC and WC. We want alerts from both instances, but we should not push these alerts to the users (and probably not the metrics either).
I'm getting a bit lost in this and there is still somethings that doesn't quite work. Some help would be much appreciated.

This is how I understand it:

- We deploy kube-prometheus-stack in both SC and WC, but don't include the default PrometheusRules here since we have our own slightly tweaked version of them.
- We deploy `prometheus-alerts` (our own chart with the default PrometheusRules and some extra) 2 times in SC to be used with SC-prometheus and wc-reader.
- We also deploy `prometheus-alerts` in WC. There it provides metrics and alerts to the user, and metrics only to us. There is no way for us currently to separate these. I added a separate values template for this so it is at least separate from the wc-reader alerts we deploy in SC.

~~What is not working? I cannot seem to get metrics from WC to work in Grafana. They do get to prometheus in WC but I'm not sure about the rest. Maybe I'm missing some labels to make them picked up by wc-reader?~~

TODO:

- ~~Get metrics from WC working~~
- Tests? Or include in QA procedures?

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -666,3 +666,7 @@ vulnerabilityExporter:
   #   memory: 64Mi
   tolerations: []
   affinity: {}
+
+monitoring:
+  rook:
+    enabled: true
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -669,4 +669,4 @@ vulnerabilityExporter:
 
 monitoring:
   rook:
-    enabled: true
+    enabled: false
--- a/migration/v0.14.x-v0.15.x/upgrade-apps.md
+++ b/migration/v0.14.x-v0.15.x/upgrade-apps.md
@@ -9,8 +9,6 @@
 
 1. Delete `restore.*` from `sc-config.yaml`.
 
-1. Set `monitoring.rook.enabled` to `false` in both `sc-config.yaml` and `wc-config.yaml` if you don't want or need to monitor Rook.
-
 1. Upgrade applications:
     ```bash
     ./bin/ck8s apply {sc|wc}
--- a/migration/v0.14.x-v0.15.x/upgrade-apps.md
+++ b/migration/v0.14.x-v0.15.x/upgrade-apps.md
@@ -9,6 +9,8 @@
 
 1. Delete `restore.*` from `sc-config.yaml`.
 
+1. Set `monitoring.rook.enabled` to `false` in both `sc-config.yaml` and `wc-config.yaml` if you don't want or need to monitor Rook.
+
 1. Upgrade applications:
     ```bash
     ./bin/ck8s apply {sc|wc}
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -10,7 +10,7 @@
 - Added option to enable dex integration for ops grafana
 - Added resource request/limits for ops grafana
 - Added support for admin group for harbor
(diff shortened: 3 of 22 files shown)


viktor-f:
I'm not seeing anything that would explain your problems, but I could look at this with you tomorrow.

lentzi90:
Thanks to @viktor-f I got the WC metrics working! With this I think it is good enough to merge. :slightly_smiling_face: 
There is more work to be done to improve how the monitoring works though, so I'll create a follow up issue for it. The dashboards could also be improved to support multiple workload clusters. 1 of them already does this but the other do not have a "cluster" selector.

migration/v0.14.x-v0.15.x/upgrade-apps.md
```diff
@@ -9,6 +9,8 @@
 
 1. Delete `restore.*` from `sc-config.yaml`.
 
+1. Set `monitoring.rook.enabled` to `false` in both `sc-config.yaml` and `wc-config.yaml` if you don't want or need to monitor Rook.
```
Xartos, line 12: Move to 0.15->0.16 script

config/config/sc-config.yaml
```diff
@@ -612,3 +612,7 @@ s3Exporter:
   tolerations: []
   affinity: {}
   nodeSelector: {}
+
+monitoring:
+  rook:
+    enabled: true
```
Xartos, line 618: Shouldn't this be `false` by default?

Xartos (changes requested):

Xartos (commented):
reset request change
EDIT: didn't work ... I'll wait for approves and approve myself then

viktor-f (approved):

OlleLarsson (approved):

PR #436, opened 2021-05-06
Add Chart Upgrade issue and PR template

cristiklein:
As discussed during the architecture meeting on 2021-05-06, this implements a more orderly process to ensure we check migration steps when upgrading Charts.

### DIFF
--- a/.github/ISSUE_TEMPLATE/chart_upgrade.md
+++ b/.github/ISSUE_TEMPLATE/chart_upgrade.md
@@ -16,6 +16,7 @@ assignees: ''
 - [ ] We need to perform a security upgrade.
 - [ ] We need it for a new feature: *which feature (link to blocked issue)*.
 - [ ] We need to keep up to prevent future issues.
+- [ ] Other: *Write the reason here*
 
 **Acceptance criteria**
 
--- a/.github/ISSUE_TEMPLATE/chart_upgrade.md
+++ b/.github/ISSUE_TEMPLATE/chart_upgrade.md
@@ -11,6 +11,12 @@ assignees: ''
 
 /e.g., dexidp/dex 2.15/
 
+**Why should this Chart be upgraded?**
+
+- [ ] We need to perform a security upgrade.
+- [ ] We need it for a new feature: *which feature (link to blocked issue)*.
+- [ ] We need to keep up to prevent future issues.
+
 **Acceptance criteria**
 
 - I checked the migration of the new Chart:
--- a/.github/ISSUE_TEMPLATE/chart_upgrade.md
+++ b/.github/ISSUE_TEMPLATE/chart_upgrade.md
@@ -1,0 +1,19 @@
+---
+name: Chart Upgrade
+about: Request to change a Chart version
+title: ''
+labels: improvement
+assignees: ''
+
+---
+
+**What Chart should be upgraded?**
+
+/e.g., dexidp/dex 2.15/
+
+**Acceptance criteria**
+
+- I checked the migration of the new Chart:
+    - [ ] I upgraded a Chart and determined that no migration steps are needed.
+    - [ ] I upgraded a Chart and added migration steps to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md).
+- [ ] I tested the functionality provided by the new Chart (e.g., Auth flow, Grafana dashboards, etc.)
--- a/.github/ISSUE_TEMPLATE/chart_upgrade.md
+++ b/.github/ISSUE_TEMPLATE/chart_upgrade.md
@@ -21,5 +21,5 @@ assignees: ''
 
 - I checked the migration of the new Chart:
     - [ ] I upgraded a Chart and determined that no migration steps are needed.
-    - [ ] I upgraded a Chart and added migration steps to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md).
+    - [ ] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).
 - [ ] I tested the functionality provided by the new Chart (e.g., Auth flow, Grafana dashboards, etc.)
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -23,6 +23,10 @@
           E.g. removes or changes any CK8S config options or Kubernetes APIs.
     - [ ] will break the cluster.
           I.e. full cluster migration is required.
+- Chart checklist (pick exactly one):
+    - [ ] I upgraded no Chart.
+    - [ ] I upgraded a Chart and determined that no migration steps are needed.
+    - [ ] I upgraded a Chart and added migration steps to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md).
 
 **Pipeline config** *(if applicable)*
 If you change some config options (e.g. add/rename variable or change the default value) you may need to update the config used by the pipeline in `pipeline/config`.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -26,7 +26,7 @@
 - Chart checklist (pick exactly one):
     - [ ] I upgraded no Chart.
     - [ ] I upgraded a Chart and determined that no migration steps are needed.
-    - [ ] I upgraded a Chart and added migration steps to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md).
+    - [ ] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).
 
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/chart_upgrade.md
```diff
@@ -0,0 +1,19 @@
...
+
+**What Chart should be upgraded?**
+
+/e.g., dexidp/dex 2.15/
+
+**Acceptance criteria**
+
+- I checked the migration of the new Chart:
+    - [ ] I upgraded a Chart and determined that no migration steps are needed.
+    - [ ] I upgraded a Chart and added migration steps to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md).
```
viktor-f, line 18: I'd like all migration steps to be included in these documents instead. https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration/v0.14.x-v0.15.x/upgrade-apps.md

.github/pull_request_template.md
```diff
@@ -23,6 +23,10 @@
           E.g. removes or changes any CK8S config options or Kubernetes APIs.
     - [ ] will break the cluster.
           I.e. full cluster migration is required.
+- Chart checklist (pick exactly one):
+    - [ ] I upgraded no Chart.
+    - [ ] I upgraded a Chart and determined that no migration steps are needed.
+    - [ ] I upgraded a Chart and added migration steps to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md).
```
viktor-f, line 29: Same here

.github/ISSUE_TEMPLATE/chart_upgrade.md
```diff
@@ -0,0 +1,19 @@
...
+about: Request to change a Chart version
+title: ''
+labels: improvement
+assignees: ''
+
+---
+
+**What Chart should be upgraded?**
+
+/e.g., dexidp/dex 2.15/
```
viktor-f, line 12: Would it be good to also state why the chart should be upgraded? Mostly so we know if there is some new/changed feature that we should keep in mind while doing the task.
E.g. upgrade chart in order to get access to the new resource request helm variables, or upgrade chart to get latest application version.

Or do we just use these issues when we are upgrading for the sake of staying up to date and use the other issues when we are after a specific feature/change?

viktor-f (approved):

.github/ISSUE_TEMPLATE/chart_upgrade.md
```diff
@@ -0,0 +1,25 @@
...
+
+**Why should this Chart be upgraded?**
+
+- [ ] We need to perform a security upgrade.
+- [ ] We need it for a new feature: *which feature (link to blocked issue)*.
+- [ ] We need to keep up to prevent future issues.
```
Xartos, line 18:
```suggestion
- [ ] We need to keep up to prevent future issues.
- [ ] Other: *Write the reason here*
```

.github/ISSUE_TEMPLATE/chart_upgrade.md
```diff
@@ -0,0 +1,25 @@
...
+
+**Acceptance criteria**
+
+- I checked the migration of the new Chart:
+    - [ ] I upgraded a Chart and determined that no migration steps are needed.
+    - [ ] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).
+- [ ] I tested the functionality provided by the new Chart (e.g., Auth flow, Grafana dashboards, etc.)
```
Xartos, line 25: Why do this have a checkbox? I'm betting on a miss
Also all of these are written in perspective of the implementor, I would like them more objective.
e.g. `The functionality provided by the new Chart (e.g., Auth flow, Grafana dashboards, etc.) are tested`

Xartos (commented):
Minor comments

.github/ISSUE_TEMPLATE/chart_upgrade.md
cristiklein: I wrote "I" to make it clear that the person who takes the issue is responsible for performing these actions and should not expect QA or someone else to do this for for them. I see a risk that `The functionality provided by the new Chart (e.g., Auth flow, Grafana dashboards, etc.) are tested` creates vagueness on who should do what. Is it okay if the functionality was tested upstream? Can I rely on rumors that it worked on the demo cluster? Can I rely on someone in CNOSA having used this in a project? No, no and no! The one who picks the issue needs to put a bit more effort into ensuring no breakage.

This is my opinion. Since this is development process, @pettersv should have the final saying.

cristiklein:
Can we bump this? :smile:

.github/ISSUE_TEMPLATE/chart_upgrade.md
pettersv: I'm with Cristian here.

pettersv (approved):
LGTM!

simonklb:
I think it would be good to specifically mention CRDs in the issue template to make CRD upgrades etc a priority. Since [ADR-0011](https://github.com/elastisys/compliantkubernetes/pull/137) makes the CRDs more hidden from our view we need to be more vary about what goes on in the upstream projects.

cristiklein:
@simonklb First, CRDs are actually more visible now, since we "vendor-in" all upstream Charts. I added a code snippet in [ADR-0011](https://compliantkubernetes.io/adr/0011-let-upstream-projects-handle-crds/#decision-outcome) to illustrate how to discover them. (BTW, I just learned that starboard-operator adds CRDs. :smile:)

Second, I don't like to put too much focus on CRDs. People should really read the migration guide, the whole migration guide and nothing but the migration guide -- so help me Codd :smile: -- of the upstream project, without having to dig into the Charts to find out what exactly has changed and what is the potential impact. For projects like Velero, worrying about CRDs is not necessary, since they added smartness to their Chart. For projects like OPA, CRDs can be an issue and they are careful about adding a migration guide for that.

So what if a project is sloppy with CRDs? Well, then that project poses a risk to our operation and should ideally not be used.

simonklb:
Hope my worries are unfounded. :smile: A sudden upgrade causing CRDs to be updated or worse yet deleted resulting in data loss would be bad. Can't really blame the open source project in that case. A upgrade e2e test with data corruption checks would make me sleep better. At least we do some QA.

cristiklein:
> Hope my worries are unfounded. 

Bringing up risks, so we can systematically decide whether to tolerate, address or outsource them, is highly appreciated and even required as part of our ISO 27001 certification. :smile: 

> A sudden upgrade causing CRDs to be updated or worse yet deleted resulting in data loss would be bad.

While we should do everything possible to prevent something like this, if we cannot trust on [disaster recovery](https://compliantkubernetes.io/operator-manual/disaster-recovery/), then CRDs is the smallest of our problem. My understanding is that we regularly test restoration, but Petter knows more. If you find gaps in our DR strategy, please raise GitHub issues or bring them up in one of our forums.

> Can't really blame the open source project in that case. A upgrade e2e test with data corruption checks would make me sleep better. At least we do some QA.

My understanding is that QA always tests at least one migration from a previous version and at least one new cluster setup. So QA should cover us when it comes to "loss due to CRD upgrade issues".

Xartos:
> While we should do everything possible to prevent something like this, if we cannot trust on disaster recovery, then CRDs is the smallest of our problem. My understanding is that we regularly test restoration, but Petter knows more. If you find gaps in our DR strategy, please raise GitHub issues or bring them up in one of our forums.

I don't think we backup any of the CRDs in the cluster. But I don't think that there are any CRDs that can't be recreated by just using our config so in a DR scenario you would just redeploy the old version and hope that the failing CRDs are overwritten. But I don't think this scenario is tested at all.

cristiklein:
> > While we should do everything possible to prevent something like this, if we cannot trust on disaster recovery, then CRDs is the smallest of our problem. My understanding is that we regularly test restoration, but Petter knows more. If you find gaps in our DR strategy, please raise GitHub issues or bring them up in one of our forums.
> 
> I don't think we backup any of the CRDs in the cluster. But I don't think that there are any CRDs that can't be recreated by just using our config so in a DR scenario you would just redeploy the old version and hope that the failing CRDs are overwritten. And from what I know this scenario isn't tested at all.

@Xartos What about Velero backups? Don't we back up all Kubernetes resources there? Should we allocate some time to re-test this?

Xartos:
> > > While we should do everything possible to prevent something like this, if we cannot trust on disaster recovery, then CRDs is the smallest of our problem. My understanding is that we regularly test restoration, but Petter knows more. If you find gaps in our DR strategy, please raise GitHub issues or bring them up in one of our forums.
> > 
> > 
> > I don't think we backup any of the CRDs in the cluster. But I don't think that there are any CRDs that can't be recreated by just using our config so in a DR scenario you would just redeploy the old version and hope that the failing CRDs are overwritten. And from what I know this scenario isn't tested at all.
> 
> @Xartos What about Velero backups? Don't we back up all Kubernetes resources there? Should we allocate some time to re-test this?

We only back up objects that are labeled `velero: backup`, see https://compliantkubernetes.io/user-guide/backup/#backing-up
So the customers need to make sure to set that on all objects they want to backup

PR #444, opened 2021-05-11
Bring in upstream charts

### LINKED ISSUE #355: Move away from "pulling" charts

cristiklein:

Compliant Kubernetes operators complained about a few things:

* It is difficult to follow the path that configs take. We cannot
"grep" a value (see https://github.com/elastisys/compliantkubernetes-apps/issues/355)

* We depend on upstream repos being up, which is impacts us when they
are down.

* `apply` and `template` take too much time.

This commit fixes all above, by bringing in upstream Charts.

fixes #355

Change should not be visible, neither to the operator, nor the user.

I tested the change as follows:

```bash
#!/bin/sh

git checkout vendor-in-upstream-charts

for c in service_cluster workload_cluster; do
    helmfile -e $c -f helmfile template > ${c}_after.yaml
done

git checkout main

for c in service_cluster workload_cluster; do
    helmfile -e $c -f helmfile template > ${c}_before.yaml
done


for c in service_cluster workload_cluster; do
    diff ${c}_before.yaml ${c}_after.yaml
done
```

Which gave the output:

```
<     kubeapiuser:$2a$10$9gRqlxCboSCs2PJlvZydJeJH6OyVOxDJfDkQ9M/lN3EXL79l57J2.
---
>     kubeapiuser:$2a$10$QwwJnJd8MIXTOXnGt.KtH.YMdWr1tyhdVGhheavRFc60Nj9UhMoAW
```

Hence, there is some non-determinism in how we generate Kubernetes resources, but at least this PR introduces no bugs.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

NA (I hope)


Xartos (commented):
> It is difficult to follow the path that configs take. We cannot
"grep" a value (see #355)

Yes, and no. This also adds complexity to the "grep" output and a common word might drown the user in lines. I guess this is mostly a matter of filtering your search. But I see some benefit to be able to do `grep -r` from the project root and have a bitesized result that I can work with to find out that I need to continue to a upstream chart and grep there instead.

*But this might just be my personal preference*

> We depend on upstream repos being up, which is impacts us when they
are down.

Not 100% true since the charts are cached locally. But they are definitely a dependency that may impact us.

> `apply` and `template` take too much time.

Is this really an issue? We might shave off from 10 -> 1 second which would only be ran once. Then you can run with `--skip-deps` and it will use cached charts and it's super quick.


I would prefere if we used the `./` notation on the charts. Makes it super clear that it's sourced locally. E.g. `upstream/dex` -> `./upstream/dex`

Super nice that you added the snippet with how to add a repo (did not know you could do this), makes it almost as easy to update the charts as it is now. And I must also say that it's great that you managed to do this change and bring 0 diffs. Would make it a super smooth change if we decide to go with it :slightly_smiling_face: 

TL;DR
I'm not sure I see the big benefit this would bring and that it outweighs the downsides.

cristiklein:
> > We depend on upstream repos being up, which is impacts us when they
> > are down.
> 
> Not 100% true since the charts are cached locally. But they are definitely a dependency that may impact us.

Just wanted to note that we also hedge against repos changing URL or getting deprecated. (Although, this can also be seen as a lazy excuse to not upgrade. :smile:)

> Is this really an issue? We might shave off from 10 -> 1 second which would only be ran once. Then you can run with --skip-deps and it will use cached charts and it's super quick.

Hmm ... This feels like one of those "contributor tips and tricks" that should be lifted up.

> TL;DR
> I'm not sure I see the big benefit this would bring and that it outweighs the downsides.

With the `--skip-deps` trick, I share your opinion. I'll think about it and let other weigh in in the meanwhile.

simonklb:
One benefit this would bring is that updating vendored charts would show the full diff compared to just a small version bump and whatever is necessary to support it. This also makes this project closer to being air-gapped which would be nice to see in the future.

The biggest downside I see here would be the cluttering of the repository. I think this could be fairly easy to resolve if we put all of the vendored files in a folder that could be easily excluded when grepping. However, as @cristiklein stated above, with all of the charts vendored you do have the possibility to grep upstream projects as well (without having to `helm pull` or whatever), which I think is a win here.

Just my 2¢

cristiklein:
As discussed during the stand-up, we will move forward with this and I offer to reverse it, if it proves to be a bad idea.

@Xartos @viktor-f Can I get your GitHub seal of approval? :smile:

Xartos (approved):
IMG: Seal of approval

simonklb (approved):
Obviously haven't made sure all the newly added files are correct but a `git diff main --diff-filter=a` LGTM!

One thing that might have been nice to have would be a helper script to vendor things or documentation on how to do it. But that is not a blocker imo.

cristiklein:
> One thing that might have been nice to have would be a helper script to vendor things or documentation on how to do it. But that is not a blocker imo.

Just to make, have you seen this: https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/upstream/README.md

Could you check if I need to add more information?

simonklb:
That's fine! Tried it by first deleting falco, ran the commands and the diff is empty. :+1:

PR #450, opened 2021-05-19
[1] Added support for OIDC for ops grafana

Xartos:

Since we want to use dex for everything I added support for ops grafana

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -194,6 +194,16 @@ prometheus:
     tolerations: []
     affinity: {}
     nodeSelector: {}
+  grafana:
+    oidc:
+      enabled: true
+      userGroups:
+        grafanaAdmin: grafana_admin #maps to grafana role admin
+        grafanaEditor: grafana_editor #maps to grafana role editor
+        grafanaViewer: grafana_viewer #maps to grafana role viewer
+      scopes: "openid profile email groups"
+      allowedDomains:
+        - example.com
 
 dex:
   additionalKubeloginRedirects: []
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -195,15 +195,25 @@ prometheus:
     affinity: {}
     nodeSelector: {}
   grafana:
+    resources:
+      requests:
+        cpu: 100m
+        memory: 128Mi
+      limits:
+        cpu: 100m
+        memory: 128Mi
+    tolerations: []
+    affinity: {}
+    nodeSelector: {}
     oidc:
-      enabled: true
-      userGroups:
-        grafanaAdmin: grafana_admin #maps to grafana role admin
-        grafanaEditor: grafana_editor #maps to grafana role editor
-        grafanaViewer: grafana_viewer #maps to grafana role viewer
-      scopes: "openid profile email groups"
-      allowedDomains:
-        - example.com
+      enabled: false
+      # Only used if the above is true
+      # userGroups:
+      #   grafanaAdmin: grafana_admin #maps to grafana role admin
+      #   grafanaEditor: grafana_editor #maps to grafana role editor
+      #   grafanaViewer: grafana_viewer #maps to grafana role viewer
+      # scopes: "openid profile email groups"
+      # allowedDomains: []
 
 dex:
   additionalKubeloginRedirects: []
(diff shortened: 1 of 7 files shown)


config/config/sc-config.yaml
```diff
@@ -190,6 +190,16 @@ prometheus:
     tolerations: []
     affinity: {}
     nodeSelector: {}
+  grafana:
```
viktor-f, line 193: While you are at it. Please add resources, tolerations, and affinity here. Preferably with defaults for the resource requests.

viktor-f (commented):
Looks good!

config/config/sc-config.yaml
```diff
@@ -190,6 +190,16 @@ prometheus:
...
+    oidc:
+      enabled: true
+      userGroups:
+        grafanaAdmin: grafana_admin #maps to grafana role admin
+        grafanaEditor: grafana_editor #maps to grafana role editor
+        grafanaViewer: grafana_viewer #maps to grafana role viewer
+      scopes: "openid profile email groups"
+      allowedDomains:
+        - example.com
```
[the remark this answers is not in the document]
Xartos: Dex will always use all the available connectors so it's still an issue. I guess we could set up multiple dex instances one for ops and one for users. But then you wouldn't get SSO for kibana both for ops and users.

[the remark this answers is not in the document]
Xartos: Sorry, I'm thinking of the one for the connector...
Yes, this should prevent users to login. Good catch :+1:

config/config/sc-config.yaml
viktor-f: Thanks for adding resources, though I'm still missing tolerations and affinity. Though it's not a blocker if you have a reason not to add it.

pipeline/config/exoscale/sc-config.yaml
```diff
@@ -184,6 +184,16 @@ prometheus:
...
+  grafana:
+    resources:
+      requests:
+        cpu: 100m
+        memory: 128Mi
+      limits:
+        cpu: 100m
+        memory: 128Mi
+    oidc:
+      enabled: false
```
viktor-f, line 196: Have you tested that the chart will install ok without specifying the other oidc values? I'm thinking that it might complain that some values are missing.

Xartos: You are correct, that didn't work. I'll add some defaults for those as well

config/config/sc-config.yaml
Xartos: Right, I missed that you wrote that. Yea, I'll add it

viktor-f (approved):
LGTM

helmfile/values/kube-prometheus-stack-sc.yaml.gotmpl
```diff
@@ -214,11 +214,30 @@ grafana:
...
+      {{- end }}
 
   serviceMonitor:
     relabelings:
     - targetLabel: cluster
       replacement: service_cluster
+  resources: {{- toYaml .Values.prometheus.grafana.resources | nindent 4  }}
+  tolerations: {{- toYaml .Values.prometheus.grafana.tolerations | nindent 4  }}
+  affinity: {{- toYaml .Values.prometheus.grafana.affinity | nindent 4  }}
+  nodeSelector: {{- toYaml .Values.prometheus.grafana.nodeSelector | nindent 4  }}
```
simonklb, line 240: Just to make sure, will this work if you upgrade without adding the values to your config? If not mention it in the changelog or add a migration script?

simonklb (approved):
I haven't actually tested this, but the changes looks good! If you want a live test before you merge let me know.

Other than the comment I approve! :smile:

helmfile/values/kube-prometheus-stack-sc.yaml.gotmpl
viktor-f: `ck8s init` is part of the migration guide, which will add these.

Xartos: With the new defaults (that comes after you run init, and thus part of the migration script) it should not make any diff except adding the request/limit. which is stated in the changelog :slightly_smiling_face:

simonklb: This is similar to the last discussion. I think the migration guide should probably be highlighted more to non-Elastisys users.

PR #451, opened 2021-05-20
[1] Rework of dex connectors config

### LINKED ISSUE #448: [1] Make it possible to use multiple dex connectors

Since you might want to have multiple connectors connected to dex we should enable config for that.

When setting config for dex, there should be an array of connectors that you want to add instead of only one choice.

Not beeing able to configure multiple connectors

**Additional context**

**Definition of done:**

It's possible to configure multiple dex connectors

Xartos:

Rework of how we configure dex connectors to enable multiple connectors.

fixes #448

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,3 +1,14 @@
-Changed:
+### Release notes
 
-* The project now requires `helm-diff >= 3.1.2`. Remove the old one (via `rm -rf ~/.local/share/helm/plugins/helm-diff/`), before reinstalling dependencies.
+- Support for multiple connectors for dex and better support for OIDC groups.
+- Check out the [upgrade guide](migration/v0.15.x-v0.16.x/upgrade-apps.md) for a complete set of instructions needed to upgrade.
+
+### Changed
+
+- The project now requires `helm-diff >= 3.1.2`. Remove the old one (via `rm -rf ~/.local/share/helm/plugins/helm-diff/`), before reinstalling dependencies.
+- Changed the way connectors are provided to dex
+
+### Fixed
+
+- Fixed issue where you couldn't configure dex google connector to support groups
+- Fixed issue where groups wouldn't be fetched for kubelogin
--- a/bin/kubeconfig.bash
+++ b/bin/kubeconfig.bash
@@ -93,7 +93,8 @@ kubectl --kubeconfig="${user_kubeconfig}" config set-credentials "${1}@${cluster
     --exec-arg=--oidc-issuer-url="https://dex.${base_domain}" \
     --exec-arg=--oidc-client-id=kubelogin \
     --exec-arg=--oidc-client-secret="$(sops -d --extract '["dex"]["kubeloginClientSecret"]' "${secrets[secrets_file]}")" \
-    --exec-arg=--oidc-extra-scope=email
+    --exec-arg=--oidc-extra-scope=email \
+    --exec-arg=--oidc-extra-scope=groups
 
 # Create context with relavant namespace
 # Pick the first namespace
--- a/migration/v0.15.x-v0.16.x/migrate-apps.sh
+++ b/migration/v0.15.x-v0.16.x/migrate-apps.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+set -euo pipefail
+
+: "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
+
+here="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
+
+config="${CK8S_CONFIG_PATH}/sc-config.yaml"
+secret="${CK8S_CONFIG_PATH}/secrets.yaml"
+secret_tmp="${CK8S_CONFIG_PATH}/secrets_tmp.yaml"
+sops_conf="${CK8S_CONFIG_PATH}/.sops.yaml"
+
+# Fetch old variables
+base_domain=$(yq r "$config" 'global.baseDomain')
+dex_oidcProvider=$(yq r "$config" 'dex.oidcProvider')
+declare -a dex_allowedDomains
+# dex_allowedDomains=( $(yq r "$config" 'dex.allowedDomains[*]') )
+mapfile -t dex_allowedDomains < <(yq r "$config" 'dex.allowedDomains[*]')
+dex_insecureSkipEmailVerified=$(yq r "$config" 'dex.insecureSkipEmailVerified')
+dex_insecureEnableGroups=$(yq r "$config" 'dex.insecureEnableGroups')
+dex_redirect="https://dex.${base_domain}/callback"
+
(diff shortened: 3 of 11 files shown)


WIP-CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+### Changed
```
simonklb, line 1: Mention the migration script in "release notes" or is "run migration scripts" always implied in the upgrade process?

bin/kubeconfig.bash
```diff
@@ -93,7 +93,8 @@ kubectl --kubeconfig="${user_kubeconfig}" config set-credentials "${1}@${cluster
     --exec-arg=--oidc-issuer-url="https://dex.${base_domain}" \
     --exec-arg=--oidc-client-id=kubelogin \
     --exec-arg=--oidc-client-secret="$(sops -d --extract '["dex"]["kubeloginClientSecret"]' "${secrets[secrets_file]}")" \
-    --exec-arg=--oidc-extra-scope=email
+    --exec-arg=--oidc-extra-scope=email \
+    --exec-arg=--oidc-extra-scope=groups
```
simonklb, line 97: This is new right? Worth mentioning in changelog?

simonklb (changes requested):
Cool seeing group support being introduced, I think this should be highlighted! :smile:

WIP-CHANGELOG.md
Xartos: It's part of the migration document https://github.com/elastisys/compliantkubernetes-apps/blob/f1bd76dcd40ce4940043e37dc9b132ae2ea5e691/migration/v0.15.x-v0.16.x/upgrade-apps.md which is where we have all these instructions :slightly_smiling_face:

bin/kubeconfig.bash
Xartos: I guess it's worth menitoning, this however will be completely transparent since if you don't have any groups from your IdP nothing will change and if you have groups you simply get it as expected.

WIP-CHANGELOG.md
Xartos: Good question, we seem to sometimes have it at least https://github.com/elastisys/compliantkubernetes-apps/blob/main/CHANGELOG.md#release-notes
But we should definitely always have that in the release notes. Good catch

migration/v0.15.x-v0.16.x/migrate-apps.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
simonklb, line 1: chmod +x ./migration/v0.15.x-v0.16.x/migrate-apps.sh

simonklb (approved):
LGTM! Just need to make the migration script executable.

Tested the migration script with from a clean init (i.e. can't say for sure it works on existing clusters with config modifications):
```
% git checkout v0.15.0
% ./bin/ck8s init
% git checkout fli/oidc-connectors-rework-main
% ./migration/v0.15.x-v0.16.x/migrate-apps.sh
```

sc-config.yaml diff:
```
@ sc-config.yaml:188 @ prometheus:
     affinity: {}
     nodeSelector: {}
 dex:
-  # supported: google|okta|aaa
-  oidcProvider: google
-  allowedDomains:
-    - example.com
   additionalKubeloginRedirects: []
   enableStaticLogin: true
   resources:
@ sc-config.yaml:200 @ dex:
   tolerations: []
   affinity: {}
   nodeSelector: {}
-  # These two settings are curently only used by okta. If you don't need the groups claims, both options can be set to 'false'.
-  # 'insecureSkipEmailVerified' should only be set to 'true' if the user's okta
-  # configuration does not require a user to verify their identity to okta when their okta account is being created. This is not recommended.
-  insecureSkipEmailVerified: false
-  insecureEnableGroups: true
```

secrets diff:
```
45,48d44
<     googleClientID: somelongsecret
<     googleClientSecret: somelongsecret
<     oktaClientID: somelongsecret
<     oktaClientSecret: somelongsecret
50c46,56
<     issuer: somelongsecret
---
>     connectors:
>     -   config:
>             clientID: somelongsecret
>             clientSecret: somelongsecret
>             issuer: https://accounts.google.com
>             redirectURI: https://dex.set-me/callback
>             hostedDomains:
>             - example.com
>         id: google
>         name: Google
>         type: oidc
```

Xartos:
> Tested the migration script with from a clean init (i.e. can't say for sure it works on existing clusters with config modifications):

From my testing i have gotten the exakt same diff and (at least google connectors) gives a non diff when running `bin/ck8s ops helmfile sc -l app=dex diff --skip-deps`

viktor-f (approved):
Very nice!

Re. the release notes: maybe we should include in the release scripts to always create the base migration script and add it to the release notes.

simonklb:
> Very nice!
> 
> Re. the release notes: maybe we should include in the release scripts to always create the base migration script and add it to the release notes.

Yea, I think that is a good idea! The only issue I see is that you might get a duplicate if someone already added it to the release notes. :laughing: Maybe if the note is something generic, it could be like:
* Check if migration scripts exists
* Add a release note "Migration scripts are required to upgrade the cluster. See X."

Or something like that. Then if someone added something more specific in the release note regarding the migration script(s) it won't look too bad.

viktor-f:
> Yea, I think that is a good idea! The only issue I see is that you might get a duplicate if someone already added it to the release notes. laughing Maybe if the note is something generic, it could be like:
> 
>     * Check if migration scripts exists
> 
>     * Add a release note "Migration scripts are required to upgrade the cluster. See X."
> 
> 
> Or something like that. Then if someone added something more specific in the release note regarding the migration script(s) it won't look too bad.

Not sure we are talking about the exact same solution. I was thinking that we could add it to the release process, so that whenever we create a new release and clean the WIP-CHANGELOG then we also add the generic release note and a basic migration script. Then we should never get duplicate scripts and it will always be there.

simonklb:
> > Yea, I think that is a good idea! The only issue I see is that you might get a duplicate if someone already added it to the release notes. laughing Maybe if the note is something generic, it could be like:
> > ```
> > * Check if migration scripts exists
> > 
> > * Add a release note "Migration scripts are required to upgrade the cluster. See X."
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or something like that. Then if someone added something more specific in the release note regarding the migration script(s) it won't look too bad.
> 
> Not sure we are talking about the exact same solution. I was thinking that we could add it to the release process, so that whenever we create a new release and clean the WIP-CHANGELOG then we also add the generic release note and a basic migration script. Then we should never get duplicate scripts and it will always be there.

Ah, gotcha! I think that could be nice. Then I think we should always try to capture what is required in the migration script. Even if it's a simple apply so that the only thing someone needs to run is that single "main" migration script.

PR #462, opened 2021-05-25
apps sc: update node-local-dns chart

### LINKED ISSUE #227: [1] Update nodelocaldns chart to match upstream
Since we've had some issues with node-local-dns for external DNS lookup *(I dobt this is the issue, but it might)* we might want to update it to match how current clusters would set it up. 

Update [our chart](https://github.com/elastisys/compliantkubernetes-apps/tree/main/helmfile/charts/node-local-dns) to match [the upstream manifest](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml)
One notable feature is that they have removed the `force_tcp` option on the external DNS which supposedly can improve performance [see this PR](https://github.com/kubernetes/kubernetes/pull/93471) 

* Disable node-local-dns from apps and start relying on kubespray to set it up.
* Stick with our current version of node-local-dns.

**Additional context**
None

**Definition of done:**
After deploying apps the node-local-dns should match [the upstream manifest](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml)

robinelastisys:
Since we've had some issues with node-local-dns for external DNS lookup we may want to update it to see if it works better.

**Which issue this PR fixes** 
fixes #227  

**Checklist:**
- [x] Some test is done to check if the problem still persists.
    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -17,6 +17,7 @@
 - Changed the way connectors are provided to dex
 - Default retention values for other* and authlog* are changed to fit the needs better
 - CK8S version validation accepts version number if exactly at the release tag, otherwise commit hash of current commit. "any" can still be used to disable validation.
+- The node-local-dns chart have been updated to match the upstream manifest. force_tcp have been removed to improve performence and the container image have beve been updated from 1.15.10 to 1.17.0.
 
 ### Fixed
 
--- a/helmfile/charts/node-local-dns/templates/node-local-dns.yaml
+++ b/helmfile/charts/node-local-dns/templates/node-local-dns.yaml
@@ -52,9 +52,6 @@ metadata:
 data:
   Corefile: |
     cluster.local:53 {
-        log {
-          class error
-        }
         errors
         cache {
                 success 9984 30
@@ -70,9 +67,6 @@ data:
         health {{ .Values.localIP }}:8080
         }
     in-addr.arpa:53 {
-        log {
-          class error
-        }
         errors
         cache 30
         reload
@@ -84,9 +78,6 @@ data:
         prometheus :9253
         }
     ip6.arpa:53 {
-        log {
-          class error
-        }
         errors
         cache 30
         reload
@@ -98,17 +89,12 @@ data:
         prometheus :9253
         }
     .:53 {
-        log {
-          class error
-        }
         errors
         cache 30
         reload
         loop
         bind {{ .Values.localIP }} {{ .Values.clusterDNS }}
-        forward . __PILLAR__UPSTREAM__SERVERS__ {
-                force_tcp
-        }
+        forward . __PILLAR__UPSTREAM__SERVERS__
         prometheus :9253
         }
 ---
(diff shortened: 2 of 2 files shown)


WIP-CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 - The project now requires `helm-diff >= 3.1.2`. Remove the old one (via `rm -rf ~/.local/share/helm/plugins/helm-diff/`), before reinstalling dependencies.
 - Changed the way connectors are provided to dex
+- node-local-dns update
```
simonklb, line 10: This needs to be more descriptive, updates what?

helmfile/charts/node-local-dns/templates/node-local-dns.yaml
```diff
@@ -98,24 +92,20 @@ data:
...
-        }
         errors
         cache 30
         reload
         loop
         bind {{ .Values.localIP }} {{ .Values.clusterDNS }}
-        forward . __PILLAR__UPSTREAM__SERVERS__ {
-                force_tcp
-        }
+        forward . __PILLAR__UPSTREAM__SERVERS__
```
simonklb, line 100: Is this change based on something upstream? Making changes like this deserves at least some explanation as well as a version bump of the chart.

simonklb (changes requested):
Please stick to the [PR template](https://github.com/elastisys/compliantkubernetes-apps/blob/main/.github/pull_request_template.md), "Some check is done to see if it works" doesn't really tell me anything about what this change implies.

helmfile/charts/node-local-dns/templates/node-local-dns.yaml
```diff
@@ -17,6 +17,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: node-local-dns
+  namespace: kube-system
```
viktor-f, line 20: We should not add namespaces like this in the helm charts. Instead we let helm add the namespace
```suggestion
```
Same for the rest of the files

viktor-f (changes requested):

helmfile/charts/node-local-dns/templates/node-local-dns.yaml
Xartos: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml

Xartos (changes requested):
This still doesn't include the headless service that the upstream chart does.

See https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/nodelocaldns/nodelocaldns.yaml#L190-L209

robinelastisys:
Oh, wasn't sure we needed that as we didn't have it before :+1:

helmfile/charts/node-local-dns/templates/node-local-dns.yaml
robinelastisys: Apparently it can improve performance, https://github.com/kubernetes/kubernetes/pull/93471

helmfile/charts/node-local-dns/templates/node-local-dns.yaml
robinelastisys: https://github.com/elastisys/compliantkubernetes-apps/blob/74788e09b707313a6f2e9bc2ff2fa3815c385279/bootstrap/namespaces/helmfile/values/namespaces-sc.yaml.gotmpl should be enough right?

helmfile/charts/node-local-dns/templates/node-local-dns.yaml
simonklb: Thanks!

helmfile/charts/node-local-dns/templates/node-local-dns.yaml
```diff
@@ -186,3 +180,23 @@ spec:
           items:
             - key: Corefile
               path: Corefile.base
+
```
Xartos, line 183: Since this is another object it should be a separate document (i.e. separated with `---`)
```suggestion
---
```

Xartos (changes requested):
Nice! Just one last thing from my side

Xartos (approved):
LGTM!

simonklb (approved):
Nice!

viktor-f (approved):

PR #464, opened 2021-05-25
apps sc & wc: fix ck8s version validate

### LINKED ISSUE #189: [5] Fix ck8s version validate
The current version validation does not work anymore as it [tries to find the latest tag on main](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/common.bash#L70-L74). The problem now is that we do not have all tags on main. The current way to get around it is to use `global.version: any` in the configs.

I don't know how this should be fixed but the goal is to be able to specify a tag and not just `any` in the configs.

robinelastisys:
The version validation now checks if you are at the exact tag/branch rather then just the latest tag.

**Which issue this PR fixes** 
fixes #189 

**Checklist:**
- [x]  Some test is done to check if the problem still persists.
    - [x] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -16,6 +16,7 @@
 - The project now requires `helm-diff >= 3.1.2`. Remove the old one (via `rm -rf ~/.local/share/helm/plugins/helm-diff/`), before reinstalling dependencies.
 - Changed the way connectors are provided to dex
 - Default retention values for other* and authlog* are changed to fit the needs better
+- CK8S version validation accepts version number if exactly at the release tag, otherwise commit hash of current commit. "any" can still be used to disable validation.
 
 ### Fixed
 
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -1,4 +1,8 @@
 global:
+  ## Compliantkubernetes-apps version.
+  ## Use version number if you are exactly at a release tag.
+  ## Otherwise use full commit hash of current commit.
+  ## 'any', can be used to disable this validation.
   ck8sVersion: ${CK8S_VERSION}
   cloudProvider: ${CK8S_CLOUD_PROVIDER}
   clusterName: ${CK8S_ENVIRONMENT_NAME}-sc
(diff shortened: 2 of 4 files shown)


config/config/sc-config.yaml
```diff
@@ -1,4 +1,7 @@
 global:
+  ## Compliantkubernetes-apps version.
+  ## You need to be at the exact branch for the version number to match.
+  ## 'any', can be used when running on non-official release.
```
viktor-f, line 4:
```suggestion
  ## Compliantkubernetes-apps version.
  ## Use version number if you are exactly at a release tag.
  ## Otherwise use full commit hash of current commit.
  ## 'any', can be used to disable this validation.
```
I'd prefer something like this. Also update the other file.

WIP-CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 
 - The project now requires `helm-diff >= 3.1.2`. Remove the old one (via `rm -rf ~/.local/share/helm/plugins/helm-diff/`), before reinstalling dependencies.
 - Changed the way connectors are provided to dex
+- ck8s version validator check for the exact branch for the version number to match with the config.
```
viktor-f, line 18: Maybe update to something like this, to make it clearer.
```suggestion
- CK8S version validation accepts version number if exactly at the release tag, otherwise commit hash of current commit.
  "any" can still be used to disable validation.
```

viktor-f (changes requested):

viktor-f (approved):

simonklb (changes requested):
Don't edit the PR template. Fill it out instead.

robinelastisys:
> Don't edit the PR template. Fill it out instead.

Fixed! :+1:

simonklb:
> > Don't edit the PR template. Fill it out instead.
> 
> Fixed! +1

You checked:
```
will change any APIs.
```
This checkbox should be treated more like "the configuration/API has changed in a way where it's no longer backwards compatible" and this should almost always be mention as a critical release note. While it's true that you have modified the way a configuration value is interpreted the old configuration value will still be valid as long as the user runs from a specific tag which should be the main use case. You should never run from a non-tagged commit if you are not very sure what you are doing or developing. :smile: 

Only mentioning this for knowledge sharing purposes, not as a blocker.

simonklb (approved):

PR #480, opened 2021-06-02
apps: Updated nfs provisioner to the new one

Xartos:

Replaced depricated nfs-client-provisioner to the new [nfs-subdir-external-provisioner](https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner/tree/master/charts/nfs-subdir-external-provisioner)

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,0 +1,3 @@
+### Release notes
+
+- Changed from depricated nfs provisioner to the new one. Migration is automatic (no manual intervention)
--- a/bootstrap/storageclass/helmfile/helmfile.yaml
+++ b/bootstrap/storageclass/helmfile/helmfile.yaml
@@ -1,6 +1,9 @@
 repositories:
+  # TODO remove when the nfs-client-provisioner is removed
 - name: stable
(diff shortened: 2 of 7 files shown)


OlleLarsson:
I guess that we kind of need a migration path from the nfs-client-provisioner to this new chart for those who happen to run it.

Xartos:
> I guess that we kind of need a migration path from the nfs-client-provisioner to this new chart for those who happen to run it.

I must test this, but I think that this should work out of the box. Since they are both using nfs, so if the release gets reinstalled then I think this new deployment should just grab the volumes present in the backing NFS. You might need to manually uninstall the old release and then install this new one. But again, I'm gonna test if it's this easy.

Xartos:
Tested with the stuff I just pushed and the migration is seamless. The volumes doesn't need to be recreated or anything.

OlleLarsson (approved):
Great!

viktor-f (approved):
I guess that we probably should change this helmfile to also use vendored copies of the upstream charts. But I'm ok if that is done in a separate issue.

PR #483, opened 2021-06-03
Remove cloud provider from config

simonklb:

This removes the requirement for cloud provider in the configuration files making ck8s-apps semi-cloud provider agnostic. Not completely agnostic in the sense that it asks for cloud provider to initialize the configuration with sane defaults.

**Checklist:**

    - [x] will change any APIs.
    - [ ] will break the cluster.


cristiklein (approved):
I am a bit split with this change. On one hand, I like how apps gets truly cloud-agnostic. On the other hand, there is at least one decision that depends on the cloud provider (see [ADR-0008 "Use HostNetwork or LoadBalancer for Ingress"](https://compliantkubernetes.io/adr/0008-use-hostnetwork-or-loadbalancer-for-ingress/)).

I think we should apply this change now, but be mentally ready to bring in a bit of cloud-specificity back later. :smile:

config/secrets/sc-secrets.yaml
```diff
@@ -10,6 +10,10 @@ grafana:
   clientSecret: somelongsecret
   opsClientSecret: somelongsecret
 harbor:
+  # persistence:
+  #   swift:
+  #     username: "set-me"
+  #     password: "set-me"
```
Xartos, line 16: Sops doesn't like comments directly under an object key, see https://github.com/mozilla/sops/issues/384 I think we need to use sops 3.7 for this to work. (Maybe this is just a matter of telling people to use it and update the docs)

Please move it

Xartos (changes requested):
Please add a comment in the sc-config template (`config/config/sc-config.yaml`) to be clear on how to set up swift if that's used.

simonklb:
> I am a bit split with this change. On one hand, I like how apps gets truly cloud-agnostic. On the other hand, there is at least one decision that depends on the cloud provider (see [ADR-0008 "Use HostNetwork or LoadBalancer for Ingress"](https://compliantkubernetes.io/adr/0008-use-hostnetwork-or-loadbalancer-for-ingress/)).
> 
> I think we should apply this change now, but be mentally ready to bring in a bit of cloud-specificity back later. smile

As I mentioned in the description, there is a lot of decisions taking place at the config init stage that depends on cloud provider (one prominent example is [storage class](https://github.com/elastisys/compliantkubernetes-apps/blob/0e39cc3457e5372a6098695e682d64da77b6dfdb/bin/init.bash#L115)) but this is only to set sane defaults. You could completely strip any mentions of provider from this repo if you forced users to manually configure everything correctly before running apply. That is not very nice though. :smile:

simonklb:
> Please add a comment in the sc-config template (`config/config/sc-config.yaml`) to be clear on how to set up swift if that's used.

I did this initially, but found it kind of spammy since you will get it initialized if you init with citycloud as the cloud provider anyway (compared to s3 which is not only used for a single corner case). I can add it if you really think it's necessary though!

config/secrets/sc-secrets.yaml
simonklb: My bad! I ran with 3.7 locally. :sweat_smile:  Added the comment at the bottom.

cristiklein:
> > I am a bit split with this change. On one hand, I like how apps gets truly cloud-agnostic. On the other hand, there is at least one decision that depends on the cloud provider (see [ADR-0008 "Use HostNetwork or LoadBalancer for Ingress"](https://compliantkubernetes.io/adr/0008-use-hostnetwork-or-loadbalancer-for-ingress/)).
> > I think we should apply this change now, but be mentally ready to bring in a bit of cloud-specificity back later. smile
> 
> As I mentioned in the description, there is a lot of decisions taking place at the config init stage that depends on cloud provider (one prominent example is [storage class](https://github.com/elastisys/compliantkubernetes-apps/blob/0e39cc3457e5372a6098695e682d64da77b6dfdb/bin/init.bash#L115)) but this is only to set sane defaults. You could completely strip any mentions of provider from this repo if you forced users to manually configure everything correctly before running apply. That is not very nice though.

@simonklb Thanks for the clarification. It seems like I slightly misunderstood your PR. You eliminated cloud-specific stuff from the apply stage, but the init stage still needs `CK8S_CLOUD_PROVIDER` for sane defaults. Correct?

I will think about the init / re-init issue and will bring up a proposal during (one of the) next arch meeting. :smile:

simonklb:
> @simonklb Thanks for the clarification. It seems like I slightly misunderstood your PR. You eliminated cloud-specific stuff from the apply stage, but the init stage still needs `CK8S_CLOUD_PROVIDER` for sane defaults. Correct?

No, I don't think cloud provider has been required in the apply stage for a long time, if ever. This is purely a configuration cleanup. The changes are stated in the changelog.

> I will think about the init / re-init issue and will bring up a proposal during (one of the) next arch meeting. smile

Feel free to ping me if you want someone to bounce things off, I'm currently thinking about the same issue.

Xartos:
> > Please add a comment in the sc-config template (`config/config/sc-config.yaml`) to be clear on how to set up swift if that's used.
> 
> I did this initially, but found it kind of spammy since you will get it initialized if you init with citycloud as the cloud provider anyway (compared to s3 which is not only used for a single corner case). I can add it if you really think it's necessary though!

I would really like this, since it would be very hard to remember every keys needed if I was to set up config for a cluster using this (which is not uncommon).

simonklb:
> > > Please add a comment in the sc-config template (`config/config/sc-config.yaml`) to be clear on how to set up swift if that's used.
> > 
> > 
> > I did this initially, but found it kind of spammy since you will get it initialized if you init with citycloud as the cloud provider anyway (compared to s3 which is not only used for a single corner case). I can add it if you really think it's necessary though!
> 
> I would really like this, since it would be very hard to remember every keys needed if I was to set up config for a cluster using this (which is not uncommon).

I might be missing something, would you not init with citycloud as the cloud provider in that case? Because if you do you should have those keys pre-set. Or are you saying that it's not uncommon to build a config from scratch without using the init command?

viktor-f:
The concern I have with not putting the swift config in `config/config/sc-config.yaml` is that it is a bit hidden. E.g. whenever I want to check our default values I will go to that file. But I do understand your concern that it might be unnecessary clutter. I would at least want this line to be updated https://github.com/elastisys/compliantkubernetes-apps/blob/a9ff3f3d53773191549c25fd79e60f4ecdcebd3a/config/config/sc-config.yaml#L156

simonklb:
> The concern I have with not putting the swift config in `config/config/sc-config.yaml` is that it is a bit hidden. E.g. whenever I want to check our default values I will go to that file. But I do understand your concern that it might be unnecessary clutter.

I see your point. I think we probably should investigate how common it is for us to go to the code repo to find config defaults instead of relying on the init command. If the code repo config files are basically the "go to" for looking up configuration options I think it's high time to create a proper config instructions doc. Perhaps even a yaml schema?

viktor-f:
But I'm totally ok with removing `global.cloudProvider` :slightly_smiling_face:

Xartos:
> > > > Please add a comment in the sc-config template (`config/config/sc-config.yaml`) to be clear on how to set up swift if that's used.
> > > 
> > > 
> > > I did this initially, but found it kind of spammy since you will get it initialized if you init with citycloud as the cloud provider anyway (compared to s3 which is not only used for a single corner case). I can add it if you really think it's necessary though!
> > 
> > 
> > I would really like this, since it would be very hard to remember every keys needed if I was to set up config for a cluster using this (which is not uncommon).
> 
> I might be missing something, would you not init with citycloud as the cloud provider in that case? Because if you do you should have those keys pre-set. Or are you saying that it's not uncommon to build a config from scratch without using the init command?

You're correct, I just saw the `yq write` commands. My bad :slightly_smiling_face:

Xartos (approved):

PR #487, opened 2021-06-07
pipeline: add back missing files

viktor-f:
**What this PR does / why we need it**: We removed some files from the pipeline folder that we didn't think was used. But seems like we didn't look properly. This should return the files that we need.
Here is a pipeline that failed because of the missing files: https://github.com/elastisys/compliantkubernetes-apps/actions/runs/911463783
Also see this line: https://github.com/elastisys/compliantkubernetes-apps/blob/417f3c148c6510222d0b2031ae869cc35e904045/.github/workflows/end-to-end.yml#L158

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: not planned

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/pipeline/apply-sc.bash
+++ b/pipeline/apply-sc.bash
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -u -o pipefail
+
+here="$(dirname "$(readlink -f "$0")")"
+pipeline_dir="${here}"
+ck8s="${here}/../bin/ck8s"
+bin_path="${here}/../bin"
+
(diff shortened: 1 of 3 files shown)


OlleLarsson (approved):

simonklb (approved):
My bad! Shouldn't always trust grep. :sweat_smile:

PR #488, opened 2021-06-08
Init command tests

simonklb:

This adds a few tests for the init command. It uses Bats which was recently added to our tech radar in [ADR-0014](https://compliantkubernetes.io/adr/0014-use-bats-for-testing-bash-wrappers/).

It also does a bit of cleaning and adds some minor fixes.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -40,7 +40,7 @@ jobs:
     - name: Run unit tests for OPA policies
       uses: ./pipeline
       with:
-        args: ./pipeline/test/services/unit-tests/test-opa.sh
+        args: ./pipeline/test-opa.bash
     - name: Run linters
       uses: ./pipeline
       with:
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -51,3 +51,15 @@ jobs:
       uses: ./pipeline
       with:
         args: ./pipeline/test-opa.bash
+
+  bats:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout compliantkubernetes-apps
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Run Bats tests
+        uses: ./pipeline
+        with:
+          args: bash -c "./pipeline/sops-pgp-setup.bash && bats ./pipeline"
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -37,11 +37,17 @@ jobs:
         CK8S_CLOUD_PROVIDER: ${{ matrix.cloud }}
         CI_CK8S_FLAVOR: ${{ matrix.flavor }}
         CK8S_PGP_FP: 529D964DE0BBD900C4A395DA09986C297F8B7757
-    - name: Run unit tests for OPA policies
-      uses: ./pipeline
-      with:
-        args: ./pipeline/test-opa.bash
     - name: Run linters
       uses: ./pipeline
       with:
         args: ./pipeline/lint.bash
+
+  opa:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout compliantkubernetes-apps
+      uses: actions/checkout@v2
+    - name: Run unit tests for OPA policies
+      uses: ./pipeline
+      with:
+        args: ./pipeline/test-opa.bash
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -37,11 +37,12 @@ generate_sops_config() {
             exit 1
         fi
         fingerprint="${CK8S_PGP_FP}"
-    elif [ "${CK8S_PGP_UID+x}" != "" ]; then
+    elif [ -n "${CK8S_PGP_UID:-}" ]; then
         fingerprint=$(gpg --list-keys --with-colons "${CK8S_PGP_UID}" | \
-                      awk -F: '$1 == "fpr" {print $10;}' | head -n 1)
+                      awk -F: '$1 == "fpr" {print $10;}' | head -n 1 || \
+                      echo "")
         if [ -z "${fingerprint}" ]; then
-            log_error "ERROR: Unable to get fingerprint from gpg keyring."
+            log_error "ERROR: Unable to get fingerprint from gpg keyring using UID."
             log_error "CK8S_PGP_UID=${CK8S_PGP_UID}"
             exit 1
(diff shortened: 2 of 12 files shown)


bin/init.bash
```diff
@@ -31,13 +31,19 @@ if [ "${CK8S_FLAVOR}" != "dev" ] &&
 fi
 
 generate_sops_config() {
-    if [ "${CK8S_PGP_FP+x}" != "" ]; then
+    if [ -n "${CK8S_PGP_FP:-}" ]; then
+        if ! gpg --list-keys | grep "${CK8S_PGP_FP}" > /dev/null 2>&1; then
```
Xartos, line 35: Nice!

pipeline/init.bats
```diff
@@ -0,0 +1,100 @@
...
+    export PATH="${BATS_TEST_DIRNAME}/../bin:${PATH}"
+
+    export CK8S_CONFIG_PATH="${BATS_TMPDIR}/ck8s-apps-config"
+    export CK8S_ENVIRONMENT_NAME=test
+    export CK8S_PGP_FP=529D964DE0BBD900C4A395DA09986C297F8B7757
+    export CK8S_CLOUD_PROVIDER=aws
+}
+
+teardown() {
+    rm -rf "${CK8S_CONFIG_PATH}"
```
Xartos, line 18: Maybe I'm a bit paranoid here, but this feels very dangerous if you've set the variable wrong and run this function.
I'd rather have `${BATS_TMPDIR}/ck8s-apps-config` here as well even though it adds duplicates.

pipeline/init.bats
```diff
@@ -0,0 +1,100 @@
...
+    CK8S_CLOUD_PROVIDER=aws CK8S_FLAVOR=foo run ck8s init
+    assert_failure
+    assert_output --partial "Unsupported flavor: foo"
+}
+
+@test "ck8s init is idempotent" {
+    # TODO: Convert this to parametric test if it gets implemented:
+    # https://github.com/bats-core/bats-core/issues/241
+    [ -n "${CK8S_SKIP_LONG_RUNNING_TESTS:-}" ] && skip
+    for cloud_provider in aws citycloud exoscale safespring; do
```
Xartos, line 71: How should we make sure that we remembers to update these as well when adding new cloud providers? Maybe we can add a checkbox on PR templates for now just to have some reminder?

Xartos (commented):
Great stuff! I just have some comments

pipeline/init.bats
viktor-f: I agree

simonklb: Good point. :smile:

In any case, it's advisable to run the whole Bats suite inside a container because it does not give guarantees about side-effects.

pipeline/init.bats
simonklb: Yea I've thought about this as well. At first I added it to the GitHub actions matrix, that made things more spread out and I kind of like that the tests can be run as-is instead of having to be run multiple times to cover all cases.

What I did in devbox was putting all the providers in a common file so that you only had to set them once. That is at least better. I can add this.

simonklb: What about adding commands that lists the available providers and flavors?

```
$ ck8s providers
aws citycloud exoscale safespring
$ ck8s flavors
dev prod
```

Then we have a single source of truth.

pipeline/init.bats
simonklb: Fixed, PTAL.

pipeline/init.bats
Xartos: Yea, I like that idea

simonklb: Check https://github.com/elastisys/compliantkubernetes-apps/pull/493

simonklb: This is based on the above PR, hopefully this should make it less error-prone when new providers are added. :)

Xartos (approved):
LGTM

.github/workflows/pipeline.yml
```diff
@@ -37,11 +37,28 @@ jobs:
         CK8S_CLOUD_PROVIDER: ${{ matrix.cloud }}
         CI_CK8S_FLAVOR: ${{ matrix.flavor }}
         CK8S_PGP_FP: 529D964DE0BBD900C4A395DA09986C297F8B7757
-    - name: Run unit tests for OPA policies
-      uses: ./pipeline
-      with:
-        args: ./pipeline/test/services/unit-tests/test-opa.sh
```
viktor-f, line 43: I sort of like the idea of actually having a script for this. The main reason is that later on if we add more unit tests (like the bat tests) we can gather them in a bigger script that you can run locally as well. I know that you can still run this command locally, but I guess it's a bit more hidden.
Since OPA has not been pushed lately it might be fine to have it like this for now and move it later again if we would like  that.

viktor-f (approved):
Looks good overall. Just some thoughts

bin/init.bash
```diff
@@ -30,13 +30,19 @@ if ! array_contains "${CK8S_FLAVOR}" "${ck8s_flavors[@]}"; then
 fi
 
 generate_sops_config() {
-    if [ "${CK8S_PGP_FP+x}" != "" ]; then
+    if [ -n "${CK8S_PGP_FP:-}" ]; then
```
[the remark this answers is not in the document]
simonklb: They are equal, I believe it was me who did the previous version as well. I've just always been bothered by the `+x`. It looks confusing to me. Most people know what `-n` means and if not it's easy to look up. The same with `${x:-y}`.

.github/workflows/pipeline.yml
simonklb: Yea I don't mind. But I think we should move it from `./pipeline/test/services/unit-tests/test-opa.sh` to just `./pipeline/opa.sh` or something like that in that case, because the previous one is also a bit hidden.

viktor-f: That is ok with me.
Alternatively put both the opa test and bat tests in some unit test folder and create a `.pipeline/unit-test.sh` that runs both. Then mention it in the main README and/or the public docs.

simonklb: We currently don't inform people on how to setup and run Bats locally other than through our pipeline Docker container. Let's cleanup the local testing workflow later! I've restored the test-opa script though.

PR #491, opened 2021-06-09
Enable additional static clients for dex

### LINKED ISSUE #456: [1] Enable additional static clients for dex

Since users might want to enable their own applications to be able to use dex, it would be nice to add a `additionalStaticClients` for dex

Configuration for dex where you can add additional static clients.
It should add it to this list https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/dex.yaml.gotmpl#L69

And since there's some secret information included in each entry this should really reside in `secrets.yaml`

Not have this option. This would not enable users to utilize the SSO of ck8s in their applications.

**Additional context**

**Definition of done:**

I'm able to add arbitrary amount of static clients to dex (with any config)

mrombout:

In this PR I've added `dex.additionalStaticClients` to be able to specify Dex clients for application that do not belong to CK8s. More details in #456.

This PR also contains a migration script that will pull any custom static clients already defined for Dex and move it into the `secrets.yaml` at the appropriate place.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #456

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

I've added `dex.additionalStaticClients` to the secrets of the pipline just because it seemed like a good thing to do.
But the pipeline didn't actually fail when it was missing.

I actually expected it to fail because the Dex template would `range` a value that doesn't exist.

### DIFF
--- a/helmfile/values/dex.yaml.gotmpl
+++ b/helmfile/values/dex.yaml.gotmpl
@@ -75,6 +75,9 @@ config:
         - https://kibana.{{ .Values.global.baseDomain }}/auth/openid/login
         - https://kibana.{{ .Values.global.baseDomain }}/app/kibana
 {{ end }}
+{{ if .Values.dex.additionalStaticClients }}
+  {{- toYaml .Values.dex.additionalStaticClients | nindent 4 }}
+{{ end }}
 
 {{ if .Values.dex.enableStaticLogin }}
   enablePasswordDB: true
--- a/migration/v0.16.x-v0.17.x/migrate-dex-additional-static-clients.sh
+++ b/migration/v0.16.x-v0.17.x/migrate-dex-additional-static-clients.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+set -euo pipefail
+
+: "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
+
+here="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
+
+secret="${CK8S_CONFIG_PATH}/secrets.yaml"
+secret_tmp="${CK8S_CONFIG_PATH}/secrets_tmp.yaml"
+sops_conf="${CK8S_CONFIG_PATH}/.sops.yaml"
+
+update_script=$(mktemp -p /tmp update-script.XXXXXXXXXX.yaml)
+cp "${here}/dex-update-script.yaml" "${update_script}"
+
+# read all dex's static clients from ck8s cluster
+static_clients_yaml=$(./bin/ck8s ops kubectl sc get secret dex --namespace dex --output jsonpath='{.data.config\\.yaml}' | base64 --decode | yq read - staticClients)
+
+# delete clients that are part of a standard v0.16.x installation so only custom clients remain
+static_clients_yaml=$(echo "$static_clients_yaml" | \
+    yq delete - "id==kubelogin" | \
+    yq delete - "id==grafana" | \
+    yq delete - "id==grafana-ops" | \
+    yq delete - "id==harbor" | \
+    yq delete - "id==kibana-sso" | \
+    yq delete - "id==grafana")
+
+# merge custom clients from ck8s cluster into secrets config under additionalStaticClients
+echo "$static_clients_yaml" | yq prefix - "[0].value" | yq merge --inplace "${update_script}" -
+sops -d "${secret}" | \
+    yq write --script "${update_script}" - | \
+    sops --config "${sops_conf}" --input-type=yaml --output-type=yaml -e /dev/stdin > "${secret_tmp}"
+
+cp "${secret}" "${secret}.bak"
+mv "${secret_tmp}" "${secret}"
+
+rm "${update_script}"
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -16,6 +16,7 @@
 ### Added
 
 - Option to set cluster admin groups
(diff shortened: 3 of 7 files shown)


helmfile/values/dex.yaml.gotmpl
```diff
@@ -75,6 +75,15 @@ config:
...
 {{ end }}
+{{ range $key, $value := .Values.dex.additionalStaticClients }}
+    - id: {{ $value.id }}
+      secret: {{ $value.secret }}
+      name: {{ $value.name }}
+      redirectURIs:
+      {{ range $key, $value := $value.redirectURIs }}
+        - {{ $value }}
+      {{ end }}
+{{ end }}
```
Xartos, line 86: Wouldn't it work with just 
```suggestion
{{- toYaml .Values.dex.additionalStaticClients | nindent 4 }}
```

Xartos (commented):
Looks good from first glance, will try to test it as well. I just have a comment for now

migration/v0.16.x-v0.17.x/migrate-dex-additional-static-clients.sh
```diff
@@ -0,0 +1,37 @@
...
+
+secret="${CK8S_CONFIG_PATH}/secrets.yaml"
+secret_tmp="${CK8S_CONFIG_PATH}/secrets_tmp.yaml"
+sops_conf="${CK8S_CONFIG_PATH}/.sops.yaml"
+
+update_script=$(mktemp -p /tmp update-script.XXXXXXXXXX.yaml)
+cp "${here}/dex-update-script.yaml" "${update_script}"
+
+# read all dex's static clients from ck8s cluster
+static_clients_yaml=$(./bin/ck8s ops kubectl sc get secret dex --namespace dex --output json | jq --raw-output '.data."config.yaml"' | base64 --decode | yq read - staticClients)
```
Xartos, line 17: I don't think it matters, I mainly wanted to highlight that you can use `jsonpath` instead of piping to jq. So you can change it if you want or just ignore this

```suggestion
static_clients_yaml=$(./bin/ck8s ops kubectl sc get secret dex --namespace dex --output jsonpath='{.data.config\.yaml}' | base64 --decode | yq read - staticClients)
```

Xartos:
Did some light testing and everything seems to work well :slightly_smiling_face:

helmfile/values/dex.yaml.gotmpl
Xartos: Just tested to change to this and it seemed to work fine. So I would like this solution instead, makes it easier if dex decides to change the structure of this data

mrombout: Nice, that seems to work, but only when `additionalStaticClients != []`. Is there a smarter way for this rather than an `if len` as well?

Xartos: > Nice, that seems to work, but only when additionalStaticClients != []. I'll put an if around it.

Ofcorse, nice catch!

> This does open things up to surprises a little bit when Dex updates and decides to change their structure in there. It might not be a problem 99% of the time, but still. In other words, Dex's abstraction is leaking into our own this way.

Sure, but that's kind of what we want. You should be able to read the documentation on dex and set it up the same way here. If we go this road and add our own layer of abstraction we might confuse people more if dex changes structure and we're not.

migration/v0.16.x-v0.17.x/migrate-dex-additional-static-clients.sh
mrombout: Cool, but somehow it's not printing any output. It's probably the `config.yaml`, because it prints fine for `{.data}`. I'll see if I can get that working.

migration/v0.16.x-v0.17.x/migrate-dex-additional-static-clients.sh
```diff
@@ -0,0 +1,37 @@
...
+    yq delete - "id==kibana-sso" | \
+    yq delete - "id==grafana")
+
+# merge custom clients from ck8s cluster into secrets config under additionalStaticClients
+echo "$static_clients_yaml" | yq prefix - "[0].value" | yq merge --inplace "${update_script}" -
+sops -d "${secret}" | \
+    yq write --script "${update_script}" - | \
+    sops --config "${sops_conf}" --input-type=yaml --output-type=yaml -e /dev/stdin > "${secret_tmp}"
+
+cp "${secret}" "${secret}.bak"
```
mrombout, line 34: Just added this as a precaution, if this is something we want to keep in the script I can add a instruction to the migration instructions to delete it after verification?

helmfile/values/dex.yaml.gotmpl
mrombout: Alright, I can see that point of view as well :+1: the `toYaml` with an `if` around seems to work fine for both empty and non-empty arrays now.

migration/v0.16.x-v0.17.x/migrate-dex-additional-static-clients.sh
Xartos: I'm not sure, ideally everyone should have their config path version controlled so then this is not needed. But it's always nice to be able to have a backup.

Maybe add a note in the migration guide after the script step with something like, `verify that the script added all your custom static clients and delete the backup`

Xartos (approved):
Looking good now!

OlleLarsson (approved):
🎉

viktor-f (approved):
LGTM

PR #493, opened 2021-06-09
Add commands that list providers and flavors

simonklb:

The providers and flavors this project supports are now declared in `bin/commons.bash` and should be the single source of truth. This can also be retrieved from two new commands:
* ck8s providers
* ck8s flavors

This commit also includes some cleanup of how shellcheck recognizes sourced files. See https://github.com/elastisys/compliantkubernetes-kubespray/pull/87 for the same fix in ck8s-kubespray.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


bin/common.bash
```diff
@@ -102,11 +118,7 @@ validate_version() {
 
 # Check if the cloud provider is supported.
 validate_cloud() {
-    if [ "${1}" != "exoscale" ] &&
-       [ "${1}" != "safespring" ] &&
-       [ "${1}" != "citycloud" ] &&
-       [ "${1}" != "baremetal" ] &&
-       [ "${1}" != "aws" ]; then
+    if ! array_contains "${1}" "${ck8s_cloud_providers[@]}"; then
         log_error "ERROR: Unsupported cloud provider: ${1}"
```
Xartos, line 122: Should we add the list of supported providers here?
```suggestion
        log_error "ERROR: Unsupported cloud provider: ${1}"
        log_error "Supported providers: ${ck8s_cloud_providers[@]}"
```

bootstrap/bootstrap.sh
```diff
@@ -4,10 +4,9 @@
 
 set -euo pipefail
 
-: "${bootstrap_path:?Missing bootstrap path}"
+here="$(dirname "$(readlink -f "$0")")"
+
 environment="${1}"
-export storageclass_path="${bootstrap_path}/storageclass"
-export namespaces_path="${bootstrap_path}/namespaces"
```
Xartos, line 10: Isn't this used in `bootstrap/namespaces/bootstrap.sh`?

Xartos (changes requested):
This commands feels so much leaner :tada: I have some minor issues though

bootstrap/bootstrap.sh
simonklb: damn, nice catch! i've tried to make them a bit more self-contained but missed that one - fixed now

bin/common.bash
simonklb: What do you think about `log_error: "Use the command **ck8s providers** to list supported providers"? It's fine with me either way.

bootstrap/bootstrap.sh
Xartos: Nice! yea, It's a bit odd that the parent script should provide the path.

bin/common.bash
Xartos: Yea, that good as well :slightly_smiling_face:

simonklb: Nevermind, I think proposing commands here could become a problem because that kind of assumes that this is the top-level entrypoint. In the devbox `ck8s` is renamed to `ck8s-apps` and that message would only be confusing in that case. I'll just list the supported providers like you proposed! :slightly_smiling_face:

simonklb: Pushed!

bootstrap/bootstrap.sh
simonklb: Fix pushed!

simonklb:
Actually, I think moving the arrays from common to init makes sense. That will make it more obvious that they are not something global but only used for initializing the configuration.

simonklb:
> Actually, I think moving the arrays from common to init makes sense. That will make it more obvious that they are not something global but only used for initializing the configuration.

https://github.com/elastisys/compliantkubernetes-apps/pull/493/commits/0c77009a0b8bcb209e341fa0e617b7bad84ab469

simonklb:
> > Actually, I think moving the arrays from common to init makes sense. That will make it more obvious that they are not something global but only used for initializing the configuration.
> 
> [0c77009](https://github.com/elastisys/compliantkubernetes-apps/commit/0c77009a0b8bcb209e341fa0e617b7bad84ab469)

Nevermind, they obviously are needed for the commands. :rofl:

bin/ck8s
```diff
@@ -5,7 +5,7 @@
 set -e -o pipefail
 
 here="$(dirname "$(readlink -f "$0")")"
-# shellcheck disable=SC1090
+# shellcheck source=bin/common.bash
 source "${here}/common.bash"
 
 usage() {
```
Xartos, line 11: Please add the new commands to the list here as well

Xartos (changes requested):

bin/ck8s
simonklb: Added completions as well! ;)

Xartos (approved):
LGTM!

pipeline/test/services/test-sc.sh
```diff
@@ -14,10 +14,10 @@ DEBUG_OUTPUT=("")
...
 export CLUSTER="ServiceCluster"
 
-source "${SCRIPTS_PATH}"/service-cluster/testPodsReady.sh
-source "${SCRIPTS_PATH}"/common/testPersistentVolumeClaims.sh
-source "${SCRIPTS_PATH}"/service-cluster/testEndpoints.sh
-source "${SCRIPTS_PATH}"/service-cluster/testPrometheusTargets.sh
+"${SCRIPTS_PATH}/service-cluster/testPodsReady.sh"
+"${SCRIPTS_PATH}/common/testPersistentVolumeClaims.sh"
+"${SCRIPTS_PATH}/service-cluster/testEndpoints.sh"
+"${SCRIPTS_PATH}/service-cluster/testPrometheusTargets.sh"
```
viktor-f, line 20: Don't change this, otherwise the count does not work.

pipeline/test/services/test-wc.sh
```diff
@@ -14,12 +14,11 @@ DEBUG_OUTPUT=("")
...
-source "${SCRIPTS_PATH}"/workload-cluster/testPodsReady.sh
-source "${SCRIPTS_PATH}"/common/testPersistentVolumeClaims.sh
-source "${SCRIPTS_PATH}"/workload-cluster/testEndpoints.sh
-source "${SCRIPTS_PATH}"/workload-cluster/testPrometheusTargets.sh
-source "${SCRIPTS_PATH}"/workload-cluster/testUserRbac.sh
+"${SCRIPTS_PATH}/workload-cluster/testPodsReady.sh"
+"${SCRIPTS_PATH}/common/testPersistentVolumeClaims.sh"
+"${SCRIPTS_PATH}/workload-cluster/testEndpoints.sh"
+"${SCRIPTS_PATH}/workload-cluster/testPrometheusTargets.sh"
+"${SCRIPTS_PATH}/workload-cluster/testUserRbac.sh"
```
viktor-f, line 21: Same here

bin/apps.bash
```diff
@@ -24,8 +24,6 @@ apps_run_sc() {
     log_info "Applying applications in service cluster"
 
     (
-        : "${scripts_path:?Missing scripts path}"
-        : "${secrets[kube_config_sc]:?Missing service cluster kubeconfig}"
```
viktor-f, line 28: Why are these and all other similar lines removed? Is it to pass some shellcheck tests?

viktor-f (changes requested):
I like the new commands :slightly_smiling_face: 

While you are doing cleanup. I tried to run the tests and noticed that `pipeline/test/services/common/testPersistentVolumeClaims.sh` is missing the executabe permission. Could you `chmod +x` it for us?

bin/apps.bash
simonklb: They were introduced because shellcheck weren't able to see them in the sourced files so shellcheck complained about the referenced variables not being assigned. By using the `-x/--external-sources` flag with shellcheck and adding the `# shellcheck source=x` directive, shellcheck is now able to see the values defined in sourced files.

pipeline/test/services/test-sc.sh
simonklb: Ah, my bad. I kind of dislike using source for executing things but I see why it's being used here. What do you think of sourcing the files and executing the functions inside the `test-*.sh` scripts instead?

bin/apps.bash
viktor-f: Ah, now I understand

pipeline/test/services/test-sc.sh
viktor-f: Yeah I understand that. I don't really like this either.
Not sure I fully like your suggestion either, but I'm ok with it and I don't think I have any better suggestions. If you go with that, then I suggest that you execute each function directly below the line that sources them. So that it's clear which file a function is coming from.

simonklb: I saw the execution inside the testUserRbac.sh script and it's way too much to clean up right now to be bothered. I think all of the tests under `test/services` really should live in `bin/`  because all of that really is the implementation of the `ck8s test` command and not for the pipeline only.

pipeline/test/services/test-wc.sh
simonklb: Fixed, PTAL!

pipeline/test/services/test-sc.sh
viktor-f: Yeah I agree that we probably should not have our tests inside `pipeline/`. Maybe we can move them to another folder (`tests/`?), but that is for another PR. You can create an issue if you want.

viktor-f (approved):
LGTM!

PR #494, opened 2021-06-09
apps sc: kube-prometheus-stack grafana upgrade

### LINKED ISSUE #466: [1] Many grafana dashboards are broken
**Describe the bug**

A lot of our dashboards are using the variable `__rate_interval` [which was introduced in grafana 7.2](https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/) while we are running grafana 7.0

This breaks a lot of our dashboards with the error `parse error: missing unit character in duration`

This can be fixed by either change all these variables to `__interval` or we can upgrade grafana to a later version (I would refer the latter)

**To Reproduce**
1. Go to Grafana
2. Go to the dashboard `Kubernetes / Compute Resources / Namespace (Workloads)`
3. Scroll down to the network dashboards on the bottom

**Expected behavior**

No error to show

**Screenshots**

screenshot_2021-05-26-162809

 - Compliant kubernetes apps version v0.15.0

**Additional context**

robinelastisys:
A lot of our dashboards are using the variable __rate_interval which was introduced in grafana 7.2 while we are running grafana 7.0. This breaks a lot of our dashboards with the error parse error: missing unit character in duration. Kube-prometheus-stack are now being upgraded from 12.8.0 to 16.6.1 to fix this.

fixes #466 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.
    - [x] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).


robinelastisys:
After asking the team for what they think we managed to create a solution and we did the following:

1. Replace the old kube-prometheus-stack folder with the newest one (https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) 
2. Removing the Prometheus and Grafana specifications in https://github.com/elastisys/compliantkubernetes-apps/blob/74788e09b707313a6f2e9bc2ff2fa3815c385279/helmfile/values/kube-prometheus-stack-sc.yaml.gotmpl
3. ./bin/ck8s ops kubectl sc apply -f '/home/robin/compliantkubernetes-apps/helmfile/upstream/kube-prometheus-stack/crds' to apply the new CRDs before upgrading to the new version.
4. Adding initChownData:enabled: false to bypass a temporary bug in the current version.

WIP-CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 - The `clusterDns` config variable now matches Kubespray defaults.
   Using the wrong value causes node-local-dns to not be used.
+- Grafana in the kube-prometheus-stack are upgraded to fix dashboard errors.
```
Xartos, line 15: You should change this to instead mention that kube-prometheus-stack are upgraded

Xartos (commented):
> Special notes:
> To upgrade from 12.8.0 you need to run:
> ./bin/ck8s ops kubectl sc apply -f '/home/robin/compliantkubernetes-apps/helmfile/upstream/kube-prometheus-stack/crds'

Please add this as a step in the migration guides instead here https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration/v0.16.x-v0.17.x/upgrade-apps.md
and don't use absolute paths, use relative paths instead.

WIP-CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 - The `clusterDns` config variable now matches Kubespray defaults.
   Using the wrong value causes node-local-dns to not be used.
+- Kube-prometheus-stack are now being upgraded from 12.8.0 to 16.6.1 to fix dashboard errors.
```
viktor-f, line 15: It would be nice if this also said which versions of prometheus and grafana that we now use.

viktor-f (changes requested):
You should remove this https://github.com/elastisys/compliantkubernetes-apps/blob/4c4c55754c2d6538ef578eb8ccc0d462b52c6168/helmfile/values/kube-prometheus-stack-wc.yaml.gotmpl#L45
and also test to upgrade in WC.

I just also realized that we do not update the user grafana in this PR. It is not necessary to do just to fix your issue. But we would now run different grafana versions (even different major versions) for ops grafana and user grafana, which I don't like. So please update user grafana as well to match the same chart that our version of kube-prometheus-stack is using (update helmfile, values, and `helmfile/upstream/grafana`).

WIP-CHANGELOG.md
robinelastisys: Hmm! Lets see what i can do! :+1:

robinelastisys: The user-grafana have now been upgraded to version 8.0.1, sharing the same major-version as the ops one! :+1:

helmfile/values/grafana-user.yaml.gotmpl
```diff
@@ -2,7 +2,6 @@ adminPassword: {{ .Values.user.grafanaPassword }}
 
 image:
   repository: grafana/grafana
-
```
viktor-f, line 5: Didn't you need to add this here?
```
  initChownData:
    enabled: false
```

viktor-f (changes requested):

viktor-f (approved):
LGTM!

Xartos (approved):
LGTM

PR #520, opened 2021-06-30
apps sc & wc: release 0.17 fixes

robinelastisys:
QA fixes.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


migration/v0.16.x-v0.17.x/migrate-openid.sh
```diff
@@ -9,6 +9,9 @@ set -euo pipefail
 # Restart the master pods so that they mount the latest version of the secret after apply
 "${here}/../../bin/ck8s" ops kubectl sc delete pod -n elastic-system -l role=master
 
+# Wait for master pod to be ready before updating the config
+sleep 60
```
Xartos, line 13: I don't really like using sleep, would it be better to just add this as snippets in the migration doc? And just write "wait until ready" instead of using sleep?

viktor-f: We could also have something like this:
```suggestion
kubectl wait --for=condition=ready pod  -n elastic-system opendistro-es-master-0
```

robinelastisys: The best thing I can come up with atm is a hybrid between both of them. Just the condition alone is not enough as the ports are having some delay in the first seconds.

pettersv (dismissed):
Fix @Xartos comment and it is :+1:

migration/v0.16.x-v0.17.x/migrate-openid.sh
viktor-f: I'm ok with that

Xartos (approved):
Just rebase on the release-0.17 branch and it's good to go

PR #496, opened 2021-06-15
apps sc: upgrade stable/dex to dex/dex

robinelastisys:
**What this PR does / why we need it**:
The chart stable/dex needs to be replaced as it is deprecated. As dex itself is backwards compatible, no config-file variables need to be changed but the version and app numbers - this is done.

**Which issue this PR fixes**: 
fixes #385 

**Special notes for reviewer**:
Migrating from stable/dex (or banzaicloud-stable/dex) chart

This chart is not backwards compatible with the stable/dex (or banzaicloud-stable/dex) chart.

However, Dex itself remains backwards compatible, so you can easily install the new chart in place of the old one and continue using Dex with a minimal downtime.

**Checklist:**

- [x] Added relevant notes to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md)
- [x] Proper commit message prefix on all commits
- [ ] Updated the [public facing documentation](https://github.com/elastisys/compliantkubernetes)
- Is this changeset backwards compatible for existing clusters? Applying:
    - [x] is completely transparent, will not impact the workload in any way.
    - [ ] requires running a migration script.
    - [ ] will create noticeable cluster degradation.
          E.g. logs or metrics are not being collected or Kubernetes API server
          will not be responding while upgrading.
    - [ ] requires draining and/or replacing nodes.
    - [ ] will change any APIs.
          E.g. removes or changes any CK8S config options or Kubernetes APIs.
    - [ ] will break the cluster.
          I.e. full cluster migration is required.
- Chart checklist (pick exactly one):
    - [ ] I upgraded no Chart.
    - [x] I upgraded a Chart and determined that no migration steps are needed.
    - [ ] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).

PR #550, opened 2021-08-04
apps sc: apply dex kibana fix

### LINKED ISSUE #523: Get dex to work in Kibana in release-0.17 on a new cluster automatically
**Describe the bug**
As of now, dex work halfway in Kibana on a new cluster. You can fix it easily by running migrate-openid.sh but before that, Kibana will say "Unauthorized" when you try to enter it. Feels like a "kubectl apply" race condition error.

**To Reproduce**
1. Setup a new exoscale cluster with "git checkout release 0.17" and configure dex
2. Try enter your Kibana page
3. You will get a page saying "Unauthorized"
4. Fix it by running /migration/v0.16.x-v0.17.x/migrate-openid.sh

**Expected behavior**
That you can log in to Kibana using your google account by dex without having to run /migration/v0.16.x-v0.17.x/migrate-openid.sh on a fresh cluster.

robinelastisys:
As of now, authentication by dex in Kibana doesn't work without restarting Kibana on a completely fresh cluster. This makes opendistro-es be installed when dex is ready.

fixes #523 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - Fixed influxdb-du-monitor to only select influxdb and not backup pods
+- Added dex/dex as a need for opendistro-es to make kibana available out-the-box at cluster initiation if dex is enabled
 
 ### Added
 
--- a/helmfile/50-applications.yaml
+++ b/helmfile/50-applications.yaml
@@ -383,6 +383,8 @@ releases:
   chart: ./elastisys/opendistro-es
   version: 1.12.1
   missingFileHandler: Error
+  needs:
+  - dex/dex
   values:
   - values/opendistro-es.yaml.gotmpl
 


OlleLarsson:
Instead of doing something like this, is not better to figure out why:

> As of now, authentication by dex in Kibana doesn't work without restarting Kibana on a completely fresh cluster

Have we changed something? Has dex changed something? etc.

robinelastisys:
> Before doing something like this, do you know why:
> 
> > As of now, authentication by dex in Kibana doesn't work without restarting Kibana on a completely fresh cluster
> 
> Have we changed something? Has dex changed something? etc.

I think so. Before the summer, we did https://github.com/elastisys/compliantkubernetes-apps/pull/496, right? We didn't catch on to it doe as we didn't test it enough until the making of the release.

OlleLarsson:
Sure I mean i get that we have changed the dex helm chart but that still does not fully answer my question.
Why does the openid init fail in Kibana now? Have you looked at the logs? Is dex not ready by the time kibana starts? etc.

robinelastisys:
> Sure I mean i get that we have changed the dex helm chart but that still does not fully answer my question.
> Why does the openid init fail in Kibana now? Have you looked at the logs? Is dex not ready by the time kibana starts? etc.

I'm not all that sure but perhaps we can figure it out together? One thing we know at least is that, to make the migration from v0.16.x to v0.17.x, migration/v0.16.x-v0.17.x/migrate-openid.sh need to run which is more or less the same issue as this I would think? So, it must be a race condition as u say?

WIP-CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 
 ### Fixed
 
+- add dex/dex as a need for opendistro-es
```
OlleLarsson, line 12: A more descriptive entry would be nice :)

viktor-f (commented):
If this fix works, then this looks good.
But I would also be interested in a more thorough explanation as to why, if you have any.

viktor-f (approved):

Xartos (approved):

PR #552, opened 2021-08-09
apps: Remove unnecessary PSPs and RBAC

### LINKED ISSUE #304: Remove unnecessary PSPs and RBAC
**Describe the bug**
This may not really count as a bug, more like cleanup of technical debt or basic maintenance.
The PSPs that we [install using this helmfile](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/00-psp_rbac.yaml) are not needed any more. I discovered this as part of https://github.com/elastisys/compliantkubernetes-apps/issues/293 since the charts had conflicts when trying to run in the same cluster. Most of the apps worked just as well without the PSP charts but in some cases (fluentd) I had to set a value like `podSecurityPolicy.enabled=true` to get it to work.

Installing PSPs and RBAC that we don't need just makes the clusters less secure. Besides, less code to maintain is great. :smile: 

**To Reproduce**
1. Deploy [simple-ck8s](https://github.com/elastisys/simple-ck8s
2. Notice that everything still works even though it doesn't use the PSP charts.

 - Compliant kubernetes apps version v0.10.0

robinAwallace:

Remove unnecessary PSPs and RBAC to make the cluster more secure and less to maintain.

fixes #304 

This is what I found which could be removed.

* Only tested om exoscale

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


crssnd (approved):

OlleLarsson:
I haven't looked through the entire PR but there needs to some migration or documentation on how to remove the releases that was deleted by the PR.

robinAwallace:
> I haven't looked through the entire PR but there needs to some migration or documentation on how to remove the releases that was deleted by the PR.

Okay, how should I document this?

robinAwallace:
> should we have the same for the wc?

If you are thinking about the migration script for sc the PSPs for the wc will be removed by the next ck8s apply.

Xartos (commented):
Maybe I'm missing something, but couldn't we just set `installed: false` this release and then delete the charts next release?
That would make it so we wouldn't need the migration script and potentially miss stuff.

I'm not super against this approach either, but think that's how we have done it previously and I kind of like that approach.

OlleLarsson:
> Maybe I'm missing something, but couldn't we just set `installed: false` this release and then delete the charts next release?
> That would make it so we wouldn't need the migration script and potentially miss stuff.
> 
> I'm not super against this approach either, but think that's how we have done it previously and I kind of like that approach.

I guess the "migration" in this case is just removing the un-needed stuff from the service-cluster-psp-rbac chart.

Xartos:
> I guess the "migration" in this case is just removing the un-needed stuff from the service-cluster-psp-rbac chart (should really be renamed as it now only contains role/binding for influxdb :D).

Right, doesn't they get deleted if they are removed from the chart though? So are the delete script really needed? 

> The `installed: false` approach is used for the common-psp-rbac chart :)

Yes, but the chart files are also deleted. Now that I think of it, doesn't that cause an error? trying to uninstall a chart that doesn't exists? I mean it exists in the cluster state, but the files themselves doesn't.

OlleLarsson:
> > I guess the "migration" in this case is just removing the un-needed stuff from the service-cluster-psp-rbac chart (should really be renamed as it now only contains role/binding for influxdb :D).
> 
> Right, doesn't they get deleted if they are removed from the chart though? So are the delete script really needed?
> 
> > The `installed: false` approach is used for the common-psp-rbac chart :)
> 
> Yes, but the chart files are also deleted. Now that I think of it, doesn't that cause an error? trying to uninstall a chart that doesn't exists? I mean it exists in the cluster state, but the files themselves doesn't.

It should work, I've done so previously.
I mean it's not running helm diff or anything like that when running `helmfile destroy`, and state is in kubernetes, but I think that @robinAwallace can confirm whether or not it works as intended :)

OlleLarsson:
> > I guess the "migration" in this case is just removing the un-needed stuff from the service-cluster-psp-rbac chart (should really be renamed as it now only contains role/binding for influxdb :D).
> 
> Right, doesn't they get deleted if they are removed from the chart though? So are the delete script really needed?

Your probably right 🤔

robinAwallace:
> > > The `installed: false` approach is used for the common-psp-rbac chart :)
> > 
> > 
> > Yes, but the chart files are also deleted. Now that I think of it, doesn't that cause an error? trying to uninstall a chart that doesn't exists? I mean it exists in the cluster state, but the files themselves doesn't.
> 
> It should work, I've done so previously.
> I mean it's not running helm diff or anything like that when running `helmfile destroy`, and state is in kubernetes, but I think that @robinAwallace can confirm whether or not it works as intended :)

I tried this again becuase it was some time ago I did it. 
There was no problem with uninstalling charts that have been deleted. They were removed as intended.

>> guess the "migration" in this case is just removing the un-needed stuff from the service-cluster-psp-rbac chart (should really be renamed as it now only contains role/binding for influxdb :D).
>
> Right, doesn't they get deleted if they are removed from the chart though? So are the delete script really needed?

You were right. The charts that were deleted was removed. So there is no need for the delete script. I will remove it.

Should i also remove the migration steps/docs?

OlleLarsson (approved):
LGTM.
Create an issue about removing the entries that you put `installed: false` on after 0.18 has been released,

Xartos (approved):
LGTM

PR #562, opened 2021-08-17
apps sc: prometheus-elasticsearch-exporter timeout fix

### LINKED ISSUE #511: Timeout for elasticsearch exporter
**Describe the bug**

When changing the scrape timeout for elasticsearch prometheus-exporter, the timeout is only changed in the service monitor and not in the exporter itself.

**To Reproduce**
1. Change `elasticsearch.exporter.serviceMonitor.scrapeTimeout` to some high value (maybe because of slow S3)
2. Check the timeout in the elasticsearch prometheus-exporter pod. It's still 30s

**Expected behavior**

The timeout in the exporter should be changed to a value slightly below the scrapeTimeout (maybe 1 or 5 seconds before the scrapeTimeout)

**Suggested solution**

The bug is here because we use a hardcoded value [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/prometheus-elasticsearch-exporter.yaml.gotmpl#L11). Instead it should use the value from `elasticsearch.exporter.serviceMonitor.scrapeTimeout`, just a bit lower

**Screenshots**

 - Compliant kubernetes apps version v0.16.0

**Additional context**

robinelastisys:
When changing the scrape timeout for prometheus-elasticsearch-exporter, the timeout is currently only changed in the service monitor and not in the exporter itself. These timeouts need to be synched. This is fixed by setting the timeout of the exporter to be 5s lower than the service monitor programmatically.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #511 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - ingress-nginx increased the value for client-body-buffer-size from 16K to 256k
 - Lowered default falco resource requests
+- The timeout of the prometheus-elasticsearch-exporter is set to be 5s lower than the one of the service monitor
 
 ### Fixed
 
--- a/helmfile/values/prometheus-elasticsearch-exporter.yaml.gotmpl
+++ b/helmfile/values/prometheus-elasticsearch-exporter.yaml.gotmpl
@@ -8,7 +8,11 @@ es:
   shards: true
   snapshots: true
   cluster_settings: false
-  timeout: 30s
+  {{- if or (contains "m" .Values.elasticsearch.exporter.serviceMonitor.scrapeTimeout) (contains "h" .Values.elasticsearch.exporter.serviceMonitor.scrapeTimeout) }}
+  {{- fail "scrapeTimeout must only be given in seconds" }}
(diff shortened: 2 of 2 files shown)


crssnd (approved):

Xartos (commented):
I'm a bit worried that this might break if anyone changes the scrapeTimeout to something like `1m30s`. But I guess that this will throw an error if that would be the case?

robinelastisys:
> I'm a bit worried that this might break if anyone changes the scrapeTimeout to something like `1m30s`. But I guess that this will throw an error if that would be the case?

Fixed by adding an if-case and throwing an error if the string contains either 'm' or 'h'. We could create a javascript function to convert them into seconds and all but that should be done in a previous step and not down at the "helm"-level if so I suppose?

Xartos (approved):
LGTM

PR #582, opened 2021-08-24
apps sc & wc: ability to change subdomains of endpoints

### LINKED ISSUE #490: Make it possible to change subdomain

We should enable some way overriding the default subdomains we use. Especially the ops domains.

For now we can only control the `opsDomain` which will result in `grafana.<opsDomain>` if we instead would like to use something like `grafana-2.<baseDomain>` this is not possible for now. We can only set the `opsDomain` == `baseDomain` but we cannot control the subdomain.

I would like to be able to set the subdomain that is used for individual services to whatever I want.

e.g.

`sc-config.yaml`
```
global:
  opsDomain: example.com
prometheus:
  grafana:
    ingress:
      subdomain: monitoring
```

Would result in `monitoring.example.com`

This would not work for all the services, especially not for influxdb since wc needs to know this address as well to be able to send metrics and this is only set in sc config (and duplicating this to `wc-config.yaml` feel like a pithole we don't want to fall into).

So I'm not sure how we should solve this issue.

**Additional context**

**Definition of done:**

* I can set the subdomain to whatever I want for all services

robinelastisys:
We should enable some way overriding the default subdomains we use. Especially the ops domains. This is now possible by changing the "subdomain" of the service of your choice in sc-config.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #490 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


robinAwallace:
The pipeline sc-config file needs to be updated as well.

helmfile/values/kube-prometheus-stack-sc.yaml.gotmpl
```diff
@@ -99,7 +99,7 @@ alertmanager:
         # See: https://github.com/helm/helm/issues/2798#issuecomment-467319526
         text: |-
           <!channel>
-          *Cluster:* grafana.{{ .Values.global.opsDomain }}
+          *Cluster:* {{ .Values.user.grafana.subdomain }}.{{ .Values.global.opsDomain }}
```
Xartos, line 102: This should be `.Values.prometheus.grafana.subdomain` because we might want to have those two different.

Same for rest of this file

Xartos (changes requested):
I'm missing the subdomain for ops grafana (could be located at `prometheus.grafana.subdomain` in sc-config.yaml), for elastic and for dex. Maybe even more I'm missing

Xartos (commented):
I'm still missing some subdomains. Like:
* elastic
* harbor.notary
* influxdb
* dex

helmfile/values/harbor-certs.yaml.gotmpl
```diff
@@ -1,5 +1,5 @@
 core:
-  domain: harbor.{{ .Values.global.baseDomain }}
+  domain: {{ .Values.harbor.subdomain }}.{{ .Values.global.baseDomain }}
 
 notary:
-  domain: notary.harbor.{{ .Values.global.baseDomain }}
+  domain: {{ .Values.harbor.notary.subdomain }}.{{ .Values.harbor.subdomain }}.{{ .Values.global.baseDomain }}
```
Xartos, line 5: I think it would make sense to merge these two so this would be
```suggestion
  domain: {{ .Values.harbor.notary.subdomain }}.{{ .Values.global.baseDomain }}
```

And then the default config for notary would be:

```yaml
harbor:
  notary:
    subdomain: notary.harbor
```

Then you could change this to not include harbor if you really wanted.

helmfile/values/harbor.yaml.gotmpl
```diff
@@ -18,8 +18,8 @@ expose:
       nginx.ingress.kubernetes.io/whitelist-source-range: {{ .Values.externalTrafficPolicy.whitelistRange.harbor }}
       {{ end }}
     hosts:
-      core: harbor.{{ .Values.global.baseDomain }}
-      notary: notary.harbor.{{ .Values.global.baseDomain }}
+      core: {{ .Values.harbor.subdomain }}.{{ .Values.global.baseDomain }}
+      notary: {{ .Values.harbor.notary.subdomain }}.{{ .Values.harbor.subdomain }}.{{ .Values.global.baseDomain }}
```
Xartos, line 22: Same here

pipeline/config/exoscale/sc-config.yaml
```diff
@@ -74,6 +75,7 @@ user:
     - severity
 harbor:
   enabled: true
+  subdomain: harbor
```
Xartos, line 78: I think you forgot to add the new subdomains (notary, elastic etc.) to these files (wc-config as well)

Xartos (approved):
LGTM

PR #587, opened 2021-08-30
apps: separate the ops and user grafana dashboards

### LINKED ISSUE #12: [3] Users elasticsearch dashboard in grafana shows errors and no data
The elasticsearch dashboard in the user grafana is empty and shows errors about the datasource. I think this dashboard should not be available to the user. It should only be in the Ops grafana.

crssnd:
**What this PR does / why we need it**: it will allow us to control what grafana dashboards from the grafana-ops chart will be visible to the user

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #12 

1. only dashboards with "labelKey" values set to 1 will be visible to the user. By default all kube-prometheus-stack dashboards. An ops dashboard can be visible to the user if you set "user_visible" to true in grafana-ops/values.yaml
2. the grafana-ops chart template was modified to automatically load the dashboards defined in the values.yaml file
3. we will have one ConfigMap for each dashboard
4. added the prometheus-sc datasource in user-grafana so they will have access to all metrics.

Before:
grafana-user-before

After: 
grafana-user-after

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


Xartos (commented):
Looking really good, nice work!
Although would it be possible to set labelValue to `user` instead of `1`? Seems a bit more descriptive.

crssnd:
> Looking really good, nice work!
> Although would it be possible to set labelValue to `user` instead of `1`? Seems a bit more descriptive.

unfortunately not really possible. 
I used "1" because this is the value hardcoded in [kube-prometheus-stack dashboards](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/grafana/configmap-dashboards.yaml). 
This way they will see this dashboards by default.

Xartos:
> > Looking really good, nice work!
> > Although would it be possible to set labelValue to `user` instead of `1`? Seems a bit more descriptive.
> 
> unfortunately not really possible.
> I used "1" because this is the value hardcoded in [kube-prometheus-stack dashboards](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/grafana/configmap-dashboards.yaml).
> This way they will see this dashboards by default.

Right, well then it makes sense.

Xartos (approved):

viktor-f (commented):
Looks nice!
I just want to note that with this we are giving the users access to all metrics from the service cluster, since we often (sometimes?) give users editor access in grafana. Then they can use whatever metrics they want.
I'm ok with this, but just want to make sure that others are also ok with it.

crssnd:
> Looks nice!
> I just want to note that with this we are giving the users access to all metrics from the service cluster, since we often (sometimes?) give users editor access in grafana. Then they can use whatever metrics they want.
> I'm ok with this, but just want to make sure that others are also ok with it.

We discussed it in an arch meeting and I also highlighted it in the issue [#12](https://github.com/elastisys/compliantkubernetes-apps/issues/12). If needed I can ask for a confirmation.

viktor-f (approved):

PR #593, opened 2021-09-03
apps: added cert-manager http challenge custom repository

crssnd:

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes [#131](https://github.com/elastisys/ck8s-apps/issues/131)

Not sure if I should create the NetworkPolicy for this.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


Xartos (approved):
LGTM

OlleLarsson (approved):

viktor-f (approved):

helmfile/00-psp_rbac_np.yaml
```diff
@@ -0,0 +1,71 @@
+# Defult settings to use with helm.
```
Xartos, line 1: I'm guessing this file should be renamed `00-psp_rbac.yaml`. Now we have two of these

Xartos (changes requested):
The only issue I see is that this is relying on a harbor instance managed by users. If the user deletes the proxy repo then this will stop working.
So I think that I rather would like to see a fix in the opa policy to allow for these particular images rather than using harbor.

helmfile/00-psp_rbac_np.yaml
crssnd: > The only issue I see is that this is relying on a harbor instance managed by users. If the user deletes the proxy repo then this will stop working. So I think that I rather would like to see a fix in the opa policy to allow for these particular images rather than using harbor.

I didn't find a good way o add this exception and to make sure that no one can't take advantage of it. 
Another option would be to create and use an elastisys harbor repo (in a HA mode), we may also need it if we want to enable OPA for all namespaces in the future.

crssnd: I deleted the old 00-psp_rbac.yaml

Xartos:
> I didn't find a good way o add this exception and to make sure that no one can't take advantage of it.
> Another option would be to create and use an elastisys harbor repo (in a HA mode), we may also need it if we want to enable OPA for all namespaces in the future.

Wouldn't just adding `jetstack/cert-manager-acmesolver` to https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/charts/gatekeeper-constraints/templates/image-registry/constraint.yaml#L25 be enough? I mean anyone would be able to create that pod, but is that an issue?

crssnd:
> > I didn't find a good way o add this exception and to make sure that no one can't take advantage of it.
> > Another option would be to create and use an elastisys harbor repo (in a HA mode), we may also need it if we want to enable OPA for all namespaces in the future.
> 
> Wouldn't just adding `quay.io/jetstack/cert-manager-acmesolver` to https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/charts/gatekeeper-constraints/templates/image-registry/constraint.yaml#L25 be enough? I mean anyone would be able to create that pod, but is that an issue?

yes you are totally right, I wrongly assumed that we can only add repositories as exceptions. thanks!
still I do not feel comfortable adding this image as a permanent exception, so I added the possibility to add it via the wc-config.yaml
if the image has any serious vulnerabilities we may not want it to run in our cluster or be created by anyone

Xartos (approved):
Super nice!

helmfile/elastisys/networkpolicy/values.yaml
```diff
@@ -0,0 +1,2 @@
+cert-manager:
+  enabled: false
```
viktor-f, line 2: I think this should also include `namespaces`
```suggestion
cert-manager:
  enabled: false
namespaces: []
```

viktor-f (commented):
This looks like a good solution.
You can move the networkpolicy chart to `helmfile/charts/`, AFAIK we only have "upstream" charts that we have created in `helmfile/elastisys/` (i.e. charts that we own but the main version of them is in some other repo).

PR #598, opened 2021-09-07
apps sc: Added buffer configuration for fluentd

### LINKED ISSUE #586: Fluentd sc - record larger than buffer chunk limit size
**Describe the bug**

Sometimes we encounter records that are larger than the chunk limit.
In essence this will lead to failed health checks resulting in fluentd being killed.

In the log below it actually seems like it's fluentds own log that causes the problem unless i'm not reading it correctly.

```
{"time":"2021-08-27 06:10:00 +0000","level":"warn","message":"emit transaction failed: error_class=Fluent::Plugin::Buffer::BufferChunkOverflowError error=\"a 6726bytes record is larger than buffer chunk limit size\" location=\"/opt/bitnami/fluentd/gems/fluentd-1.11.4/lib/fluent/plugin/buffer.rb:738:in `block in write_step_by_step'\" tag=\"kubernetes.var.log.containers.fluentd-bsgl8_fluentd_fluentd-7f7e5f393893cf07edc91ff23fac0cc4ce0dd4e5e91c194906103874bc0177ca.log\"","worker_id":0}
```

**To Reproduce**
Not sure how to properly reproduce.

**Expected behavior**

**Screenshots**

v0.17.0

**Additional context**
Sometimes this causes a real problem, and sometimes it just complains a few times but manages to survive and continue.

robinAwallace:

This allow us to set the `chunk_limit_size` and `queue_limit_length` from the `sc-config`.  Set these to a reasonable size for the clusters so that fluentd does not complain about `record is larger than buffer chunk limit size`.

fixes #586

This does not fix the problem of crash-loop. That is a separat issue.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/pipeline/config/exoscale/sc-config.yaml
+++ b/pipeline/config/exoscale/sc-config.yaml
@@ -415,6 +415,8 @@ fluentd:
     affinity: {}
     nodeSelector: {}
     useRegionEndpoint: true
+    chunkLimitSizeBytes: 2048
+    queueLimitSizeBytes: 4096
   aggregator:
     resources:
       limits:
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -23,5 +23,6 @@
 - calico-felix-metrics helm chart to enable calico targets discovery and scraping
   calico felix grafana dashboard to visualize the metrics
 - Added JumpCloud as a IDP using dex.
+- Setting chunk limit size and queue limit size for fluentd from sc-config file
 
 ### Removed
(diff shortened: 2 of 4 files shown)


pipeline/config/exoscale/sc-config.yaml
```diff
@@ -415,6 +415,8 @@ fluentd:
     affinity: {}
     nodeSelector: {}
     useRegionEndpoint: true
+    chunkLimitSize: 2048
+    queueLimitSize: 4096
```
OlleLarsson, line 419: I think this makes it a bit more clear on what they "represent"
```suggestion
    chunkLimitSizeBytes: 2048
    queueLimitSizeBytes: 4096
```

OlleLarsson (approved):

Xartos (approved):
LGTM

PR #601, opened 2021-09-08
apps sc: reworked harbor restore script

### LINKED ISSUE #386: Rework Harbor restore script
The current restore script for Harbor feels a bit clunky and adds dependencies on a postgres CLI and the AWS CLI.

I propose to change the script to do the following:
1. download the backup using s3cmd instead of the AWS CLI (since we use that in other places as well).
2. Upload the backup to the database pod
3. Excec into the pod and restore from the backup (then we don't need the postgres client locally)
4. Clean up backup files locally and in the pod.

We should probably also add support for GCS since our backup scripts support storing backups there.

We could do just the change with s3cmd or the change with the postgres client instead of both.

**Additional context**

**Definition of done:**
We have a restore script that has less dependencies on other tools.
We have clear instructions on how to use the script, both in this repository and in the official documentation (complinatkubernetes.io)

robinAwallace:

This PR converts the harbor backup script into a k8s job. This reduces the need to have aws cli and postgres installed locally. Also the documentation on how to restore a backup has been updated, including from specific backups.

It also include a fix for harbor backup job, ensuring it run as a non-root user. 

**Which issue this PR fixes** : 
fixes #386

Tried on safespring with no specific backup and specific backup set. Both worked.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/scripts/restore/README.md
+++ b/scripts/restore/README.md
@@ -1,30 +1,64 @@
 ### Restore Harbor
-With the script `restore-harbor.sh` you can restore the database in Harbor from a backup in S3.
+With the k8s job you can restore the database in Harbor from a backup in S3.
+The steps should be preformed from the `compliantkubernetes-apps` root directory.
+*Note this k8s job does not include restore from gcs.*
 
-Before restoring the database, make sure that Harbor is installed. It can be installed normally.
+Before restoring the database, make sure that Harbor is installed.
+It can be installed normally.
 
-In order to run the restore script you need the aws client and the postgres client installed:
-```bash
-curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
-unzip awscli-bundle.zip
-sudo ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
-sudo apt install postgresql-client-10
-sudo apt install postgresql-client-common
-```
-Set these variables for the restore-script:
+#### Env variables
+Set these variables for the restore job:
 ```bash
 export CK8S_CONFIG_PATH=<your config path>
 export S3_BUCKET=$(yq read "${CK8S_CONFIG_PATH}/sc-config.yaml" objectStorage.buckets.harbor)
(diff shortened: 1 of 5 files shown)


scripts/restore/restore-harbor-job.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: scripts
+        configMap:
+          name: restore-harbor
+      - name: backup
+        emptyDir: {}
+     securityContext:
+      runAsUser: 1000
+      runAsGroup: 1000
+      fsGroup: 1000
+     restartPolicy: OnFailure
```
OlleLarsson, line 20: Maybe?
```suggestion
     restartPolicy: Never
```

scripts/restore/README.md
```diff
@@ -1,30 +1,64 @@
...
+export SPECIFIC_BACKUP=<backups/xxxxxxxxxx.sql.gz> # Optional
+```
+
+#### Restore
+
+Setup the necessary job.yaml and configmap:
+
 ```bash
+envsubst > tmp-job.yaml < scripts/restore/restore-harbor-job.yaml
+./bin/ck8s ops kubectl sc create configmap -n harbor restore-harbor --from-file=scripts/restore/restore-harbor.sh
```
OlleLarsson, line 36: Could we use yq instead of envsubst?

OlleLarsson (approved):

scripts/restore/restore-harbor-job.yaml
robinAwallace: Yeah maybe makes more sense when it is a single time job.

scripts/restore/README.md
robinAwallace: What is the benefits of using yq here instead of envsubst?

OlleLarsson: Envsubst is kind of unpredictable if you aren't careful as it will replace any `$` it finds in the file.
It this case it probably fine

robinAwallace: Ah yeah well that is true. Yes should be fine in this case.

Xartos (approved):
LGTM

PR #652, opened 2021-11-01
apps: Add missing value for prometheus-servicemonitor

aarnq:
Needed to render the prometheus-servicemonitor chart with rook monitor enabled, occurs after the split of alerts/servicemonitors.

I don't really know if the labels are used or if they should have a different value, but with this they will have the same as they would have had before the split.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/helmfile/charts/prometheus-servicemonitor/values.yaml
+++ b/helmfile/charts/prometheus-servicemonitor/values.yaml
@@ -1,3 +1,6 @@
+defaultRules:
+  labels: {}
+
 certMonitor:
   enabled: true
   name: "cert-monitor"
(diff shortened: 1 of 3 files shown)


viktor-f (approved):
LGTM
Maybe check if this should be part of the current release process as well.

> I don't really know if the labels are used or if they should have a different value, but with this they will have the same as they would have had before the split.
The label does matter, at least in sc: https://github.com/elastisys/compliantkubernetes-apps/pull/652

robinAwallace (approved):

PR #657, opened 2021-11-02
bin: Introduce common config

### LINKED ISSUE #602: [3] Common-config

There are multiple variables in sc/wc-config that are "common" (you have to write the same in both config files). This can lead to misconfiguration between the config files. Also the documentation is more clear in the wc-config then in the sc-config. This would solve some of the confusion, especially for new members. 

I suggests that there should be a common configuration which is used when installing apps.

The other alternative is to keep it as it is.

**Additional context**
From sc-config:
```yaml
global:
  ## Compliantkubernetes-apps version.
  ## Use version number if you are exactly at a release tag.
  ## Otherwise use full commit hash of current commit.
  ## 'any', can be used to disable this validation.
  ck8sVersion: ${CK8S_VERSION}
  baseDomain: set-me
  opsDomain: set-me
  issuer: letsencrypt-staging
  verifyTls: true
  clusterDns: 10.233.0.3
```
From wc-config:
```yaml
## Some common options used in various helm charts.
##
global:
  ## Compliantkubernetes-apps version.
  ## Use version number if you are exactly at a release tag.
  ## Otherwise use full commit hash of current commit.
  ## 'any', can be used to disable this validation.
  ck8sVersion: ${CK8S_VERSION}
  ## The cluster name.
  ## Used in logs and metrics as to separate these from other clusters.
  clusterName: ${CK8S_ENVIRONMENT_NAME}-wc
  ## Domain intended for ingress usage in the workload cluster
  ## and to reach user facing services such as kibana, grafana, and harbor.
  ## E.g. with 'prod.domain.com', kibana is reached via 'kibana.prod.domain.com'.
  baseDomain: set-me
  ## Domain intended for ingress usage in the service cluster and to reach
  ## non-user facing services such as influxdb and elasticsearch.
  ## E.g. with 'ops.prod.domain.com', elasticsearch is reached via 'elastic.ops.prod.domain.com'.
  opsDomain: set-me
  ## Default cert-manager issuer to use for issuing certificates for ingresses.
  ## Normally one of 'letsencrypt-staging' or 'letsencrypt-prod'.
  issuer: letsencrypt-staging
  ## Verify ingress certificates
  verifyTls: true
  ## IP of the cluster DNS in kubernetes
  clusterDns: 10.233.0.3
```

Both set of variables need to be set to the same.

**Definition of done:**
When could this feature be considered done.

* I can configure common attributes from a common-config file.

aarnq:
This simplifies the configuration by using a common configuration that applies to both the service and workload cluster for common options.
Includes a few fixes for how data is handled by `yq` in certain situations and a fix to render the prometheus-servicemonitor with rook-monitor enabled.

fixes #602

**Public facing documentation PR**

Please let me know if I should change some configuration variables to make them the same on the service and workload cluster.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

**Pipeline config**
Done and tested.


config/config/common-config.yaml
```diff
@@ -0,0 +1,358 @@
+## Common Kubernetes configuration options.
```
crssnd, line 1: Does this file contain only the common fields between sc and wc configs?

crssnd (approved):
I did some small tests and it looks ok.

config/config/common-config.yaml
aarnq: Yes, it contains all fields where the values was the same between sc and wc configs.

aarnq:
Rebased, cleared the wip-changelog from things not supposed to be there, and set cert-manager cainjector cpu req to 25m so the whole cert-manager config for sc and wc are now the same and fully contained in the common-config.

Xartos (commented):
Quickly read the code, and I got one comment

config/config/common-config.yaml
```diff
@@ -0,0 +1,370 @@
...
+##
+global:
+  ## Compliantkubernetes-apps version.
+  ## Use version number if you are exactly at a release tag.
+  ## Otherwise use full commit hash of current commit.
+  ## 'any', can be used to disable this validation.
+  ck8sVersion: ${CK8S_VERSION}
+  ck8sCloudProvider: ${CK8S_CLOUD_PROVIDER}
+  ck8sEnvironmentName: ${CK8S_ENVIRONMENT_NAME}
+  ck8sFlavor: ${CK8S_FLAVOR}
```
Xartos, line 30: Do we need this variable? Isn't this only used for the first init to decide on default values?

Xartos (commented):
Done some testing, seems to work very well! Just have one comment otherwise I think it looks good

config/config/common-config.yaml
aarnq: As `init` now also can be used to update the default configs by pulling values from the template configs from the repo, this is needed to ensure that these values are not inadvertently changed.
So we basically can't change the config from a Safespring production cluster to an Exoscale dev without warnings if an env is set wrong or missed.

bin/init.bash
```diff
@@ -12,29 +12,29 @@ here="$(dirname "$(readlink -f "$0")")"
...
 
 # Load cloud provider, environment name, and flavor from config if available.
-if [ -f "${config[default_config_file_sc]}" ]; then
-    cloud_provider=$(yq read "${config[default_config_file_sc]}" 'global.ck8sCloudProvider')
-    environment_name=$(yq read "${config[default_config_file_sc]}" 'global.ck8sEnvironmentName')
-    flavor=$(yq read "${config[default_config_file_sc]}" 'global.ck8sFlavor')
+if [ -f "${config[default_common]}" ]; then
+    cloud_provider=$(yq read "${config[default_common]}" 'global.ck8sCloudProvider')
+    environment_name=$(yq read "${config[default_common]}" 'global.ck8sEnvironmentName')
+    flavor=$(yq read "${config[default_common]}" 'global.ck8sFlavor')
```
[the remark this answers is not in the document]
aarnq: I've added to the readme that these are set by `init` and are needed to generate the correct config during upgrades, and how they can be changed if needed by removing the file.
Since these are not intended to be changed or overriden by the user directly I don't think it should be an issue just checking the default common config.

config/config/common-config.yaml
Xartos: Valid argument. Then I'm fine with this :+1:

Xartos (approved):

PR #706, opened 2021-12-08
apps sc: falco alert group-by

### LINKED ISSUE #696: Investigate how to make falco more usable
What technology/product/concept should be investigated?

Investigate if it's possible to refine the config for falco and make the alerts nicer.

Maybe this is just to fix the `group_by` variable in alertmanager and filter some alerts or do we need to do more to fix this.

How should the knowledge gathered from this task be captured.

A decision for how to improve falco alerts

robinelastisys:
We had the issue of that all FalcoAlerts came in the same alert, not separated by a rule. Now, a separation of rules has been made. There are 12 different falco-rules.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #696

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -10,6 +10,8 @@
  - Updated opensearch helm chart to version `1.7.1`, which upgrades opensearch  to `v1.2.4`.
  - Renamed release `blackbox` to `prometheus-blackbox-exporter`.
  - Added new panel to backup dashboard to reflect partial, failed and successful velero backups
+ - Alertmanager group-by parameters was removed and replaced by the special value `...`
+     See https://github.com/prometheus/alertmanager/blob/ec83f71/docs/configuration.md#route for more information
 
 ### Fixed
 
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -86,10 +86,7 @@ user:
   # Todo remove dependencie on alertmanager from service cluster
   alertmanager:
     group_by:
-      - cluster
-      - alertname
-      - severity
-
+      - '...'
 harbor:
   enabled: true
   subdomain: harbor


viktor-f:
I think I previously discussed with someone about testing to remove all of the group-by values, so that everything comes in separate alerts.
Did you try that? If not, could you try that and add some thoughts on which variant works best for us? I'm happy to help with this testing if you want.

robinelastisys:
> I think I previously discussed with someone about testing to remove all of the group-by values, so that everything comes in separate alerts. Did you try that? If not, could you try that and add some thoughts on which variant works best for us? I'm happy to help with this testing if you want.

Screenshot from 2021-12-14 11-09-13
Screenshot from 2021-12-14 11-10-05

robinelastisys:
After some conversation, we decided that less grouping is what we want for more details.

WIP-CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Reworked configuration handling to use a common config in addition to the service and workload configs. This is handled in the same way as the sc and wc configs, meaning it is split between a default and an override config. Running `init` will update this configuration structure, update and regenerate any missing configs, as well as merge common options from sc and wc overrides into the common override.
 - Updated fluentd config to adhere better with upsream configuration
 - Fluentd now logs reasons for 400 errors from elasticsearch
+- 'rule' was added in Falco alerts group-by parameter
```
Xartos, line 17: I guess this one needs to be updated as well

config/config/sc-config.yaml
```diff
@@ -46,8 +46,6 @@ user:
   alertmanager:
     group_by:
       - cluster
```
Xartos, line 48: Should this one still be there?

viktor-f: I don't think so. My guess is that it would group everything from one cluster into one giant alert (though I might be wrong)

robinelastisys: Will see if we can test this in our long-running cluster!

WIP-CHANGELOG.md
```diff
@@ -65,6 +65,7 @@
 - increased the influxDB pvc size [#739](https://github.com/elastisys/compliantkubernetes-apps/pull/739)
 - Exposed velero's backup timetolive for both sc and wc.
 - disabled internal database for InfluxDB
+- Alertmanager group-by parameters was removed and replaced by an empty list
```
Xartos, line 68: Since it's technically not an empty list, I would rather have something like
```suggestion
- Alertmanager group-by parameters was removed and replaced by the special value `...`
    See https://github.com/prometheus/alertmanager/blob/ec83f71/docs/configuration.md#route for more information
```

config/config/sc-config.yaml
```diff
@@ -80,11 +80,7 @@ user:
           memory: 100Mi
   # Todo remove dependencie on alertmanager from service cluster
   alertmanager:
-    group_by:
-      - cluster
-      - alertname
-      - severity
-
+    group_by: ['...']
```
OlleLarsson, line 83: Any reason why not to use a yaml list formatting?
I feel like yq will just convert this to a yaml list eventually.
```suggestion
    group_by:
      - '...'
```

Xartos (approved):

viktor-f (changes requested):
Looks like you rebased this a bit badly, adding a lot of things to the wip-changelog from the previous release. Please remove the things that should not be there.

robinelastisys:
Now it should be correct!

Xartos (approved):

viktor-f (approved):

PR #708, opened 2021-12-10
apps: ingres nginx lb sourceranges

robinAwallace:

So we could whitlist IP address to the loadbalancer service.

fixes #

Two commits so it is possible to cherrypick for older versions of apps.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/helmfile/values/ingress-nginx.yaml.gotmpl
+++ b/helmfile/values/ingress-nginx.yaml.gotmpl
@@ -99,6 +99,7 @@ controller:
     type: {{ .Values.ingressNginx.controller.service.type }}
     annotations:
     {{- .Values.ingressNginx.controller.service.annotations | nindent 6 }}
+    loadBalancerSourceRanges: {{- toYaml .Values.ingressNginx.controller.service.loadBalancerSourceRanges | nindent 6 }}
     {{ else }}
     enabled: false
     {{ end }}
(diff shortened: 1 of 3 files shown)


helmfile/values/ingress-nginx.yaml.gotmpl
```diff
@@ -99,6 +99,7 @@ controller:
     type: {{ .Values.ingressNginx.controller.service.type }}
     annotations:
     {{- .Values.ingressNginx.controller.service.annotations | nindent 6 }}
+    loadBalancerSourceRanges: {{- toYaml .Values.ingressNginx.controller.service.loadBalancerSourceRanges | nindent 6 }}
```
Xartos, line 102: Should this be wrapped in some sort of `if ...` or will this not break anything if you're not using octavia LB?

robinAwallace: Yes that would be a good idé. Should we add a another variable where you have to specify if it is supported? Hopefully tho it is just skipped if it is not supported.

Just a quick look and I see that aws and google supports this.

aarnq: The default in ingress-nginx seems to be an empty array when not used. And it should apply to any load balancer as it is not Octavia specific, though I'm not sure it applies to a service load balancer.

Xartos: Then I'm fine with leaving it as it is

Xartos (approved):

aarnq (approved):

PR #725, opened 2021-12-23
docs: updated the kubectl version

crssnd:
**What this PR does / why we need it**: updated the kubectl version to v1.20.7
https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ The apps are installed using a combination of helm charts and manifests with the
 ### Requirements
 
 * A running cluster based on [compliantkubernetes-kubespray][compliantkubernetes-kubespray]
-* [kubectl](https://github.com/kubernetes/kubernetes/releases) (tested with 1.19.8)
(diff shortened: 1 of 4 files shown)


OlleLarsson (approved):

OlleLarsson:
I think the pipeline should be update as well

crssnd:
> I think the pipeline should be update as well

done

viktor-f (commented):
We already have some customers on k8s 1.21, so maybe we should go ahead and upgrade to that already, instead of 1.20.

OlleLarsson:
> We already have some customers on k8s 1.21, so maybe we should go ahead and upgrade to that already, instead of 1.20.

Yes good catch, don't know why but i read it as 1.21.7 😅

viktor-f (approved):
From slack: will stay on 1.20 to be able to support 1.19 for a little while longer.

PR #728, opened 2021-12-28
apps: index per namespace

### LINKED ISSUE #603: Make curator delete indices based on name timestring instead of creation date
We had this before we migrated to opendistro, see [this](https://github.com/elastisys/ck8s/pull/526/files) pr.
With the migration to opendistro I seem to have made a mistake and not include that change in the curator action config.

Apply changes [here](https://github.com/elastisys/opendistro-build/commit/9ae1fbac51a81b4dd838fe2c2c9f33cec2d6a8a1#diff-52b31dc31f80a27245ea4d395fc885d2a831c171cc645df79112b0fb525e1ad6).

None.

**Additional context**
This ensures that old logs are removed before newer ones even if certain indices have been reindexed.

**Definition of done:**
Curator removes indices (age-wise) based on timestring instead of creation date.

### LINKED ISSUE #693: Make it possible to have one index per namespace

Currently, all application logs end up in the same Elasticsearch index pattern `kubernetes*`, e.g., `kubernetes-default-2021.11.25-000010`.

In certain environments, it is desirable to have one index per namespace. This brings the following short-term benefits:

- Less index mapping issues: Each application team can have their own fields convention, without causing index mapping headaches.

This also opens up the path to long-term benefits, including:

- Per-namespace retention, e.g., via separate curator jobs.
- Access control to logs, e.g., `team1` -- identified via OpenID group -- can only see logs in `demo1`.

- A configuration flag that enables per-namespace indexes.

- Not offering this feature, but customers seem to really want it and it seems to open up quite some benefits..

**Additional context**

**Definition of done:**

- I can configure Compliant Kubernetes to write container logs to an index named based on the namespace in which the containers are running.
- A screenshot is available (in this issue) showing that the `user-demo` (NodeJS) deployed in `demo1` and `user-demo-dotnet` (.NET) deployed in `demo2` write logs to separate indices.

OlleLarsson:
To allow for a different setup where logs will be stored in indices based on the namespace from which they originate.
This is an alpha feature that has not been thoroughly tested,

fixes #693 
fixes #603 

**Public facing documentation PR**
Leaving this for now.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


migration/v0.18.x-v0.19.x/curator-retention.sh
```diff
@@ -0,0 +1,38 @@
...
+echo -e "Retention settings:\n$retention"
+
+otherGB=$(yq r "${CK8S_CONFIG_PATH}"/sc-config.yaml elasticsearch.curator.retention.otherSizeGB)
+otherAge=$(yq r "${CK8S_CONFIG_PATH}"/sc-config.yaml elasticsearch.curator.retention.otherAgeDays)
+authlogGB=$(yq r "${CK8S_CONFIG_PATH}"/sc-config.yaml elasticsearch.curator.retention.authLogSizeGB)
+authlogAge=$(yq r "${CK8S_CONFIG_PATH}"/sc-config.yaml elasticsearch.curator.retention.authLogAgeDays)
+kubeauditGB=$(yq r "${CK8S_CONFIG_PATH}"/sc-config.yaml elasticsearch.curator.retention.kubeAuditSizeGB)
+kubeauditAge=$(yq r "${CK8S_CONFIG_PATH}"/sc-config.yaml elasticsearch.curator.retention.kubAauditAgeDays)
+kubernetesGB=$(yq r "${CK8S_CONFIG_PATH}"/sc-config.yaml elasticsearch.curator.retention.kubernetesSizeGB)
+kubernetesAge=$(yq r "${CK8S_CONFIG_PATH}"/sc-config.yaml elasticsearch.curator.retention.kubernetesAgeDays)
```
aarnq, line 18: Won't this create an issue if only some parts are set, as then they would be read as null?

Xartos: Agree on this, maybe we'll need to check if it's set in this file otherwise use from the `default/sc-config.yaml` file. Or some yq merge magic I guess

helmfile/values/opendistro-es.yaml.gotmpl
```diff
@@ -275,7 +275,11 @@ elasticsearch:
 
 
     ############## Elasticsearch configuration settings ##############
+    {{- if .Values.elasticsearch.indexPerNamespace }}
+    action.auto_create_index: "-kubeaudit,-authlog,-other,+*"
```
Xartos, line 279: I probably just don't understand this particular config. But I'm guessing this is some index pattern thing. So should you really have the `-` sign before all these? And are you creating one pattern for `+*`?

Xartos (commented):
Looks good, just some questions since I don't understand it too well

helmfile/values/opendistro-es.yaml.gotmpl
OlleLarsson: `-` means dissallow, `+` means allow.

Fluentd should be be able to create `kubeaudit/other/authlog` indexes, remember it still writes to those aliases even with this new featre. It's not likely but can happen unless the configurer has gotten around to create the indices before fluentd starts to write to the aliases.

helmfile/values/opendistro-es.yaml.gotmpl
```diff
@@ -565,20 +561,30 @@ configurer:
           allowed_actions:
           - "indices_monitor"
         - index_patterns:
+          {{- if .Values.elasticsearch.indexPerNamespace }}
+          - "*"
```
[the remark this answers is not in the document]
OlleLarsson: We can do that.

helmfile/values/opendistro-es.yaml.gotmpl
OlleLarsson: And yeah maybe we could remove the `+*`, it didn't test without.

WIP-CHANGELOG.md
```diff
@@ -52,6 +54,7 @@
 - Added the possibility to whitelist IP addresses to the loadbalancer service
 - Added pwgen and htpasswd as requirements
 - Added the blackbox installation in the wc cluster based on ADR to monitor the uptime of internal services as well in wc .
+- Added option to enable index per namepspace feature in fluentd and elasticsearch
```
crssnd, line 57: small typo in namespace

migration/v0.18.x-v0.19.x/curator-retention.sh
OlleLarsson: There's no reason for us to do anything if it is not present in the `defaults` config.
But yes I will make update the script to include checks.

migration/v0.18.x-v0.19.x/curator-retention.sh
```diff
@@ -0,0 +1,36 @@
...
+  "
+
+  [ -n "${size}" ] && patch+="
+        sizeGB: ${size}
+  "
+
+  [ -n "${age}" ] && patch+="
+        ageDays: ${age}
+  "
+done
```
aarnq, line 34: This will still be problematic since `elasticsearch.curator.retention` is an array, and when the config is merging it will be overwriting arrays.
So if only a single property is set, say `otherSizeGB`, then the whole list needs to copied over and set otherwise you would only end up with the entry for `other`.

OlleLarsson: I have updated the script again. Now it should merge with the defaults config.

Xartos (approved):
LGTM

aarnq (approved):

PR #736, opened 2022-01-11
apps sc: Add off-site backup replication using rclone sync

### LINKED ISSUE #318: Upstream rclone for off-site backups

When running production workloads, we use rclone to replicate backups from one region to another. Please upstream this code to Compliant Kubernetes.

Optional, disabled-by-default, configuration to enable rclone. It should clone the backup S3 buckets from the default region to an off-site one.

Keep the code hidden from our users. We agreed this is not aligned with our product strategy.

**Definition of done:**

* I can enable off-site replication of backups by editing `*-config.yaml`. This is disabled by default.

aarnq:
Upstreams the work done on enabling off-site backup replication using rclone sync.
Includes additional section in the "Backup status" dashboard to show the
state of it and successful sync jobs.

fixes #318 

Pictures of dashboard additions, note that the text has been changed to "Off-site backup replication is..." instead to make it a bit clearer.

Enabled, with one suspended cronjob:
enabled

Disabled, with all cronjobs and jobs removed:
disabled

The state of the cronjobs and jobs are extracting the bucket name using a regex, and it has been tested to work with a suffix containing `manual`.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -10,6 +10,42 @@ objectStorage:
     opensearch: ${CK8S_ENVIRONMENT_NAME}-opensearch
     scFluentd: ${CK8S_ENVIRONMENT_NAME}-sc-logs
 
+  ## Off-site backup replication between two providers or regions using rclone sync
+  sync:
+    enabled: false
+    # dryrun: false
+
+    ## Options are 's3'
+    type: none
+    # s3:
+    #   region: set-me
+    #   regionEndpoint: set-me
+    #   # Generally false when using AWS and Exoscale and true for other providers.
+    #   forcePathStyle: set-me
+    #   # v2Auth: false
+
+    ## Sync all buckets under 'objectStorage.buckets'
+    ## These will be appended to 'buckets' using the same name from source as destination, and the default schedule.
+    syncDefaultBuckets: false
+
+    ## Default schedule for sync jobs
+    defaultSchedule: 0 5 * * *
+
+    ## Buckets to sync.
+    buckets: []
+      # - source: source-bucket
+      #   # destination: destination-bucket # if unset it will be the same as 'source'
+      #   # schedule: 0 5 * * *             # if unset it will be the same as 'defaultSchedule'
+
+    ## Sync job resources
+    resources:
+      requests:
+        cpu: 100m
+        memory: 100Mi
+      limits:
+        cpu: 300m
+        memory: 300Mi
+
 user:
   grafana:
     enabled: true
--- a/helmfile/charts/rclone-sync/Dockerfile
+++ b/helmfile/charts/rclone-sync/Dockerfile
@@ -1,0 +1,31 @@
+FROM ubuntu:latest as download
+
+# Install dependencies
+RUN apt-get update && apt-get install -y curl unzip
+
+# Install rclone
+RUN curl -O https://downloads.rclone.org/rclone-current-linux-amd64.zip && \
+    unzip rclone-current-linux-amd64.zip && \
+    cd rclone-*-linux-amd64 && \
+    install rclone /usr/bin/rclone
+
+FROM ubuntu:latest as final
+
+
+# Install root certificates
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install -y ca-certificates && \
+    apt-get clean -y
+
+# Copy in rclone
+COPY --from=download /usr/bin/rclone /usr/bin/rclone
+
+# Create rclone user
+RUN adduser --system --uid 10000 rclone
+
+# Run as rclone user
+USER 10000
+
+ENTRYPOINT ["rclone"]
+CMD ["--version"]
--- a/helmfile/charts/rclone-sync/files/rclone.conf
+++ b/helmfile/charts/rclone-sync/files/rclone.conf
@@ -1,0 +1,21 @@
+[{{ .Values.config.source.name }}]
+type = {{ .Values.config.source.type }}
+{{- if eq .Values.config.source.type "s3" }}
+access_key_id = {{ .Values.config.source.s3.accessKey }}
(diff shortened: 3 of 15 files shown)


config/config/sc-config.yaml
```diff
@@ -10,6 +10,35 @@ objectStorage:
...
+    type: none
+    # s3:
+    #   region: set-me
+    #   regionEndpoint: set-me
+    #   # Generally false when using AWS and Exoscale and true for other providers.
+    #   forcePathStyle: set-me
+    #   # v2Auth: false
+
+    ## Buckets to sync.
+    buckets: []
```
Xartos, line 28: Would it make sense to have a `syncDefaultBuckets` option that would add all the buckets from the `objectStorage.buckets` list? If it's too cumbersome, then we can have that as a separate issue.

helmfile/values/rclone-sync.yaml.gotmpl
```diff
@@ -0,0 +1,36 @@
...
+{{- end }}
+  source:
+    name: default
+    type: {{ .Values.objectStorage.type }}
+    s3:
+      accessKey: {{ .Values.objectStorage.s3.accessKey }}
+      secretKey: {{ .Values.objectStorage.s3.secretKey }}
+      region: {{ .Values.objectStorage.s3.region }}
+      regionEndpoint: {{ .Values.objectStorage.s3.regionEndpoint }}
+      forcePathStyle: {{ .Values.objectStorage.s3.forcePathStyle }}
```
Xartos, line 17: I guess this should to be wrapped in a `{{ if eq .Values.objectStorage.type "s3" }}`
since we have the thing in the beginning of the file it shouldn't be an issue now, but I think it would look better to include it already

helmfile/charts/rclone-sync/templates/cronjob.yaml
```diff
@@ -0,0 +1,45 @@
...
+                {{- end }}
+                - --log-level
+                - DEBUG
+                {{- if $.Values.config.dryrun }}
+                - --dry-run
+                {{- end }}
+              resources: {{ toYaml $.Values.resources | nindent 16 }}
+              volumeMounts:
+                - name: {{ $.Chart.Name }}-config
+                  mountPath: /root/.config/rclone/
```
Xartos, line 40: Since this is our image. I think it would look better if we updated the image to not use the root user. If it already runs as non-root, then I think we should put this somewhere else (like `/tmp/rclone` or something) and point to it.
Then we could also add the `CMD ["rclone"]` so that you could skip the `command: ["rclone"]` line

Xartos (commented):
Looks overall good, just some things

config/config/sc-config.yaml
aarnq: Yes, I can add that.

helmfile/charts/rclone-sync/templates/cronjob.yaml
aarnq: Definitely

helmfile/charts/rclone-sync/files/rclone.conf
```diff
@@ -0,0 +1,25 @@
+[{{ .Values.config.source.name }}]
+type = {{ .Values.config.source.type }}
+env_auth = false
```
OlleLarsson, line 3: Looking at the documentation this option ([--s3-env-auth](https://rclone.org/s3/#s3-env-auth)) is false by default.

helmfile/charts/rclone-sync/files/rclone.conf
```diff
@@ -0,0 +1,25 @@
...
+{{- if eq .Values.config.source.type "s3" }}
+access_key_id = {{ .Values.config.source.s3.accessKey }}
+secret_access_key = {{ .Values.config.source.s3.secretKey }}
+region = {{ .Values.config.source.s3.region }}
+endpoint = {{ .Values.config.source.s3.regionEndpoint }}
+force_path_style = {{ .Values.config.source.s3.forcePathStyle }}
+v2_auth = {{ .Values.config.source.s3.v2Auth }}
+acl = private
```
OlleLarsson, line 11: Maybe we can just not set this option and let it be [default](https://rclone.org/s3/#s3-acl)?

helmfile/charts/rclone-sync/templates/cronjob.yaml
aarnq: I've added an updated Dockerfile.
Can someone with access to our Docker hub repos build the image, tag it with version `1.2.0` and push it to the `elastisys/rclone-sync` repo?

helmfile/charts/rclone-sync/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM alpine:3.15 as download
```
robinAwallace, line 1: We have seen DNS problems when using alpine image compared to Ubuntu image. I have not found any definitive answer that it is fixed yet. So i would suggest to switch to a Ubuntu image.

helmfile/charts/rclone-sync/templates/cronjob.yaml
Xartos: There is now a `1.2.0` version pushed

helmfile/charts/rclone-sync/Dockerfile
aarnq: Sure I can change that.

aarnq:
I've updated the image to use an Ubuntu base. Someone with access please build and push it with the tag `1.3.0` if it looks alright.

OlleLarsson:
> I've updated the image to use an Ubuntu base. Someone with access please build and push it with the tag `1.3.0` if it looks alright.

Done

PR #737, opened 2022-01-11
apps: added relabling for rook/cert servicemonitor

### LINKED ISSUE #721: prometheus-servicemonitor - fix the relabelings
**Describe the bug**
the relabelings for the cert-manager and rook service monitors are not correct

**Expected behaviour**
both service monitors should contain the correct relabelings names

 - Compliant kubernetes apps version 0.18

**Definition of Done**
The service monitors for certmanager and rook should have the correct label:
    - targetLabel: cluster
      replacement: service_cluster

robinAwallace:

The relabelings for the cert-manager and rook service monitors are not correct

fixes #721

Notice in second image the addition of cluster: service_cluster

Before:
monitor1

After:
monitor2

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/helmfile/values/sc-servicemonitor.yaml.gotmpl
+++ b/helmfile/values/sc-servicemonitor.yaml.gotmpl
@@ -4,6 +4,12 @@ defaultRules:
 
 rookMonitor:
   enabled: {{ .Values.monitoring.rook.enabled }}
+  relabelings:
+  - targetLabel: cluster
+    replacement: service_cluster
 
 certMonitor:
   enabled: true
+  relabelings:
+  - targetLabel: cluster
+    replacement: service_cluster
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -50,6 +50,7 @@
 - Fixed rendering of new prometheus alert rule to allow it to be admitted by the operator
 - Fixed rendering of s3-exporter to be idempotent
 - Fixed bug where init'ing a config path a second time without the `CK8S_FLAVOR` variable set would fail.
+- Fixed relabeling for rook-ceph and cert servicemonitor.
 
 ### Added
 
--- a/helmfile/charts/prometheus-servicemonitor/templates/cert-monitor.yaml
+++ b/helmfile/charts/prometheus-servicemonitor/templates/cert-monitor.yaml
@@ -14,4 +14,6 @@ spec:
       app.kubernetes.io/instance: {{ .Values.certMonitor.target.instance }}
   endpoints:
   - targetPort: {{ .Values.certMonitor.port }}
+    relabelings:
+    {{- toYaml .Values.certMonitor.relabelings | nindent 6 }}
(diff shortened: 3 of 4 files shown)


Xartos (approved):
LGTM

helmfile/values/sc-servicemonitor.yaml.gotmpl
```diff
@@ -4,6 +4,12 @@ defaultRules:
 
 rookMonitor:
   enabled: {{ .Values.monitoring.rook.enabled }}
+  relabelings:
+  - targetLabel: cluster
+    replacement: service_cluster
```
Xartos, line 9: Will not this one be in wc as well?

Xartos: Then you might need to add some wrapper around this to only affect the values in a `{{ if eq .Environment.Name "service_cluster" }}` otherwise we will get this in wc as well

crssnd: yes, for wc we do not do relabelling at the servicemonitor level
we use externalLabels in prometheusSpec instead 
https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/kube-prometheus-stack-wc.yaml.gotmpl#L56

Xartos: We might want to add `externalLabels` here as well. Just to be sure that we get these labels for all metrics

robinAwallace: @Xartos will add that :+1:

crssnd: there is a separate issue for that, as it may have some undesired effects

crssnd: I do not this we need an extra wrapper, we already have it here:
https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/50-applications.yaml#L200

but correct me if I am wrong

OlleLarsson: Do not add external labels in prometheus-sc.
Previously there were some caveats when using external labels do definitely do not fit our use of prometheus in the service cluster.

OlleLarsson: Externallabels are as the name suggest - external 😄, so if prometheus is configured with externallabels set, they will be removed from the metrics once fetched from remote storage. At least that is how i remember it.

Xartos: Right, just found an old message about it. It seems to be a prometheus with influx as backend issue, so (at least with thanos) this shouldn't be an issue anymore

crssnd (approved):

PR #756, opened 2022-01-19
Added option to enable Thanos 

### LINKED ISSUE #742: [5] Replace InfluxDB with Thanos

As discussed during the arch meeting on 2022-01-13, we want to move forward and replace InfluxDB with Thanos for long-term metrics storage. Thanos won the competition due to:

- Governance: It is a CNCF project.
- Licence: Open source (Apache 2.0).
- Popularity: Most popular among the candidates (9900 stars, 1500 forks, used by many big players, e.g., Banzai clouds, Ebay, Jetstack)
- Features: remote_read, remote_write, alerting (recommends Prometheus AlertManager), compaction, downsampling, deduplication, S3 backend
- Security and compliance: HA, disaster recovery not needed, simple upgrading and simple patching, good security track record
- Migration: A process exists for migrating [away from InfluxDB](https://github.com/thanos-io/thanos/issues/2112).

Time to take the next step and actual implement Thanos in Compliant Kubernetes.

1. A flag to enable and configure Thanos, e.g., `thanos.enabled`.
2. A flag to disable InfluxDB, e.g., `influxdb.enabled`.
3. A migration guide usable for production systems, i.e., migration can be done without losing metrics and with minimum downtime to the observability stack.
4. We decided for a "push-style" architecture, i.e., the workload cluster pushes metrics to the service cluster, as done today. We will later re-evaluate this decision.
5. Updated onboarding guide and documentation.

* Keeping InfluxDB and optimizing it.
* M3DB, VictoriaMetrics: Rejected mainly due to not being CNCF projects.
* Cortex: Very similar to Thanos, rejected mainly due to lack of downsampling. (Otherwise, the competition was tough, but there can be only one winner. :smile:)

**Additional context**

N/A

**Definition of done:**

The solution as described above.

Xartos:

Add option to enable thanos, and to disable influxdb

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #742 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


config/config/common-config.yaml
```diff
@@ -165,6 +166,23 @@ prometheus:
   affinity: {}
   nodeSelector: {}
 
+thanos:
+  enabled: false
```
cristiklein, line 170: Can you add a comment for this config?

helmfile/values/grafana-user.yaml.gotmpl
```diff
@@ -70,6 +71,22 @@ datasources:
       type: prometheus
       url: http://kube-prometheus-stack-prometheus:9090
       version: 1
+    {{- end }}
+    {{- if and .Values.thanos.enabled .Values.thanos.receiver.enabled }}
+    {{- range .Values.global.clustersMonitoring }}
+    - name: "Ck8s metrics"
```
crssnd, line 77: we may also need sc data source, e.g.: elasticsearch dashboard needs sc metrics

Xartos: Right, they'll need it for that. I thought we could get away from them needing sc cluster data. But I'll add that for them

crssnd:
does it make sense to add 
```
prometheus:
  prometheusSpec:
    externalLabels:
      cluster: {{ .Values.global.clusterName }}
```
in helmfile/values/kube-prometheus-stack-sc.yaml.gotmpl 

this will help use remove all the serviceMonitor relabelings and with alerts grouping

Xartos:
> does it make sense to add, as part of this PR, the externalLabels for sc prometheus in helmfile/values/kube-prometheus-stack-sc.yaml.gotmpl?
> 
> ```
> prometheus:
>   prometheusSpec:
>     externalLabels:
>       cluster: {{ .Values.global.clusterName }}
> ```
> 
> this will help use remove all the serviceMonitor relabelings and with alerts grouping

As someone stated, this breaks some backup/restore functionality with influx. So I think that we start with this in a release (depricate influx) and for the release where we remove influx completely we add the externalLabel and remove the relabeling.

crssnd:
> > does it make sense to add, as part of this PR, the externalLabels for sc prometheus in helmfile/values/kube-prometheus-stack-sc.yaml.gotmpl?
> > ```
> > prometheus:
> >   prometheusSpec:
> >     externalLabels:
> >       cluster: {{ .Values.global.clusterName }}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this will help use remove all the serviceMonitor relabelings and with alerts grouping
> 
> As someone stated, this breaks some backup/restore functionality with influx. So I think that we start with this in a release (depricate influx) and for the release where we remove influx completely we add the externalLabel and remove the relabeling.

can we try using a relabel_action under the remote_write for thanos?
I would like to have the correct cluster label value in thanos from the beginning.

config/config/sc-config.yaml
```diff
@@ -2,6 +2,8 @@
 # It will override settings set in "defaults/common-config.yaml".
 global:
   clusterName: ${CK8S_ENVIRONMENT_NAME}-sc
+  clustersMonitoring:
+    - ${CK8S_ENVIRONMENT_NAME}-wc
```
viktor-f, line 6: `clustersMonitoring` could also have a comment explaining the config.

viktor-f (commented):
Could you update the tests and the pipeline as well?

Xartos:
> Could you update the tests and the pipeline as well?

Would you like to enable thanos only? I'm a little bit split about this. Because right now thanos is an opt in thing. But since we are sure that it's going to be replacing influx it might be a smart idea to make it the default here right now.

viktor-f:
> > Could you update the tests and the pipeline as well?
> 
> Would you like to enable thanos only? I'm a little bit split about this. Because right now thanos is an opt in thing. But since we are sure that it's going to be replacing influx it might be a smart idea to make it the default here right now.

I'm also not sure about enabling only thanos. I could see having both (the same way we will use it while migrating away from influxdb), then removing influx from the pipeline after a while.

pipeline/test/services/service-cluster/testPodsReady.sh
```diff
@@ -128,6 +144,21 @@ fi
...
     statefulsets+=("fluentd fluentd")
 fi
+if [[ "${enable_thanos}" == "true" ]] && [[ "${enable_thanos_receiver}" == "true" ]]; then
+    deployments+=(
+        "thanos thanos-receiver-receive"
+        "thanos thanos-receiver-storegateway"
+    )
+fi
+if [[ "${enable_thanos}" == "true" ]] && [[ "${enable_thanos_ruler}" == "true" ]]; then
+    deployments+=("thanos thanos-receiver-ruler")
```
viktor-f:
```suggestion
    statefulsets+=(
        "thanos thanos-receiver-receive"
        "thanos thanos-receiver-storegateway"
    )
fi
if [[ "${enable_thanos}" == "true" ]] && [[ "${enable_thanos_ruler}" == "true" ]]; then
    statefulsets+=("thanos thanos-receiver-ruler")
```

pipeline/test/services/service-cluster/testEndpoints.sh
```diff
@@ -29,7 +30,10 @@ if [ "$enable_harbor" == true ]; then
 fi
 
 testEndpoint Grafana "https://${grafana_ops_subdomain}.${ops_domain}/"
-testEndpoint InfluxDB "https://${influxdb_subdomain}.${ops_domain}/health"
+
+if [ "$enable_influxdb" == true ]; then
+    testEndpoint InfluxDB "https://${influxdb_subdomain}.${ops_domain}/health"
+fi
```
viktor-f: You can add a test for the thanos receiver endpoint here.

viktor-f (commented):
Should we add the thanos receiver endpoint (and other ops.endpoints) to blackbox?

Xartos:
> Should we add the thanos receiver endpoint (and other ops.endpoints) to blackbox

Yes, good idea

config/config/sc-config.yaml
```diff
@@ -202,7 +207,14 @@ harbor:
     enabled: true
     retentionDays: 7
 
-
+grafanaLabelEnforcer:
```
OlleLarsson, line 210: Is this a thanos component?

config/config/sc-config.yaml
```diff
@@ -290,6 +302,126 @@ dex:
...
     # SASecretName: set-me
 
+thanos:
+  query:
+    # Enables the query/query-frontend components
+    enabled: true
+
+    # Number of query pods
+    # They are infinitly scalable
+    replicaCount: 1
```
OlleLarsson, line 312: So this is used for both query and query frontend nr of replicas?

viktor-f (approved):
I can't say that I know enough to say that all of this is correct. But it looks good to me.

config/config/sc-config.yaml
```diff
@@ -290,6 +302,126 @@ dex:
...
+    # Retention for the metrics in the receiver
+    tsdbRetention: 15d
+
+    # Number of receiver instances
+    replicaCount: 2
+
+    # How many nodes that a time-series needs to be written to for a write to be considered successful.
+    # It needs to write to (REPLICATION_FACTOR + 1)/2 nodes for success.
+    # For more info: https://thanos.io/v0.24/proposals-done/201812-thanos-remote-receive.md/#replication
+    replicationFactor: 1
```
OlleLarsson, line 372: So only  1 as default.
Any benefits making this 2?

config/config/sc-config.yaml
Xartos: Nope, this is just for the query pods. I think that the frontend pods shouldn't need to be scaled that much.

config/config/sc-config.yaml
Xartos: If we make this 2 then we `(REPLICATION_FACTOR + 1)/2` == `(2 + 1) / 2 = 1.5` aka needs to write to 1 node so that will be the same.
But if we set it to 3 so that it'll be `(3+1)/2 = 2` then the distributor will need to write to both receivers before a write is successful. I don't think we want this since half the point of having 2 is that 1 should be able to be down and it should still work.

config/config/sc-config.yaml
OlleLarsson: Okay sure.

OlleLarsson (approved):

PR #785, opened 2022-01-28
apps sc: expose max seconds for curl and verbose curl for slm job

robinAwallace:

Have noticed that slm jobs crashes because of error 28 which is curls error code for timeout. As such i exposed the max seconds for requests for curl. Also changed --silent to --no-progress-meter to make the curl request more verbose. 

From curl man:
```
--no-progress-meter

Option to switch off the progress meter output without muting or 
otherwise affecting warning and informational messages like --silent does.
```

```
Error code 28

Operation timeout. The specified time-out period was reached according to the conditions.

```

```
kubectl describe pod -n elastic-system elasticsearch-slm-manual-x9kpc
Name:         elasticsearch-slm-manual-x9kpc
Namespace:    elastic-system
...
Status:       Succeeded
IP:           10.233.115.12
IPs:
  IP:           10.233.115.12
Controlled By:  Job/elasticsearch-slm-manual
Containers:
  run:
    ...
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Fri, 28 Jan 2022 08:37:23 +0100
      Finished:     Fri, 28 Jan 2022 08:40:57 +0100
    Last State:     Terminated
      Reason:       Error
      Exit Code:    28
      Started:      Fri, 28 Jan 2022 08:25:55 +0100
      Finished:     Fri, 28 Jan 2022 08:37:21 +0100
    Ready:          False
    Restart Count:  1
    ...
Events:
  Type    Reason     Age                  From               Message
  ----    ------     ----                 ----               -------
  Normal  Scheduled  20m                  default-scheduler  Successfully assigned elastic-system/elasticsearch-slm-manual-x9kpc to pwelt-sc-cluster-k8s-node-b-xlarge-1
  Normal  Pulled     8m34s (x2 over 20m)  kubelet            Container image "elastisys/curl-jq:ubuntu" already present on machine
  Normal  Created    8m33s (x2 over 20m)  kubelet            Created container run
  Normal  Started    8m33s (x2 over 20m)  kubelet            Started container run
```

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -638,6 +638,7 @@ opensearch:
     min: 7
     max: 14
     ageSeconds: 864000
+    maxRequestSeconds: 1200
     backupSchedule: 0 */2 * * *
     backupStartingDeadlineSeconds: 600
     backupActiveDeadlineSeconds: 600
--- a/helmfile/charts/opensearch/slm/values.yaml
+++ b/helmfile/charts/opensearch/slm/values.yaml
@@ -11,8 +11,8 @@ opensearch:
   userSecret: opensearch-snapshotter-user
   clusterEndpoint: opensearch-cluster-master:9200
 
-startingDeadlineSeconds: 300
-activeDeadlineSeconds: 600
+startingDeadlineSeconds: 600
+activeDeadlineSeconds: 2700
 
 schedule: "@daily"
 snapshotRepository: s3
@@ -20,6 +20,7 @@ snapshots:
   min: 7
   max: 14
   maxAgeSeconds: 864000 # = 60 * 60 * 24 * 10 = 10 days
+  maxRequestSeconds: 1200
 
 restartPolicy: OnFailure
 
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -12,6 +12,8 @@
  - Added new panel to backup dashboard to reflect partial, failed and successful velero backups
  - Alertmanager group-by parameters was removed and replaced by the special value `...`
      See https://github.com/prometheus/alertmanager/blob/ec83f71/docs/configuration.md#route for more information
+ - Exposed opensearch-slm-job max request seconds for curl.
+ - Made opensearch-slm-job more verbose when using curl.
 
 ### Fixed
 
--- a/helmfile/charts/opensearch/slm/scripts/slm-retention.bash
+++ b/helmfile/charts/opensearch/slm/scripts/slm-retention.bash
@@ -18,22 +18,24 @@ set -euo pipefail
 : "${MAX_SNAPSHOTS:?Missing MAX_SNAPSHOTS}"
 : "${MAX_AGE_SECONDS:?Missing MAX_AGE_SECONDS}"
 : "${SNAPSHOT_REPOSITORY:?Missing SNAPSHOT_REPOSITORY}"
+: "${REQUEST_TIMEOUT_SECONDS:?Missing REQUEST_TIMEOUT_SECONDS}"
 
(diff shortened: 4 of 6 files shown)


config/config/sc-config.yaml
```diff
@@ -640,6 +640,7 @@ opensearch:
     min: 7
     max: 14
     ageSeconds: 864000
+    maxRequestSeconds: 600
```
Xartos, line 643: If we get timeout when we've used this. Should we increase it? If so I think we also need to change the activeDeadline for this as well.

Xartos (commented):
Looks good, just one thing

config/config/sc-config.yaml
OlleLarsson: Yes we'll need to change the deadlineseconds as well :)

robinAwallace: Yes true about the activeDeadline. I did not change the default value as it works for some environments. But i can just double it.

config/config/sc-config.yaml
```diff
@@ -640,11 +640,12 @@ opensearch:
     min: 7
     max: 14
     ageSeconds: 864000
+    maxRequestSeconds: 1200
     backupSchedule: 0 */2 * * *
     backupStartingDeadlineSeconds: 600
     backupActiveDeadlineSeconds: 600
     retentionSchedule: '@daily'
-    retentionStartingDeadlineSeconds: 600
+    retentionStartingDeadlineSeconds: 1200
```
OlleLarsson, line 648:
```suggestion
    retentionStartingDeadlineSeconds: 600
```

helmfile/charts/opensearch/slm/values.yaml
```diff
@@ -12,14 +12,15 @@ opensearch:
   clusterEndpoint: opensearch-cluster-master:9200
 
 startingDeadlineSeconds: 300
-activeDeadlineSeconds: 600
+activeDeadlineSeconds: 1200
```
OlleLarsson, line 15:
```suggestion
activeDeadlineSeconds: 600
```

OlleLarsson (commented):
When looking we already expose the activeDeadline for the slm job in the [sc-config](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/config/sc-config.yaml#L648). And it seems to be set to 45min by default which should be enough for most environments.

robinAwallace:
> When looking we already expose the activeDeadline for the slm job in the [sc-config](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/config/sc-config.yaml#L648). And it seems to be set to 45min by default which should be enough for most environments.

Yes i saw that. :+1:

OlleLarsson (approved):

config/config/sc-config.yaml
Xartos: I still think that we need to increase the activeDeadline. Because right now it's hardcoded to 600 seconds so it will kick in before the request timeout.

robinAwallace: It is was already set to 45min https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/config/sc-config.yaml#L646

aarnq (approved):

config/config/sc-config.yaml
Xartos: Right, I missed that! But I still think that the default values in the chart should match https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/charts/opensearch/slm/values.yaml#L15

PR #864, opened 2022-02-28
Patch 0.19 prometheus rule fix

aarnq:
Fixes for the v0.19 release:
- Prometheus rules
- OpenSearch migration script

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.


crssnd:
do we want to modify directly the upstream kube-prometheus chart?
everything will be overwritten when we will upgrade it

aarnq:
> do we want to modify directly the upstream kube-prometheus chart? everything will be overwritten when we will upgrade it

This is just for the patch, for the next release with the in progress thanos ruler pr we won't use the upstream rules as we need to split the alerts and records, and the same problem was found there as well so it fixes it in a different way.

CHANGELOG.md
```diff
@@ -17,6 +18,15 @@
 - [v0.5.0](#v050---2020-08-06)
 <!-- END TOC -->
 
+-------------------------------------------------
+## v0.19.1 - 2022-02-28
+
+
+### Fixed
+
+- Disabled prometheus rules for the kube-apiserver which lacked metrics and caused the wc-reader to go out of memory when evaluating them
```
OlleLarsson, line 27:
```suggestion
- Disabled prometheus rules for the kube-apiserver that lacked metrics and caused the wc-reader to go out of memory when trying to evaluate them
```

OlleLarsson (approved):

Xartos (approved):

PR #876, opened 2022-03-04
apps: fix the predict linear alert not beeing disabled

crssnd:
**What this PR does / why we need it**: if we disable the predictlinear alerts in the config that will not be applied to prometheus-alerts-wc. this pr fixes that

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.


helmfile/values/prometheus-alerts-sc.yaml.gotmpl
```diff
@@ -12,8 +12,6 @@ defaultRules:
     {{- if not .Values.prometheus.predictLinearLimit }}
     predictLinear: false
     {{- end }}
-    thanosMonitor: {{ .Values.thanos.metrics.enabled }}
```
robinAwallace, line 15: Should this be removed?

helmfile/values/prometheus-user-alerts-wc.yaml.gotmpl
```diff
@@ -19,5 +19,8 @@ defaultRules:
     # Rook is handeled by the cluster operators. Users would normally not care
     # about these alerts, but we have no other way of gathering them currently.
     rookMonitor: {{ .Values.monitoring.rook.enabled }}
+    {{- if not .Values.prometheus.predictLinearLimit }}
```
robinAwallace, line 22: Feels a bit strange that we use predictLinearLimit to describe the limit and disabling it. We usually do not do that.

crssnd: yes it is, I will see if I can add the enabled flag instead.

crssnd: let me know if it looks better now

aarnq: I think it looks good, but we need a migration script for the change.

crssnd: yes I will create that. I wanted some feedback before creating the migration script

crssnd: added the migration script

robinAwallace (approved):

aarnq (approved):

PR #935, opened 2022-04-08
bin: added support for kubecolor

pettersv:
I wanted kubecolor to work with ck8s ops and using an alias is not enough as bash does not expand aliases. 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/bin/ck8s
+++ b/bin/ck8s
@@ -21,6 +21,7 @@ usage() {
     # TODO: We might want to make this command less visible once we have proper
     #       support for OIDC logins.
     echo "  ops kubectl <wc|sc>                               run kubectl as cluster admin" 1>&2
+    echo "  ops kubecolor <wc|sc>                             run kubecolor as cluster admin" 1>&2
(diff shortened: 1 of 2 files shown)


Xartos (commented):
It looks good to me. Just a question, are there any reason to why we want this instead of just telling people to use `kubecolor` as an alias to `kubectl`?

pettersv:
> It looks good to me. Just a question, are there any reason to why we want this instead of just telling people to use `kubecolor` as an alias to `kubectl`?

It doesn't work. Bash wont expand aliases and even if the "shopt -s expand_aliases" thing is used, you would then need to specify the alias in the script.

Xartos:
> It doesn't work. Bash wont expand aliases and even if the "shopt -s expand_aliases" thing is used, you would then need to specify the alias in the script.

I see, yea then it makes sense why we would like this

Xartos (approved):

Issue #809, opened 2022-02-04
Fluentd json logging


OlleLarsson:
**Describe the bug**
I'm not sure if this is a bug or something else, but it seems like fluentd does not only log in json format.

We have this in the systems config
```
❯ k -n fluentd get cm fluentd-fluentd-elasticsearch -oyaml|grep -A 6  system.conf
  system.conf: |-
    <system>
      root_dir /tmp/fluentd-buffers/
      <log>
        format json
      </log>
    </system>
```

But you can see non json logs when looking

```
{"time":"2022-01-31 14:36:03 +0000","level":"warn","message":"dump an error event: error_class=Fluent::Plugin::ElasticsearchErrorHandler::ElasticsearchError error=\"400 - Rejected by Elasticsearch [error type]: mapper_parsing_exception [reason]: 'failed to parse field [keyValues.newToDate] of type [date] in document with id 'ZxyQsH4BKvTs0LJccfOE'. ...}
...
2022-01-31 14:36:03.688200120 +0000 fluent.warn: {"error":"#<Fluent::Plugin::ElasticsearchErrorHandler::ElasticsearchError: 400 - Rejected by Elasticsearch [error type]: mapper_parsing_exception [reason]: 'failed to parse field [keyValues.newToDate] of type [date] in document with id 'ZxyQsH4BKvTs0LJccfOE'. ...}
```

**To Reproduce**
Install in any cluster.
Wait for some  error or warning.
Observe that there seems to be non json logs.

**Expected behavior**

**Screenshots**

**Version (add all relevant versions):**
 - Compliant kubernetes apps version 0.18.2

**Additional context**
Again, not sure if this is working as expected or if there is something strange happening.

robinAwallace:
Closing this, look at the discussion in this issue: https://github.com/elastisys/compliantkubernetes-apps/pull/999.

PR #999, opened 2022-05-16
apps: fluentd ignore own logs and set type of keys

### LINKED ISSUE #544: Elasticsearch - bad mapping of date field
**Describe the bug**
In a new index in elasticsearch the `@timestamp` field was mapped to `long` instead of `date` as it usually is. I'm not sure why this happened.

This caused errors when trying to look at the logs in Kibana. They looked something like this (the original message is lost):
```
"caused_by": {
            "type": "null_pointer_exception",
            "reason": "Cannot invoke \"org.elasticsearch.search.aggregations.InternalAggregations.getSerializedSize()\" because \"reducePhase.aggregations\" is null"
          }
```

Fluentd also stopped sending logs and said that elasticsearch was rejecting the logs (I didn't find any useful reason in the logs for why they were rejected).

**To Reproduce**
Not sure. Though you can probably force elasticsearch to map `@timestamp` to a bad type, just to see the errors.

**Expected behavior**
The field `@timestamp` should always be mapped to  `date`.
Perhaps we can force elasticsearch to map it correctly instead of relying on automatic mapping.

**Screenshots**
timestamp-long

 - Compliant kubernetes apps version v0.16.0

**Additional context**
I managed to manually fix the issue by doing a rollover on the index. After that logs started to arrive in the new index and the field was mapped correctly. However I was then still left with one index that had the wrong mapping. As of writing this I have not reindexed that, but it should work in order to change the field mapping.

robinAwallace:

Fixed issue when fluentd start adding `\` to a log after failing to push to opensearch.
Also fixed two mapping issues. 

**Which issue this PR fixes**: fixes #544 

image

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -16,6 +16,9 @@
 - Fixed deploy script with correct path to `extra-user-view` manifest.
 - Fixed issue when `keys` in config had `'.'` in its name and was being moved from `sc/wc` to `common` configs.
 - Fixed broken index per namespace feature for logging. The version of `elasticsearch_dynamic` plugin in Fluentd no longer supports OpenSearch. Now the OpenSearch output plugin is used for the feature thanks to the usage of placeholders.
+- Fixed conflicting type `ts` in opensearch, where multiple services log `ts` as different types.
+- Fixed conflicting type `@timestamp`, should always be `date` in opensearch.
+- Fluentd no longer tails its own container log. Fixes the issue when Fluentd failed to push to OpenSearch and started filling up its logs with `\`. Because recursive logging of its own errors to OpenSearch which kept failing and for each fail adding more `\`.
 
 ### Added
 
--- a/helmfile/charts/opensearch/configurer/files/index-templates/kubernetes.template.json
+++ b/helmfile/charts/opensearch/configurer/files/index-templates/kubernetes.template.json
@@ -7,6 +7,16 @@
       "index.refresh_interval": "10s",
       "plugins.index_state_management.rollover_alias": "kubernetes",
       "mapping.total_fields.limit": "1000"
+    },
+    "mappings" : {
+      "properties" : {
+        "@timestamp" : {
+          "type" : "date"
+        },
+        "ts" : {
+          "type" : "text"
+        }
+      }
     }
   }
 }
--- a/helmfile/values/fluentd-configmap.yaml.gotmpl
+++ b/helmfile/values/fluentd-configmap.yaml.gotmpl
@@ -183,6 +183,7 @@ forwarder:
         @id fluentd-containers.log
         @type tail
         path /var/log/containers/*.log
+        exclude_path ["/var/log/containers/fluentd*"]
         pos_file /var/log/containers.log.pos
         pos_file_compaction_interval 72h
         tag raw.kubernetes.*
--- a/helmfile/values/fluentd-user.yaml.gotmpl
+++ b/helmfile/values/fluentd-user.yaml.gotmpl
@@ -96,6 +96,7 @@ extraConfigMaps:
       @id fluentd-containers.log
       @type tail
       path /var/log/containers/*.log
+      exclude_path ["/var/log/containers/fluentd*"]
       pos_file /var/log/containers.log.pos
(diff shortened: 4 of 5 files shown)


OlleLarsson:
Isn't ignoring fluentd's own logs already done?
```
root@fluentd-fluentd-elasticsearch-dd4wk:/# cat etc/fluent/fluent.conf 
# This is the root config file, which only includes components of the actual configuration

# Do not collect fluentd's own logs to avoid infinite loops.
<label @FLUENT_LOG>
  <match fluent.*>
    @type stdout
  </match>
</label>

@include /etc/fluent/config.d/*.conf
```

robinAwallace:
> Isn't ignoring fluentd's own logs already done?
> 
> ```
> root@fluentd-fluentd-elasticsearch-dd4wk:/# cat etc/fluent/fluent.conf 
> # This is the root config file, which only includes components of the actual configuration
> 
> # Do not collect fluentd's own logs to avoid infinite loops.
> <label @FLUENT_LOG>
>   <match fluent.*>
>     @type stdout
>   </match>
> </label>
> 
> @include /etc/fluent/config.d/*.conf
> ```
> 
> Or nvm, I see the issue now. So there is now way of getting logs from fluentd to elasticsearch withough getting circular logs when certain errors/situations occur?

I do not see that we have labeled any source so that match will never happen. But i may be missing something. 

For some certain errors it will become a circular logs. Where the error is propagated to fluentds error and each time failing it is adding more `\`.

OlleLarsson:
> I do not see that we have labeled any source so that match will never happen. But i may be missing something.

https://docs.fluentd.org/deployment/logging#capture-fluentd-logs ?

robinAwallace:
> https://docs.fluentd.org/deployment/logging#capture-fluentd-logs ?

Interesting. Im wondering if the container source still picks up the fluentd logs. As the issue stopped when i excluded the container logs starting with fluentd.

OlleLarsson:
> > https://docs.fluentd.org/deployment/logging#capture-fluentd-logs ?
> 
> Interesting. Im wondering if the container source still picks up the fluentd logs. As the issue stopped when i excluded the container logs starting with fluentd.

Yeah, I mean putting fluentd's logs on stdout just means that they will end up in the container log file which is then picked up by fluentd (until you now excluded it).

robinAwallace:
I think that explains this issues aswell: https://github.com/elastisys/compliantkubernetes-apps/issues/809. So i think we can close it.

helmfile/charts/opensearch/configurer/files/index-templates/kubernetes.template.json
```diff
@@ -7,6 +7,16 @@
...
       "plugins.index_state_management.rollover_alias": "kubernetes",
       "mapping.total_fields.limit": "1000"
+    },
+    "mappings" : {
+      "properties" : {
+        "ts" : {
+          "type" : "text"
+        },
+        "@timestamp" : {
+          "type" : "date"
```
OlleLarsson, line 17: NIT, swap places please :)
```suggestion
      "properties" : {
        "ts" : {
          "type" : "text"
        },
        "@timestamp" : {
          "type" : "date"
```

OlleLarsson (approved):

WIP-CHANGELOG.md
```diff
@@ -9,12 +9,17 @@
 - Patched Falco rules for  `write_etc_common` , `Launch Package Management Process in Container` , `falco_privileged_images` & `falco_sensitive_mount_containers`. Will be removed if upstream Falco Chart accepts these.
 - Improved error handling for applying manifests in wc deploy script
 - `kube-prometheus-stack-alertmanager` is configured to have 2 replicas to increase stability and make it highly available.
+- Fluentd ignore its own logs. Fixes the issue when Fluentd failed to push to OpenSearch and
```
OlleLarsson, line 12:
```suggestion
- Fluentd no longer tails its own container log. Fixes the issue when Fluentd failed to push to OpenSearch and
```

robinelastisys (approved):

PR #1012, opened 2022-05-25
apps sc: Add option to encrypt with rclone sync

aarnq:
Allows the contents of a bucket to be encrypted before it's sent when using rclone sync.

fixes #

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/helmfile/charts/rclone-sync/files/rclone.conf
+++ b/helmfile/charts/rclone-sync/files/rclone.conf
@@ -19,3 +19,21 @@ endpoint = {{ .Values.config.destination.s3.regionEndpoint }}
 force_path_style = {{ .Values.config.destination.s3.forcePathStyle }}
 v2_auth = {{.Values.config.destination.s3.v2Auth }}
 {{- end }}
+
+{{- if .Values.config.encrypt.enabled }}
+{{- range .Values.buckets }}
+
+[{{ $.Values.config.encrypt.name }}-{{ .destination }}]
+type = crypt
+remote = {{ $.Values.config.destination.name }}:{{ .destination }}
+password = {{ $.Values.config.encrypt.password }}
+password2 = {{ $.Values.config.encrypt.salt }}
+{{- if not $.Values.config.encrypt.directoryNames }}
+directory_name_encryption = false
+{{- end }}
+{{- if not $.Values.config.encrypt.fileNames }}
+filename_encryption = off
+{{- end }}
+
+{{- end }}
+{{- end }}
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,6 +1,16 @@
 ### Release notes
 
+### Updated
+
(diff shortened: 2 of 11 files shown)


helmfile/charts/rclone-sync/files/rclone.conf
```diff
@@ -19,3 +19,29 @@ endpoint = {{ .Values.config.destination.s3.regionEndpoint }}
 force_path_style = {{ .Values.config.destination.s3.forcePathStyle }}
 v2_auth = {{.Values.config.destination.s3.v2Auth }}
 {{- end }}
+
+{{- if .Values.config.encrypt.enabled }}
+{{- range .Values.buckets }}
+
+{{ if hasKey . "destination" -}}
```
Xartos, line 26: Doesn't they always have destination?

Xartos: right, source and destination is the bucket name :facepalm: I was somehow thinking it was the endpoint to the other S3 service

Xartos (approved):

aarnq:
> If using this should we update our docs on how to do a disaster recovery? I mean instructions on how to restore from the 2nd site.

Would we then want to add steps to restore services directly from second site, or to restore a bucket from second site, or both?
For encrypted ones the second option is required.

OlleLarsson:
> > If using this should we update our docs on how to do a disaster recovery? I mean instructions on how to restore from the 2nd site.
> 
> Would we then want to add steps to restore services directly from second site, or to restore a bucket from second site, or both? For encrypted ones the second option is required.

I would say we want to be able to restore directly from the 2nd site.

aarnq:
I'll update the public docs as well with how to restore from off-site backup.

OlleLarsson (approved):

aarnq:
Updated docs: https://github.com/elastisys/compliantkubernetes/pull/358

I had to update the Velero plugin for AWS to be able to get it to work in a good way. Though we should've update it along with Velero.

Xartos (approved):

OlleLarsson (approved):
Do I understand correctly that is what needs to happen if we have the following situation:
1. S3 site x, unencrypted buckets.
2. S3 site y, encrypted buckets

If S3 site x blows up and we need let's say restore opensearch from a snapshot, we would create a new bucket in site y and follow the instructions you've writte in order to get an unencrypted bucket that we can then restore from?

aarnq:
> If S3 site x blows up and we need let's say restore opensearch from a snapshot, we would create a new bucket in site y and follow the instructions you've writte in order to get an unencrypted bucket that we can then restore from?

Yes, through the instructions doesn't capture this situation so I can add some additional documentation on how to configure the source and destination manually if needed. Would also be a good way to be able to recover an unencrypted one if required.

PR #1014, opened 2022-05-25
apps sc: Split dashboard configmaplist into separate configmaps

aarnq:
Makes each dashboard a separate configmap instead of a massive configmaplist that sometimes causes issues in helm. 

fixes #

I had strange issues when running dry-run on the service cluster and narrowed it down to the `grafana-ops`, where the command would fail and close the shell session. And it only happened when I tried to enable rclone-sync which adds changes into the backup dashboard. But I knew it could render the chart so the problem was when it tried to run diff or apply.

So I look in to it and saw that it rendered all dashboards into one massive resource, which I think is to large for helm to handle properly in all situations.

This change has the added benefit of showing us changes in individual dashboards rather than print out a massive log each time.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.


robinAwallace (approved):

OlleLarsson (approved):
LGTM, I take it there's no migration needed for this change

aarnq:
> LGTM, I take it there's no migration needed for this change

Yeah, no steps needed.

PR #1044, opened 2022-06-09
apps sc: Add securityadmin job and alerting roles

### LINKED ISSUE #804: Opensearch - automatic reloading of securityconfig
Modifying the [securityconfig](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/opensearch/common.yaml.gotmpl#L126) updates the secret but the pods (masters) are not recreated (double check this please).
Furthermore, updating the security config needs to be done by running the  securityadmin.sh script after the initial bootstrap.

It would be nice to have a feature that you can enable that makes it so that the securityconfig is remounted and reloaded using the securityadmin.sh script.

I propose that we use the same approach as [IBM](https://www.ibm.com/docs/en/cloud-paks/cp-biz-automation/19.0.x?topic=distro-maintaining-users-roles-permissions-in-open).
We could have the configurer or some other job running the opensearch image and run the securityconfig from there.

Not having the securityconfig be automatically reloaded.

**Additional context**

**Definition of done:**'
It is possible to update the securityconfig and upon apply it is automatically reloaded.

aarnq:
To automatically reload security config in OpenSearch.
To use the "predefined" alerting roles.

Fixes #804 

I figured it was best to set it up as a separate job since it would allow a simple way to disable it completely, and you also get a release you can sync to run it manually.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -15,6 +15,7 @@
 
 - Bump falco-exporter chart to v0.8.0.
 - Users are now not forced to use proxy for connecting to alertmanager but can use port-forward as well.
+- The OpenSearch security config will now be managed completely by securityadmin
 
 ### Fixed
 
@@ -26,6 +27,10 @@
 
 - Add option to encrypt off-site buckets replicated with rclone sync
 - Added metrics for field mappings and an alert that will throw an error if the fields get close to the max limit.
+- Add support for automatic reloading of the security config for OpenSearch
+  - **Warning**: When this runs the security plugin settings will be reset. All users, roles, and role mappings created via the API will be removed, so create a backup or be prepared to recreate the resources.
+  - The securityadmin can be disabled to protect manually created resources, but it will prevent the OpenSearch cluster to initialize the security plugin when the cluster is forming.
+- Add missing roles for alerting in OpenSearch
 
 ### Removed
 - wcReader mentions from all configs files
--- a/helmfile/50-applications.yaml
+++ b/helmfile/50-applications.yaml
@@ -268,6 +268,7 @@ releases:
   chart: ./upstream/opensearch
   version: 1.7.1
   installed: true
+  wait: true
   needs:
   {{- if .Values.opensearch.sso.enabled }}
   - dex/dex
@@ -313,6 +314,7 @@ releases:
   chart: ./upstream/opensearch-dashboards
   version: 1.1.2
   installed: true
+  wait: true
   needs:
   - opensearch-system/opensearch-master
   values:
@@ -328,7 +330,7 @@ releases:
   installed: {{ .Values.opensearch.snapshot.enabled }}
   missingFileHandler: Error
   needs:
-  - opensearch-system/opensearch-master
+  - opensearch-system/opensearch-configurer
   values:
   - values/opensearch/backup.yaml.gotmpl
 
@@ -342,7 +344,7 @@ releases:
   installed: {{ .Values.opensearch.snapshot.enabled }}
   missingFileHandler: Error
   needs:
-  - opensearch-system/opensearch-master
+  - opensearch-system/opensearch-configurer
   values:
   - values/opensearch/slm.yaml.gotmpl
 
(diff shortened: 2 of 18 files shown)


OlleLarsson (commented):
Q: Is it possible to have [this](https://github.com/opensearch-project/security/blob/main/securityconfig/opensearch.yml.example#L160) set to false and have the new securityadmin job as the only user of the securityadmin script?

aarnq:
> Q: Is it possible to have [this](https://github.com/opensearch-project/security/blob/main/securityconfig/opensearch.yml.example#L160) set to false and have the new securityadmin job as the only user of the securityadmin script?

Yes, so we could do that when it's enabled.

OlleLarsson:
> > Q: Is it possible to have [this](https://github.com/opensearch-project/security/blob/main/securityconfig/opensearch.yml.example#L160) set to false and have the new securityadmin job as the only user of the securityadmin script?
> 
> Yes, so we could do that when it's enabled.

I think I might prefer that approach, otherwise you might want to consider adding the checksum of the security secret secret to the opensearch statefulsets.

OlleLarsson:
I tested upgrading on a single node cluster and I got this message
```
❯ k -n opensearch-system logs opensearch-securityadmin-q6d76 -f
Security Admin v7
Will connect to opensearch-master:9300 ... done
/usr/share/opensearch/plugins/opensearch-security/tools/securityadmin.sh: line 25:    14 Killed                  "$BIN_PATH" $JAVA_OPTS -Dorg.apache.logging.log4j.simplelog.StatusLogger.level=OFF -cp "$DIR/../*:$DIR/../../../lib/*:$DIR/../deps/*" org.opensearch.security.tools.SecurityAdmin "$@" 2> /dev/null
```

helmfile/values/opensearch/securityadmin.yaml.gotmpl
```diff
@@ -0,0 +1,173 @@
...
+  whitelist.yml: |-
+    _meta:
+      type: "whitelist"
+      config_version: 2
+    config:
+      enabled: false
+
+activeDeadlineSeconds: 1200
+
+resources:
```
OlleLarsson, line 167: Maybe expose these values

OlleLarsson (commented):
I tried a fresh install and it succeded. I had to increase the memory limit at it seemed to get OOMKilled
```
opensearch-securityadmin-drn4m          0/1     OOMKilled   3          3m6s
```

OlleLarsson:
I think you can add `wait: true` for the opensearch-master release as the securityadmin will then start a bit later and it might reduce the number of crashes.

helmfile/values/opensearch/securityadmin.yaml.gotmpl
aarnq: Yeah, I missed to test the resources :sweat_smile:, but now they're exposed and have bit more sane defaults.

helmfile/50-applications.yaml
```diff
@@ -369,9 +383,9 @@ releases:
   version: 0.1.0
   installed: true
   needs:
-  - opensearch-system/opensearch-master
-  - opensearch-system/opensearch-dashboards
+  - opensearch-system/opensearch-securityadmin
   values:
```
viktor-f, line 387: Should we add opensearch dashboard here as well? Though I know we have some check in the configurer script as well.

aarnq: Right, I got to eager while cleaning up.

WIP-CHANGELOG.md
```diff
@@ -26,6 +27,10 @@
 
 - Add option to encrypt off-site buckets replicated with rclone sync
 - Added metrics for field mappings and an alert that will throw an error if the fields get close to the max limit.
+- Add support for automatic reloading of the security config for OpenSearch
+  - **Warning**: When this runs it will replace all created users, roles, and role mappings, so create a backup or be prepared to recreate the manually created users.
```
OlleLarsson, line 31:
```suggestion
  - **Warning**: When this runs all users, roles, and role mappings created via the API will be removed, so create a backup or be prepared to recreate the manually created users.
```

OlleLarsson (approved):

viktor-f (approved):

PR #1058, opened 2022-06-21
bin: add KUBECONFIG override, can use unencrypted kubeconfigs

viktor-f:
The KUBECONFIG env var will override default kubeconfigs
Kubeconfigs can be encrypted or unencrypted

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


bin/kubeconfig.bash
```diff
@@ -7,37 +7,16 @@ here="$(dirname "$(readlink -f "$0")")"
 source "${here}/common.bash"
 
 usage() {
-    echo "Usage: kubeconfig <user|admin <wc|sc> [cluster_name]>" >&2
+    echo "Usage: kubeconfig user" >&2
```
Xartos, line 10: Do we need the `user` argument since it's always user?

viktor-f: No you're right, we don't really need it.
Would it be good to keep it to be more backwards compatible or is that just unnecessary?

bin/common.bash
```diff
@@ -61,16 +61,20 @@ config["override_common"]="${CK8S_CONFIG_PATH}/common-config.yaml"
...
-secrets["kube_config_sc"]="${state_path}/kube_config_sc.yaml"
-secrets["kube_config_wc"]="${state_path}/kube_config_wc.yaml"
-
 log_info() {
     echo -e "[\e[34mck8s\e[0m] ${*}" 1>&2
 }
 
+log_info_no_newline() {
+    echo -e -n "[\e[34mck8s\e[0m] ${*}" 1>&2
+}
```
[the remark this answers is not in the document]
viktor-f: Since it's so minimal I'm not sure it matters, but I can change it if you want.

bin/kubeconfig.bash
Xartos: If we can make it so it doesn't error out if you have arguments, then it will be fully compatible with the old way as well.

bin/common.bash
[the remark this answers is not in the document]
Xartos: I think it would look nicer to have it like that. But I agree that it isn't that big of a deal

bin/kubeconfig.bash
viktor-f: Good point, then I'll change that.

Xartos (approved):

viktor-f:
Would we want a config option here that sets the default kubeconfig name that apps should search for if KUBECONFIG is not set? it would default to `.state/kube_config_sc.yaml` and `.state/kube_config_wc.yaml` in `defaults/sc-config.yaml` and `defaults/sc-config.yaml`, but then you could override it if you want other filenames. This would also open up the option of having kubeconfigs in other places than `.state`.

Xartos:
> Would we want a config option here that sets the default kubeconfig name that apps should search for if KUBECONFIG is not set? it would default to `.state/kube_config_sc.yaml` and `.state/kube_config_wc.yaml` in `defaults/sc-config.yaml` and `defaults/sc-config.yaml`, but then you could override it if you want other filenames. This would also open up the option of having kubeconfigs in other places than `.state`.

Interresting, that would decouple ck8s-kubespray and apps even more.
However I don't see any real benefit of it other than that. But it shouldn't break anything, only open up for more possibilities so you have a thumbs up from my side

viktor-f:
> Interresting, that would decouple ck8s-kubespray and apps even more. However I don't see any real benefit of it other than that. But it shouldn't break anything, only open up for more possibilities so you have a thumbs up from my side +1

After looking a bit at it it was slightly more complicated/ugly than I wanted it to be. So I'm skipping it for now and creating an issue for it instead. Then we can merge this PR and include it in the 0.23 release.

viktor-f:
Follow-up issue: https://github.com/elastisys/compliantkubernetes-apps/issues/1062

davidumea (approved):

aarnq (approved):

PR #1077, opened 2022-07-07
apps: re-work predict linear alerts

### LINKED ISSUE #1224: Upgrade gatekeeper to version v3.9.0
**What Chart should be upgraded?**

upstream/gatekeeper [chart](https://github.com/open-policy-agent/gatekeeper/blob/v3.9.0/charts/gatekeeper/Chart.yaml)

**Why should this Chart be upgraded?**

- [ ] We need to perform a security upgrade.
- [ ] We need it for a new feature: *which feature (link to blocked issue)*.
- [x] We need to keep up to prevent future issues.
- [ ] Other: *Write the reason here*

**Acceptance criteria**

- I checked the migration of the new Chart:
- [ ] I tested the functionality provided by the new Chart (e.g., Auth flow, Grafana dashboards, etc.)

crssnd:
**What this PR does / why we need it**: alert if resources requests go over a certain limit for masters, workers or elastisys nodes

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1224

**Screenshot**
regular cluster with default resources:
image

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.


config/config/common-config.yaml
```diff
@@ -158,6 +158,12 @@ prometheus:
   tolerations: []
   affinity: {}
   nodeSelector: {}
+  ## Predictive alert if the resource usage will hit the set percentage in 3 days
+  predictLinear:
+    enabled: true
+    disklimit: 66
+    usagelimit: 66
+    requestslimit: 85
```
viktor-f, line 166: Did all of these work (not alert) on our regular cluster sizing?

crssnd: I will double check and update them after I remove the alerts for master

crssnd: done + added a screenshot

viktor-f (approved):
LGTM

raviranjanelastisys (approved):
LGTM !!

viktor-f:
> It might not be the best that the filter on node types are hardcoded, but I guess it is a good start to be able to actually use these alerts. Perhaps later we could make it configurable with a list where one would specify node patterns and set limits for each of these patterns.

Yes that would be really nice. You don't have to implement that in this PR, but please write an issue for it.

raviranjanelastisys (approved):
LGTM !!!

config/config/common-config.yaml
```diff
@@ -158,6 +158,19 @@ prometheus:
...
+    disklimit: 66
+    usagelimit: 95
+    ## for each cpu and memory add the node pattern for which you want to create an alert and the limit value
+    requestlimit:
+      cpu:
+        worker: 80
+        elastisys: 80
+      memory:
+        worker: 80
+        elastisys: 80
```
viktor-f, line 173: You could move `elastisys` to wc.

aarnq (approved):
Yes, this looks really good.

config/config/common-config.yaml
viktor-f, line 173: Could there be some issues if someone wanted a more advanced pattern here? Since the pattern is also a yaml key I think it might have some restrictions.

Would this make more sense? (with appropriate change in the alert) To make it more free for the user to write whaterver they want.
```suggestion
    requestlimit:
      cpu:
        - pattern: ".*-worker-.*"
          limit: 80
      memory:
        - pattern: ".*-worker-.*"
          limit: 80
```

viktor-f (commented):
Nice, yes this is what I wanted.

config/config/common-config.yaml
crssnd: if I move it I think I will exclude it from our alerts

viktor-f: Right, because the alerts are evaluated in sc. Didn't think of that.

viktor-f: That looks good

viktor-f:
I have one more nit-pick now.
Since only two of the alerts actually use "predict linear", should we maybe rename this to something else?

viktor-f (approved):
LGTM

PR #1109, opened 2022-08-11
apps sc: update welcoming dashboard

### LINKED ISSUE #1068: [3] Edit welcoming dashboards in grafana and opensearch.

The welcoming dashboards in opensearch and grafana contain information that should not be in the opensource project.

Screenshot from 2022-07-01 08-54-09

Move this part of the welcoming dashboards in opensearch and grafana to config in ck8s-ops

Remove the part from the dashboards all together

**Additional context**

none

**Definition of done:**

The solution should be scaleable, meaning that we shouldn't need to add this to every environment.

davidumea:

Removed the "support" part of the welcoming dashboards from opensource project and added it as an option in the cluster configs.

Wasn't able to find a solution that didn't involve adding the config to customer envs, but I made a simple optional script to ease the "migration".

**Which issue this PR fixes** : fixes #1068 

Added a migration script, but is not required.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [X] I upgraded no Chart.

### DIFF
--- a/helmfile/charts/grafana-ops/files/welcome.md
+++ b/helmfile/charts/grafana-ops/files/welcome.md
@@ -24,10 +24,9 @@ In case you get lost, don't forget to check out the [public docs](https://elasti
 
 As a user of Compliant Kubernetes you can request to see your Cluster configuration (without secrets) by asking your administrator.
 
-## Support
-
-Did not find the answers you were looking for in the docs? Please contact us via [Jira Service Desk](https://elastisys.atlassian.net/servicedesk/customer/portals)!
-
-Phone: 010-405 42 12
+{{ if .Values.dashboard.extraTextGrafana }}
+## {{ .Values.dashboard.extraTextGrafana }}
+{{ else }}
+{{ end }}
 
 [//]: # (If you update this file, remember to also edit compliantkubernetes-apps/helmfile/charts/opensearch/configurer/files/dashboards-resources/welcome.md)
--- a/helmfile/charts/opensearch/configurer/values.yaml
+++ b/helmfile/charts/opensearch/configurer/values.yaml
@@ -3,6 +3,7 @@ image: elastisys/curl-jq:latest
 
 dashboard:
   ck8sVersion: ""
+  extraTextOpensearch: ""
 
 baseDomain: ""
 
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,4 +1,5 @@
 ### Release notes
+
 - In 1.7 the cert-manager API versions v1alpha2, v1alpha3, and v1beta1, have been removed from the custom resource definitions (CRDs).
 - In 1.8 the cert-manager will validate the spec.privateKey.rotationPolicy on Certificate resources. Valid options are Never and Always.
 - Automated CIS tests are preformed on each node using kube-bench
@@ -6,12 +7,15 @@
 ### Updated
 
 ### Changed
+
 - The ThanosReceiveHttpRequestErrorRateHigh, OpenSearchTooFewNodesRunning, and MetricsFromScClusterIsMissing prometheus-alerts were optimized to not be triggered by kured and to reduce other unwanted behaviors.
+- Support section of the grafana/opensearch welcome dashboards is removed by default, you can now add custom text at the end of the dashboards via the sc-config.
 
 ### Fixed
 - Fixed so grafana can show data from thanos that's older than 30 days (downsampled data)
 
 ### Added
+
 - Option to create custom solvers for letsencrypt issuers, including a simple way to add secrets.
 - Add external redis database as option for harbor
 - a new alert `FluentdAvailableSpaceBuffer`, notifies when the fluentd buffer is filling up
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -1005,3 +1005,13 @@ prometheusBlackboxExporter:
     fluentd: true
     thanosQuery: true
     thanosReceiver: true
+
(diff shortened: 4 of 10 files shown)


helmfile/charts/grafana-ops/files/welcome.md
```diff
@@ -24,10 +24,9 @@ In case you get lost, don't forget to check out the [public docs](https://elasti
...
 
-## Support
-
-Did not find the answers you were looking for in the docs? Please contact us via [Jira Service Desk](https://elastisys.atlassian.net/servicedesk/customer/portals)!
-
-Phone: 010-405 42 12
+{{ if .Values.dashboard.extraTextGrafana }}
+## {{ .Values.dashboard.extraTextGrafana }}
+{{ else }}
+{{ end }}
```
viktor-f, line 30: Can we move the `##` into the value instead? Or why would we want them here?
```suggestion
{{ if .Values.dashboard.extraTextGrafana }}
{{ .Values.dashboard.extraTextGrafana }}
{{ else }}
{{ end }}
```

davidumea: I wasn't able to use the character `#` in when converting from the configs into the `.md` file, wasn't able to escape it either. So right now the first string (until newline) you write in the config will be a header.

viktor-f: Ok, and it didn't help to quote the string with `' '` instead of `" "`?
Otherwise I don't think I have any suggestions right now. Could you then in the comment in the config write that the first line will be a header and that you cannot add more?

davidumea: Yup the `#` just comments out the whole string and if I escape it then the `#` character just shows up as text, single quotes didnt help.

I got a suggestion from @robinAwallace  to try to append another `.md` file to the end of this but if I cant get that working I will add the extra comments that you suggested.

davidumea:
Made some changes @robinAwallace @viktor-f, the text content is no longer customizable from config but I dont think it has to be.

viktor-f (commented):
Not 100% sure I like keeping the text in this repo since it is so specific to Elastisys. But I could be ok with it for now.

helmfile/charts/opensearch/configurer/values.yaml
```diff
@@ -3,6 +3,7 @@ image: elastisys/curl-jq:latest
 
 dashboard:
   ck8sVersion: ""
+  isElastisysCustomer: ""
```
viktor-f, line 6: You seem to have missed adding this to the values file in the grafana chart.

```suggestion
dashboard:
  ck8sVersion: ""
  isElastisysCustomer: false
```

davidumea: This isn't necessary for the grafana chart, they are a bit different.

Not sure if this has to be set because it will read the value from `sc-config.yaml` if set or the default value in `defaults/sc-config.yaml` which is already set to false.

viktor-f: Yeah it's not strictly needed there. But we still try to include all possible options in those values files. So I would like you to add it there.

davidumea:
There is no way to reach files outside of the chart, the only workaround is using a symlink. This is not a good solution because you can not use environment variables in symlinks, i.e. `$CK8S_CONFIG_PATH` so the symlink would need to point to something like `../../../../ops/support.md` and that would assume that everyone has the `ops` and `apps` repo in the same folder. Could also introduce complications if we were to re-structure this repo.

I will go back to the "one-line string" in config solution.

migration/v0.24.x-v0.25.x/add-support-message.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -euo pipefail
+
+: "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
+
+sc_config="${CK8S_CONFIG_PATH}/sc-config.yaml"
+support_file="../template/common/dashboard-support-string.yaml"
```
viktor-f, line 8: Could you maybe add the file to the migration folder in this repo instead? Or do we need the file later in the other repo?

davidumea: Changed it now to grab the value from the templates via yq4 instead, what do you think?

viktor-f: I have not learned the yq4 syntax yet, but if it works then that seems good.
Could you also add in the `upgrade-apps` doc that one should install yq4 as well?

So you will move it here?

davidumea: that other file isn't necessary anymore, it just contained the yaml values for the `welcomingDashboards[]` keys which are also set in the templates. Sure I will add it as a note

viktor-f: Not sure we are talking about the same file. I'm referring to `support_file="../template/common/dashboard-support-string.yaml"` that seem to be in another repo.

davidumea: I made that file in `ck8s-ops/environments/templates/common/` but it isn't needed anymore

viktor-f: Sorry I missed that you changed which file you are looking at.
I'm still not sure that i 100% like referring to files outside this repo, but for this specific thing it might be fine.

viktor-f (approved):

crssnd (approved):

PR #1113, opened 2022-08-15
apps: tolerate control-plane taint

### LINKED ISSUE #970: Tolerate control-plane taint
For k8s 1.24 the `node-role.kubernetes.io/master:NoSchedule` taint will be replaced by `node-role.kubernetes.io/control-plane:NoSchedule`
They recommend to add this toleration in advance so there's no surprises when we upgrade to 1.24 

So lets add it to all components that have the current `master` taint

### Definition of done

* All deployments that are tolerate the `node-role.kubernetes.io/master:NoSchedule` taint should also tolerate the `node-role.kubernetes.io/control-plane:NoSchedule` taint

crssnd:
**What this PR does / why we need it**: to support the new `node-role.kubernetes.io/control-plane:NoSchedule` taint

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #970 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/config/config/wc-config.yaml
+++ b/config/config/wc-config.yaml
@@ -89,10 +89,19 @@ fluentd:
     - effect: NoSchedule
       key: node-role.kubernetes.io/master
       value: ""
+    - key: node-role.kubernetes.io/control-plane
+      effect: NoSchedule
+      value: ""
 
   ## Only run on master nodes.
-  nodeSelector:
-    node-role.kubernetes.io/master: ""
+  nodeSelector: ""
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: node-role.kubernetes.io/control-plane
+                operator: Exists
 
   resources:
     limits:
@@ -101,8 +110,6 @@ fluentd:
     requests:
       cpu: 200m
       memory: 300Mi
-
-  affinity: {}
 
   ## Extra fluentd config to mount.
   extraConfigMaps: {}
--- a/migration/v0.24.x-v0.25.x/add-control-plane-toleration.sh
+++ b/migration/v0.24.x-v0.25.x/add-control-plane-toleration.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -euo pipefail
+
+: "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
+
+common_config="${CK8S_CONFIG_PATH}/common-config.yaml"
+sc_config="${CK8S_CONFIG_PATH}/sc-config.yaml"
+wc_config="${CK8S_CONFIG_PATH}/wc-config.yaml"
+
+add_toleration() {
+    echo "Adding control-plane toleration in $i."
+    yq4 -i '((.. | select(key == "tolerations") | (select(tag == "!!seq" and contains([{"key": "'"$1"'"}]) and contains([{"key": "'"$2"'"}]) == false))) += [{"key": "'"$2"'", "operator": "Exists", "effect": "NoSchedule"}])' "$3"
+}
+
+add_affinity() {
+    echo "Replace master affinity label with control-plane in $i."
+    yq4 -i '((.. | select(key == "matchExpressions") | select(tag == "!!seq" and contains([{"key": "'"$1"'"}]))) = [{"key": "'"$2"'", "operator": "Exists"}])' "$3"
+}
+
+for i in ${sc_config} ${wc_config} ${common_config}; do
+    if [[ ! -f "$i" ]]; then
+        echo "$i does not exist, skipping."
+        exit 1
+    else
+       add_toleration 'node-role.kubernetes.io/master' 'node-role.kubernetes.io/control-plane' "$i"
+       add_affinity 'node-role.kubernetes.io/master' 'node-role.kubernetes.io/control-plane' "$i"
+    fi
+done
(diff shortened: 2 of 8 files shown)


aarnq (approved):

migration/v0.24.x-v0.25.x/add-control-plane-toleration.sh
```diff
@@ -0,0 +1,23 @@
...
+set -euo pipefail
+
+: "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
+
+common_config="${CK8S_CONFIG_PATH}/common-config.yaml"
+wc_config="${CK8S_CONFIG_PATH}/wc-config.yaml"
+
+add_toleration() {
+    yq4 -i '((.. | select(key == "tolerations") | (select(tag == "!!seq" and contains([{"key": "'"$1"'"}]) and contains([{"key": "'"$2"'"}]) == false))) += [{"key": "'"$2"'", "operator": "Exists", "effect": "NoSchedule"}])' "$3"
+}
```
viktor-f, line 12: Looks like this will add
```
  - key: node-role.kubernetes.io/control-plane
    operator: Exists
    effect: NoSchedule
```
But in the config in the other files you have skipped the `operator`. I'd prefer if you add operator to the config files above as well to make it more consistent.

migration/v0.24.x-v0.25.x/add-control-plane-toleration.sh
```diff
@@ -0,0 +1,23 @@
...
+
+for i in ${wc_config} ${common_config}
```
viktor-f, line 14: Maybe add sc-config to the loop as well? just in case there is some cluster that has added master toleration there.

viktor-f (commented):
`upgrade-apps.md` looks like it is in the root folder instead of the migration folder.

I think your script need to manage affinities as well. I know some clusters that are using affinity to control fluentd instead of the nodeselector.

config/config/wc-config.yaml
```diff
@@ -144,10 +148,14 @@ fluentd:
...
       key: node-role.kubernetes.io/master
       value: ""
+    - key: node-role.kubernetes.io/control-plane
+      effect: NoSchedule
+      value: ""
 
   ## Only run on master nodes.
   nodeSelector:
     node-role.kubernetes.io/master: ""
+    node-role.kubernetes.io/control-plane: ""
```
viktor-f, line 158: Could you change this into affinity instead? AFAIK node selector requires nodes to have all of the specified labels. So with this we require nodes to have both `master` and `control-plane` labels, otherwise it will not be scheduled there.

migration/v0.24.x-v0.25.x/add-control-plane-toleration.sh
```diff
@@ -0,0 +1,23 @@
...
+do
```
Xartos, line 15: Code convention nit

```suggestion
for i in ${wc_config} ${common_config}; do
```

config/config/wc-config.yaml
crssnd: ok 
I do not remember the reason for keeping the nodeSelector here instead of affinity

config/config/wc-config.yaml
```diff
@@ -144,10 +152,19 @@ fluentd:
...
-  nodeSelector:
-    node-role.kubernetes.io/master: ""
+  nodeSelector: ""
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: node-role.kubernetes.io/master
+                operator: Exists
```
Xartos, line 167: Do you know if adding the `control-plane` thing here will make this break? Like is the `matchExpressions` an **and** or an **or**

viktor-f: I think it would be rather safe to just have `control-plane` here, I think that label was added in all clusters starting with kuberentes v1.20
https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-master

Xartos: Sure, that makes sense to do.

But I was also curious if someone knew if it works like an and or an or.

viktor-f: Looks like the list under nodeSelectorTerms is ORed. So it should be ok as long as one of them matches.
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#nodeselector-v1-core

viktor-f: Then you can do AND logic inside matchExpressions and matchFields inside one of the nodeSelectorTerms

Xartos: Right, nice find!

crssnd: do we want the control-plane key inside the matchExpressions?

crssnd:
I think I addressed all comments. Please check again as I also updated the migration script.

viktor-f (approved):

raviranjanelastisys (approved):
LGTM !!

OlleLarsson (approved):

aarnq (approved):

Xartos (approved):

Pavan-Gunda (approved):

lucianvlad (approved):
LGTM
Please merge the commits into 1 if possible

PR #1149, opened 2022-09-02
apps sc: added network policies to harbor

### LINKED ISSUE #1130: [2] Add NetworkPolicies for harbor
This task include creating NetworkPolicies for harbor. This includes:
- harbor (the whole namespace)

This is a subtask to https://github.com/elastisys/compliantkubernetes-apps/issues/654. See that issue for a more detailed instruction, including definition of done.

robinAwallace:

Added network policies to harbor. 

1. Move harbor charts to `harbor/`
2. Divided the network policies chart into separate charts for sc/wc
3. Added `harbor-np.yaml` with all harbors network polices

**Which issue this PR fixes**: fixes #1130 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.


robinAwallace:
What do you guys think. Should network policies be enabled by default next release or should it be disabled. As it may require some manual addition to ensure that extra pods work. Specifically when using external db/redis. But default with internal db/redis it works out of the box.

helmfile/charts/networkpolicy-service/values.yaml
```diff
@@ -0,0 +1,36 @@
+harbor:
```
Xartos, line 1: Do we want to have a `additionalNetworkPolicies` so that we can add arbitrary NPs from config? Just in case we find some corner case and we can't wait until next version

Xartos:
> What do you guys think. Should network policies be enabled by default next release or should it be disabled. As it may require some manual addition to ensure that extra pods work. Specifically when using external db/redis. But default with internal db/redis it works out of the box.

I would really like to have it enabled by default. As long as it's easy to revert if we find that it doesn't work.

helmfile/charts/networkpolicy-service/values.yaml
robinAwallace: Yes i can add that. Do you want to be applied to all or do you want to have individual control?

robinAwallace:
> > What do you guys think. Should network policies be enabled by default next release or should it be disabled. As it may require some manual addition to ensure that extra pods work. Specifically when using external db/redis. But default with internal db/redis it works out of the box.
> 
> I would really like to have it enabled by default. As long as it's easy to revert if we find that it doesn't work.

It works out of the box with the default setup. But if you would for some reason talk to any helping harbor services it wont work.

robinAwallace:
@Xartos I have now added that is possible to add additional network policies.

Xartos (approved):

helmfile/charts/networkpolicy-service/values.yaml
```diff
@@ -0,0 +1,52 @@
...
+    type: internal
+    internal: []
+      # - ingressNSMatchLabels: {}
+      #   ingressNSPodMatchLabels: {}
+    external: []
+      # - egressNSMatchLabels:
+      #     kubernetes.io/metadata.name: <postgres namespace>
+      #   egressNSPodMatchLabels: {}
+      #   ports:
+      #    - 5432
```
viktor-f, line 35: Is this correct?
```suggestion
    internal: []
      # - ingressNSMatchLabels: {}
      #   ingressPodMatchLabels: {}
    external: []
      # - egressNSMatchLabels:
      #     kubernetes.io/metadata.name: <postgres namespace>
      #   egressPodMatchLabels: {}
      #   ports:
      #    - 5432
```

helmfile/charts/networkpolicy-service/values.yaml
```diff
@@ -0,0 +1,52 @@
...
+#     podSelector:
+#       matchLabels:
+#         foo: bar
+#     ingress: {}
+#     egress: {}
+harbor:
+  enable: true
+  chartmuseum:
+    httpsIPs:
+      - "0.0.0.0/0" # S3
```
viktor-f, line 21: Would it maybe be more clear to rename it to `egressIPs` or something similar? To make it clear that it belongs to egress and not ingress.
```suggestion
  chartmuseum:
    egressIPs:
      - "0.0.0.0/0" # S3
```

scripts/restore/README.md
```diff
@@ -48,6 +48,7 @@ While restoring we need to stop all harbor pods except for the database.
 Create the job and wait until it has completed:
 
 ```
+./bin/ck8s ops kubectl sc apply -n harbor -f scripts/restore/network-policies.yaml
```
viktor-f, line 51: The other scripts in the folder have harbor in the name.
```suggestion
./bin/ck8s ops kubectl sc apply -n harbor -f scripts/restore/network-policies-harbor.yaml
```

viktor-f: Also, will this work for external databases as well? Maybe this whole restore section does not work for that.

helmfile/charts/networkpolicy-service/templates/harbor-np.yaml
```diff
@@ -0,0 +1,609 @@
...
+              {{- range $key, $value := .egressNSPodMatchLabels }}
+              {{ $key | quote }}: {{ $value | quote }}
+              {{- end }}
+          {{- end }}
+      {{- end }}
+      ports:
+    {{- range $port := .ports }}
+        - port: {{ $port }}
+    {{- end }}
+  {{- end }}
```
viktor-f, line 126: I know we have not used it a lot. But is it possible to use some sort of template for all of these redis and database sections that we repeat several times in the file? So that you just have to write it once and then add a reference to it.

viktor-f (commented):
Nice work. I have not checked the whole list of networkpolicies.

helmfile/charts/networkpolicy-service/values.yaml
robinAwallace: I wanted to make a bit more clear that that pod selector is part of the ns mach label.

scripts/restore/README.md
robinAwallace: Harbor restore does not work for external db.

helmfile/charts/networkpolicy-service/templates/harbor-np.yaml
robinAwallace: That would have been nice to have done yes :slightly_smiling_face:

scripts/restore/README.md
viktor-f: Ok, then you could add a note about that if you want (even though that is slightly out of scope).

helmfile/charts/networkpolicy-service/values.yaml
viktor-f: Ok. Since they are part of the same list item I think it's probably fine without the `NS` for the pod labels. But if you think it makes it more clear then I'm ok to keep it.

robinAwallace: Yeah i think i remove it as you say they are part of the same list item.

viktor-f (approved):

PR #1150, opened 2022-09-05
apps: add option for denying network traffic by default

### LINKED ISSUE #1040: [3] Add config option to deny network traffic by default

We constantly want to improve the security posture of Compliant Kubernetes. [An idea](https://blog.digitalis.io/kubernetes-network-policies-with-calico-f037064efc4a) that we recently bumped into is to deny all network traffic by default. This would complement our [existing safeguard](https://elastisys.io/compliantkubernetes/user-guide/safeguards/enforce-networkpolicies/) to enforce network policies and ensure Application Developers explicitly configure expected ingress and egress traffic from each Pod.

Please decide on a suitable config key. Below is just and idea:

```yaml
network:
  defaultDeny: true  # default false
```

- Only keeping the safeguard

**Additional context**

Please read [this blog post](https://blog.digitalis.io/kubernetes-network-policies-with-calico-f037064efc4a) and see if there are other goodies in it.

**Definition of done:**

- I can enable deny by default in Compliant Kubernetes.
- I'd suggest really denying all network traffic by default, including DNS and Kubernetes API access. If this turns out to be a bad idea, please let us know why.

Ajarmar:
**What this PR does / why we need it**: Add option for denying network traffic by default.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1040

**Special notes for reviewer**: Note that if you apply this setting to a cluster, you need to set several `additionalEgressPolicies` / `additionalIngressPolicies` in order for the cluster to function properly. Perhaps we would like a config option to add this automatically, but I deemed it not part of this task.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.

### DIFF
--- a/helmfile/values/calico-default-deny.yaml.gotmpl
+++ b/helmfile/values/calico-default-deny.yaml.gotmpl
@@ -1,0 +1,10 @@
+namespaceSelector:
+{{- if .Values.networkPolicies.allowedNameSpaces }}
+  has(projectcalico.org/name) && projectcalico.org/name not in {"{{- join "\", \"" .Values.networkPolicies.allowedNameSpaces }}"}
+{{- else }}
+  has(projectcalico.org/name)
+{{- end }}
+
+egressPolicies: {{- toYaml .Values.networkPolicies.additionalEgressPolicies | nindent 2 }}
+
+ingressPolicies: {{- toYaml .Values.networkPolicies.additionalIngressPolicies | nindent 2 }}
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -9,4 +9,6 @@
 
 ### Added
 
+- Option to deny network traffic by default
+
 ### Removed
--- a/config/config/common-config.yaml
+++ b/config/config/common-config.yaml
@@ -596,3 +596,10 @@ kured:
   affinity: {}
   nodeSelector: {}
   dsAnnotations: {}
+
+networkPolicies:
+  # Default deny requires Calico as network plugin
+  defaultDeny: false
+  allowedNameSpaces: []
+  additionalEgressPolicies: []
+  additionalIngressPolicies: []
--- a/helmfile/50-applications.yaml
+++ b/helmfile/50-applications.yaml
@@ -234,6 +234,17 @@ releases:
   values:
     - values/falco-exporter.yaml.gotmpl
 
+# Calico default deny
+- name: calico-default-deny
+  labels:
+    app: calico-default-deny
+  chart: ./charts/calico-default-deny
+  version: 0.1.0
+  installed: {{ .Values.networkPolicies.defaultDeny }}
+  missingFileHandler: Error
+  values:
+    - values/calico-default-deny.yaml.gotmpl
+
 # Service cluster releases
 {{ if eq .Environment.Name "service_cluster" }}
 # Dex
--- a/helmfile/charts/calico-default-deny/.helmignore
+++ b/helmfile/charts/calico-default-deny/.helmignore
@@ -1,0 +1,22 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/helmfile/charts/calico-default-deny/Chart.yaml
+++ b/helmfile/charts/calico-default-deny/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for calico default deny network policies
+name: calico-default-deny
+version: 0.1.0
--- a/helmfile/charts/calico-default-deny/README.md
+++ b/helmfile/charts/calico-default-deny/README.md
@@ -1,0 +1,5 @@
+# calico-default-deny
+
+This helm chart deploys `calico-default-deny`, a global network policy that denies all traffic in a cluster by default.
+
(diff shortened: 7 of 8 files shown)


helmfile/values/calico-default-deny.yaml.gotmpl
```diff
@@ -0,0 +1,22 @@
...
+  has(projectcalico.org/name)
+{{- end }}
+
+egressPolicies:
+{{- if .Values.network.allowDns }}
+  # allow all namespaces to communicate to DNS pods
+  - action: Allow
+    protocol: UDP
+    destination:
+      selector: 'k8s-app == "kube-dns"'
```
robinAwallace, line 14: The pods should communicate to node-local-dns and it as the label `k8s-app=node-local-dns`
I guess that node-local-dns then require to talk to coredns and then coredns out from the cluster.

Xartos: I'm not sure if this is true, since the pods don't know about `node-local-dns` but it rather intercepts traffic that has the destination `kube-dns`

robinAwallace: That can be so.

config/config/sc-config.yaml
```diff
@@ -1041,3 +1041,10 @@ welcomingDashboard:
   extraTextGrafana: ""
   # For newline in Opensearch dashboard use format "\\\\n"
   extraTextOpensearch: ""
+
+network:
+  defaultDeny: false
+  allowKubeSystem: true
+  allowDns: true
+  additionalEgressPolicies: []
+  additionalIngressPolicies: []
```
viktor-f, line 1050: You could move this and the config in wc-config.yaml to common-config if you want.

Also, would it make sense to be able to allow more namespaces via values?

viktor-f (commented):
Will this global network policy merge together with regular networkpolicies? So that this denies everything, but then you can open up things using regular netowrkpolicies?

Ajarmar:
> LGTM, only thing is that maybe you should sync with @robinAwallace so you are using the same config names. see [this PR](https://github.com/elastisys/compliantkubernetes-apps/pull/1149/files)
> 
> Right now I think there will be two separate "root config keys" `network` for this one and `networkPoliciesSc` from the other PR. I think that you could use same name for both

After discussing with @robinAwallace , we decided to go with `networkPolicies`. I have changed it.

> Will this global network policy merge together with regular networkpolicies? So that this denies everything, but then you can open up things using regular netowrkpolicies?

Yes, in the examples I've seen from Calico they use Calico NetworkPolicies but it is my understanding that it works with regular Kubernetes NetworkPolicies as well.

config/config/sc-config.yaml
viktor-f: Even if we put this in common-config you can still override that for a single cluster, e.g. for sc by adding the part you want to change in your own sc-config. Moving this to common-config just shows that this is config that is available for both and that the defaults are the same.

I was thinking of a list where the one can add whatever namespace one wants.

viktor-f:
> > Will this global network policy merge together with regular networkpolicies? So that this denies everything, but then you can open up things using regular netowrkpolicies?
> 
> Yes, in the examples I've seen from Calico they use Calico NetworkPolicies but it is my understanding that it works with regular Kubernetes NetworkPolicies as well.

Have you verified this? Otherwise, please do a quick test with this global policy and some regular networkpolicy.

config/config/sc-config.yaml
viktor-f: Could you also add a comment saying that this requires one to have calico as the network plugin?

Ajarmar: I moved it to common-config, added a variable for a list of namespaces, and added a comment.

Ajarmar:
> > > Will this global network policy merge together with regular networkpolicies? So that this denies everything, but then you can open up things using regular netowrkpolicies?
> > 
> > 
> > Yes, in the examples I've seen from Calico they use Calico NetworkPolicies but it is my understanding that it works with regular Kubernetes NetworkPolicies as well.
> 
> Have you verified this? Otherwise, please do a quick test with this global policy and some regular networkpolicy.

I have tested it now, and it works.

helmfile/values/calico-default-deny.yaml.gotmpl
viktor-f: In my testing with regular networkpolicies it did not necessarily work to just select the dns pods. Me and wallace decided to just accept UDP egress traffic to port 53 in general. Did you test that this seemed to work, otherwise just opening it up more might be better.

viktor-f: Also, don't you need an ingress rule for the dns pods? So that they can accept the traffic

Xartos: From a safety perspective, I think opening it up more is a bad idea. If you can control DNS, then it's trivial to spoof the validity of certificates and thus very easy to do a MITM attack. So I would rather get this done correctly from the start tbh. although I would be okey to leave it more open if we create an issue to fix it later

Ajarmar: I took this from the [example in Calico's docs](https://projectcalico.docs.tigera.io/security/kubernetes-default-deny), I did not test it directly. I suppose the reason why it doesn't have ingress rules for the DNS pods is because it assumes you allow all traffic in kube-system.

That said, I saw that the harbor networkpolicy includes its own DNS rule. If we want to do this for all the different networkpolicies we're adding, then I suppose it would make more sense to just remove this option from the globalnetworkpolicy and handle it on a per-app basis. What do you think?

Xartos: That might actually make most sense, if we want to limit all apps as much as possible we might not even want to have dns enabled for some namespace (although I highly doubt it).

Ajarmar: I removed the option, we'll handle it on a per-app basis instead.

raviranjanelastisys (approved):

Xartos (approved):

PR #1204, opened 2022-09-27
apps sc: added index size alerts for opensearch

crssnd:
**What this PR does / why we need it**: to detect when an index size becomes too big. that might indicate an issue with the ISM

- I also added alerts for cluster status to alerts on issues that are not covered by the existing alerts
- the default index size is based on the highest value from a dev cluster multiplied by 2

index_alert

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -1,6 +1,7 @@
 ### Release notes
 
 - Several default resource requests and limits have changed. When upgrading these might need to be adjusted.
+- The settings for the new Opensearch index size alerts might need to be tweaked to better suit the environment.
 
 ### Updated
 
@@ -31,7 +32,7 @@
 - Fixed harbor network policies
 
 ### Added
-
+- Option to configure alerts for growing indices in OpenSearch
 - Option to deny network traffic by default
 - Network policies for monitoring stack (prometheus, thanos, grafana, some exporters)
 - An alert for failed evicted pods (KubeFailedEvictedPods)
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -902,6 +902,19 @@ opensearch:
     nodeSelector: {}
     tolerations: []
 
+  # Config for https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/charts/prometheus-alerts/templates/alerts/opensearch.yaml#L70:L77
+  promIndexAlerts:
+    - prefix: kubernetes-default
+      alertSizeMB: 5500
+    - prefix: kubeaudit-default
+      alertSizeMB: 5500
+    - prefix: other-default
+      alertSizeMB: 400
+    - prefix: security-auditlog
+      alertSizeMB: 16
+    - prefix: authlog-default
+      alertSizeMB: 0.2
+
   # Snapshot and snapshot lifecycle configuration
   # Requires S3 or GCS to be enabled
   snapshot:
--- a/helmfile/charts/prometheus-alerts/templates/alerts/opensearch.yaml
+++ b/helmfile/charts/prometheus-alerts/templates/alerts/opensearch.yaml
@@ -50,4 +50,30 @@ spec:
       annotations:
         description: Index {{`{{ $labels.index }}`}} is using {{`{{ $value }}`}} percent of max field limit
         summary: Index {{`{{ $labels.index }}`}} is using {{`{{ $value }}`}} percent of max field limit
+    - alert: OpensearchClusterYellow
+      expr: elasticsearch_cluster_health_status{color="yellow"} == 1
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        summary: Opensearch Cluster Yellow (instance {{`{{ $labels.instance }}`}})
+        description: Opensearch Cluster is in a Yellow status VALUE = {{`{{ $value }}`}}  LABELS = {{`{{ $labels }}`}}
+    - alert: OpensearchClusterRed
+      expr: elasticsearch_cluster_health_status{color="red"} == 1
+      for: 15m
+      labels:
+        severity: critical
+      annotations:
+        summary: Opensearch Cluster Red (instance {{`{{ $labels.instance }}`}})
+        description: Opensearch Cluster is in a Red status VALUE = {{`{{ $value }}`}}  LABELS = {{`{{ $labels }}`}}
+    {{- range $prefixes := .Values.osIndexAlerts }}
+    - alert: {{ $prefixes.prefix | title }}SizeIncreasedOverLimit
+      expr: elasticsearch_indices_store_size_bytes_primary{index=~"{{ $prefixes.prefix }}.+"} / (1024^2) > {{ $prefixes.alertSizeMB }}
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        message: Primary shard size for index {{`{{ $labels.index }}`}} has increased over the limit of {{ $prefixes.alertSizeMB }}MB current size is {{`{{ $value | printf "%.0f" }}`}}MB
+        description: If the size keeps increasing, that might indicate a problem with ISM
+    {{- end }}
 {{- end }}
--- a/helmfile/values/prometheus-alerts-sc.yaml.gotmpl
+++ b/helmfile/values/prometheus-alerts-sc.yaml.gotmpl
@@ -5,6 +5,7 @@
 # TODO: We should update the rules for things with set namespace and job name to be able to monitor instances in the workload cluster.
 
 osNodeCount: {{ add .Values.opensearch.dataNode.count .Values.opensearch.clientNode.count .Values.opensearch.masterNode.count }}
+osIndexAlerts: {{ toYaml .Values.opensearch.promIndexAlerts | nindent 2 }}
 
 defaultRules:
   alertLabels:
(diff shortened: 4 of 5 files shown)


config/config/sc-config.yaml
```diff
@@ -899,6 +899,18 @@ opensearch:
...
+    - pattern: kubeaudit-default
+      alertSizeMB: 4000
+    - pattern: authlog-default
+      alertSizeMB: 0.2
+    - pattern: other-default
+      alertSizeMB: 400
+    - pattern: security-auditlog
+      alertSizeMB: 16
+    - pattern: kubernetes-default
+      alertSizeMB: 4000
```
aarnq, line 912: Great stuff! Though I'd want to have this and perhaps `kubeaudit` set to say `5500` instead to be `500` over [the limit set as default for production environments](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/config/flavors/prod/sc-config.yaml#L41).

crssnd: thanks you. I missed that.

helmfile/charts/prometheus-alerts/templates/alerts/opensearch.yaml
```diff
@@ -50,4 +50,30 @@ spec:
...
+        summary: Opensearch Cluster Red (instance {{`{{ $labels.instance }}`}})
+        description: Opensearch Cluster is in a Red status VALUE = {{`{{ $value }}`}}  LABELS = {{`{{ $labels }}`}}
+    {{- range $patterns := .Values.osIndexAlerts }}
+    - alert: {{ $patterns.pattern | title }}SizeIncreasedOverLimit
+      expr: elasticsearch_indices_store_size_bytes_primary{index=~"{{ $patterns.pattern }}.+"} / (1024^2) > {{ $patterns.alertSizeMB }}
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        message: Index size for {{`{{ $labels.index }}`}} has increased over the limit of {{ $patterns.alertSizeMB }}MB current size is {{`{{ $value | printf "%.0f" }}`}}MB
```
OlleLarsson, line 76:
```suggestion
        message: Primary shard size for index {{`{{ $labels.index }}`}} has increased over the limit of {{ $patterns.alertSizeMB }}MB current size is {{`{{ $value | printf "%.0f" }}`}}MB
```

OlleLarsson:
I'm a bit picky with our wording.
If we want to alert on _index size_ I would say that we should look at the total store size as that includes both primary and replica shards.
Am I too picky here @aarnq 😄 ?

aarnq:
> I'm a bit picky with our wording. If we want to alert on _index size_ I would say that we should look at the total store size as that includes both primary and replica shards. Am I too picky here @aarnq smile ?

Sure, a change of wording is best then to have it clarified. I wouldn't agree that *index size* includes replicas but I get your point. And checking this alert on total index size including replicas wouldn't really be intuitive in the context of ISM behaviour.

WIP-CHANGELOG.md
```diff
@@ -8,5 +8,6 @@
 ### Fixed
 
 ### Added
+- Alert for opensearch index size increase
```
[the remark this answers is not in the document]
aarnq: Wouldn't "primary shards of indices" be more correct since each index can have multiple primaries?

helmfile/values/prometheus-alerts-sc.yaml.gotmpl
```diff
@@ -41,7 +43,7 @@ defaultRules:
 capacityManagementAlertsDiskLimit: {{ .Values.prometheus.capacityManagementAlerts.disklimit }}
 capacityManagementAlertsUsageLimit: {{ .Values.prometheus.capacityManagementAlerts.usagelimit }}
 capacityManagementAlertsRequestLimit:
-{{ toYaml .Values.prometheus.capacityManagementAlerts.requestlimit | indent 2}}
+{{ toYaml .Values.prometheus.capacityManagementAlerts.requestlimit | indent 2 }}
```
OlleLarsson, line 46: same as comment above

OlleLarsson:
> And checking this alert on total index size including replicas wouldn't really be intuitive in the context of ISM behaviour.

True

WIP-CHANGELOG.md
```diff
@@ -8,5 +8,6 @@
 ### Fixed
 
 ### Added
+- Option to configure alerts for growing primary shards in OpenSearch
```
OlleLarsson, line 11: I'm fine just writing `indices` instead of `primary shards` as not to confuse people
```suggestion
- Option to configure alerts for growing primary shards in OpenSearch
```

OlleLarsson (approved):

config/config/sc-config.yaml
```diff
@@ -899,6 +899,18 @@ opensearch:
...
+    - pattern: kubernetes-default
+      alertSizeMB: 5500
+    - pattern: kubeaudit-default
+      alertSizeMB: 5500
+    - pattern: other-default
+      alertSizeMB: 400
+    - pattern: security-auditlog
+      alertSizeMB: 16
+    - pattern: authlog-default
+      alertSizeMB: 0.2
```
robinAwallace, line 912: Any specific reason why 200KB chosen when the retention size is 1GB?
Feels like we could choose a % of the retention size instead of specifying MB here. So if you choose a new retention size the alert follows.

crssnd: I agree, but I do not think this index will every reach 1GB in size (at least for the env I checked) so maybe we should re-evaluate the retention size for security-auditlo and authlog-default
do you want to do that as part of this PR?

robinAwallace: Yes maybe worth to look at the retention size. No I would say create an new issue for it.

helmfile/charts/prometheus-alerts/templates/alerts/opensearch.yaml
```diff
@@ -50,4 +50,30 @@ spec:
...
+        description: Opensearch Cluster is in a Red status VALUE = {{`{{ $value }}`}}  LABELS = {{`{{ $labels }}`}}
+    {{- range $patterns := .Values.osIndexAlerts }}
+    - alert: {{ $patterns.pattern | title }}SizeIncreasedOverLimit
+      expr: elasticsearch_indices_store_size_bytes_primary{index=~"{{ $patterns.pattern }}.+"} / (1024^2) > {{ $patterns.alertSizeMB }}
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        message: Primary shard size for index {{`{{ $labels.index }}`}} has increased over the limit of {{ $patterns.alertSizeMB }}MB current size is {{`{{ $value | printf "%.0f" }}`}}MB
+        description: If the size keeps increasing, that might indicate a problem with the ISM
```
OlleLarsson, line 77:
```suggestion
        description: If the size keeps increasing, that might indicate a problem with ISM
```

WIP-CHANGELOG.md
```diff
@@ -31,7 +31,7 @@
 - Fixed harbor network policies
 
 ### Added
-
+- Option to configure alerts for growing indices in OpenSearch
```
OlleLarsson, line 34: Maybe make a note in the release notes on that these new option should or might be tweaked to better suit the environment

config/config/sc-config.yaml
```diff
@@ -902,6 +902,18 @@ opensearch:
     nodeSelector: {}
     tolerations: []
 
+  promIndexAlerts:
+    - pattern: kubernetes-default
```
OlleLarsson, line 906: I'm guessing that the the name _pattern_ here is a bit ambiguous to what we have used elsewhere, like in the curation retention or opensearch roles where we usually have a `*` following the index name.

crssnd: true, but I didn't want to add them here. just to trim the * in the alert.
do you have another idea?

robinAwallace (approved):

config/config/sc-config.yaml
OlleLarsson: Maybe add a comment/example saying that a wildcard is added to the end of the given string

aarnq: Or rename it to prefix perhaps, just to make it more clear?

OlleLarsson: prefix makes sense 👍🏻

crssnd: @aarnq and @OlleLarsson I hope the new name works better

WIP-CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### Release notes
 
 - Several default resource requests and limits have changed. When upgrading these might need to be adjusted.
+- The new Opensearchi, index size increased over limit, alert should be tweaked to better suit the environment.
```
OlleLarsson, line 4: Perhps
```suggestion
- The settings for the new Opensearch index size alers might need to be tweaked to better suit the environment.
```

WIP-CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### Release notes
 
 - Several default resource requests and limits have changed. When upgrading these might need to be adjusted.
+- The settings for the new Opensearch index size alers might need to be tweaked to better suit the environment.
```
aarnq, line 4:
```suggestion
- The settings for the new Opensearch index size alerts might need to be tweaked to better suit the environment.
```

aarnq (approved):

PR #1216, opened 2022-09-30
apps sc: fix harbor backup

davidumea:

Harbor backup was pointed to the wrong service name, so it could not take backups.

@OlleLarsson Also made a new tag for the `elastisys/backup-postgres` image which uses newer postgres version

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [X] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -8,7 +8,8 @@
 ### Fixed
 
 - Blackbox exporter now looks at the correct error code for the opensearch-dashboards target
-- Harbor backup is now pointed to the correct service to make backups from
+- Harbor backup is now pointed to the correct internal service to make backups from
+- Bumped backup-postgres image to use tag `1.2.0`, which includes newer versions of the postgresql client
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


robinAwallace (approved):

WIP-CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 ### Fixed
 
 - Blackbox exporter now looks at the correct error code for the opensearch-dashboards target
+- Harbor backup is now pointed to the correct service to make backups from
+- Bumped backup-postgres image to use newer postgres version
```
OlleLarsson, line 12:
```suggestion
- Harbor backup is now pointed to the correct internal service to make backups from
- Bumped backup-postgres image to use tag `1.2.0`, which includes newer versions of the postgresql client
```

OlleLarsson (commented):
We don't have a process for our internal helm charts when it comes to versioning but I would like that we bump the chart version from like 010 to 020

crssnd (approved):

PR #1220, opened 2022-10-04
apps sc: added alert for evicted pods

### LINKED ISSUE #1173: [1]Alert for kublet low resources: ephemeral-storage
**Describe the bug**

We currently have a alert for node storage filling up. But it is not fast enough and in some cases the kublet fails to create a container and the pod is stuck in the state `ContainerStatusUnknown` with the error:

```
Warning  Evicted  37m   kubelet  The node was low on resource: ephemeral-storage. Container X was using 2596Ki, which exceeds its request of 0.
```

**To Reproduce**
1. Try to schedule a pod on a node with low disk space

**Expected behavior**

Get an alert that kubelet failed to schedule a pod becuase of low ephemeral-storage.

**Screenshots**

 - Compliant kubernetes apps version [e.g. v0.8.0]

**Additional context**

crssnd:
**What this PR does / why we need it**: to add an alert to detect evicted pods

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1173 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/helmfile/charts/prometheus-alerts/templates/alerts/kubernetes-apps.yaml
+++ b/helmfile/charts/prometheus-alerts/templates/alerts/kubernetes-apps.yaml
@@ -261,4 +261,15 @@ spec:
 {{- if .Values.defaultRules.additionalRuleLabels }}
 {{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
 {{- end }}
+    - alert: KubeFailedEvictedPods
+      annotations:
+        description: '{{`{{`}} $value {{`}}`}} Failed Evicted pods in {{`{{`}} $labels.namespace {{`}}`}} namespace {{`{{`}} $labels.cluster {{`}}`}} cluster'
+        summary: Kubernetes failed evicted pods
+      expr: sum by (namespace, cluster) (kube_pod_status_phase{phase="Failed"} > 0 and on(namespace, cluster) kube_pod_status_reason{reason="Evicted"} > 0) > 0
+      for: 10m
+      labels:
+        severity: warning
+{{- if .Values.defaultRules.additionalRuleLabels }}
+{{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
+{{- end }}
 {{- end }}
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -15,6 +15,7 @@
 
 - Option to deny network traffic by default
 - Network policies for monitoring stack (prometheus, thanos, grafana, some exporters)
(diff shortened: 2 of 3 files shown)


Xartos (commented):
I'm a bit hesitant on this, do we really want alerts for eviction?

crssnd:
> I'm a bit hesitant on this, do we really want alerts for eviction?
> 
> Feels like that's something that kubernetes should be able to do and shouldn't be an issue and if it is, the pod would probably get stuck in pending state and we'd get an alert for that

yes I agree, we have other alerts that can indirectly notify us in case of eviction, so I can remove it
and alternative would be to make this alert trigger when to many pods in a cluster got evicted (e.g. > 10)

@robinAwallace what are your thoughts on this?

robinAwallace:
There was a case when pods got stuck in `evicted` and not in pending. Im guessing here that the evicted pod are removed once the new pod is healthy. But in the case when the pods got stuck in evicted was when the nodes did not have disk space. So i think  the node disk space is not checked during scheduling (as for cpu and memory) but during kubelet execution? So the environment got filled with pods stuck in `evicted`. So i would like a alert for pods stuck in `evicted`.

helmfile/charts/prometheus-alerts/templates/alerts/kubernetes-apps.yaml
```diff
@@ -260,5 +260,16 @@ spec:
...
+    - alert: KubePodEvicted
+      annotations:
+        description: '{{`{{`}} $value {{`}}`}} pods were Evicted from {{`{{`}} $labels.namespace {{`}}`}} namespace {{`{{`}} $labels.cluster {{`}}`}} cluster in the last 10 minutes.'
+        summary: Kubernetes pods got evicted.
+      expr: sum by (namespace, cluster) (kube_pod_status_reason{reason="Evicted"}) >= 1
+      for: 10m
+      labels:
+        severity: warning
+{{- if .Values.defaultRules.additionalRuleLabels }}
+{{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
```
Xartos, line 273: Isn't this a duplicate? Looks like these are present at line 261-262

crssnd: this lines are added for each alert, to add additional labels, if used
I think they are from the upstream files
do you want to remove them?

Xartos (approved):

robinAwallace (approved):

PR #1240, opened 2022-10-10
apps: backofflimit set to 0 for s3 jobs

### LINKED ISSUE #1232: [1] backoffLimit: 0 for rclone-sync jobs
To avoid, potentially incredibly costly, mistakes when failing jobs are restarting, they should not be allowed to restart if the first one fails.

backoffLimit set to 0 by default in the jobtemplate for the sync jobs.

If they fail let them hit the default back off limit (6)

**Additional context**
A job that syncs to S3 that fails can be incredibly costly as it might transfer huge amounts of data every time.

**Definition of done:**
Failing sync jobs do not ever restart by default.

robinAwallace:

Sets the backofflimit for jobs that interact with s3 to 0.

**Which issue this PR fixes**: fixes #1232 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -17,6 +17,7 @@
 - Changed some default resource requests and limits for multiple components
 - Gatekeeper audits every 10 min instead of 1 min
 - Gatekeeper audit only looks at the resource types mentioned in constraints, instead of all resource types
+- Set backoffLimit for rclone-jobs to 0.
 
 ### Fixed
 
--- a/helmfile/charts/rclone-sync/templates/cronjob.yaml
+++ b/helmfile/charts/rclone-sync/templates/cronjob.yaml
@@ -16,6 +16,7 @@ spec:
   concurrencyPolicy: Forbid
   jobTemplate:
     spec:
+      backoffLimit: {{ .Values.backoffLimit }}
       template:
         metadata:
           labels:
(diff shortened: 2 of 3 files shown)


helmfile/charts/opensearch/slm/values.yaml
```diff
@@ -23,6 +23,7 @@ snapshots:
   maxRequestSeconds: 1200
 
 restartPolicy: OnFailure
+backoffLimit: 0
```
aarnq, line 26: I don't think this makes sense for opensearch-slm and sc-logs-rentention.
Especially opensearch-slm sometimes needs retries since it relies on OpenSearch to actually carry out the clean up which may take a while depending on the number of snapshots and indices in them.

OlleLarsson: Agreeing with @aarnq

robinAwallace: Sure i can remove it from slm. But i keep it for opensearch backup.

OlleLarsson: I would still only want the backoff limit set on the rclone jobs

aarnq (approved):

OlleLarsson (approved):

OlleLarsson:
> @OlleLarsson @aarnq it is now only rclone that has backofflimit set to 0.

Awesome, thank you!

PR #1255, opened 2022-10-17
apps sc: made dex id token expiration time configurable

### LINKED ISSUE #1242: [1] Dex id token expiration time configurable?

Can we make dex id token expiration time configurable?

In the dexidp helm chart there is no option to configure how fast the id token should expire, is there a reasonable solution to go around this?

There should be a written investigation with a conclusion whether this is doable in an easy way. If doable, then another issue for a fix should be created.

davidumea:

Made dex id token expiration time configurable

Tested setting it to `1m`, then `5m` and then `24h` in my cluster and it worked as expected

**Which issue this PR fixes** : fixes #1242

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [X] I upgraded no Chart.

### DIFF
--- a/helmfile/values/dex.yaml.gotmpl
+++ b/helmfile/values/dex.yaml.gotmpl
@@ -28,6 +28,14 @@ volumes:
 {{- end }}
 
 config:
+  expiry:
+    deviceRequests: {{ .Values.dex.expiry.deviceRequests }}
+    signingKeys: {{ .Values.dex.expiry.signingKeys }}
+    idTokens: {{ .Values.dex.expiry.idToken }}
+    refreshTokens:
+      reuseInterval: {{ .Values.dex.expiry.refreshTokens.reuseInterval }}
+      validIfNotUsedFor: {{ .Values.dex.expiry.refreshTokens.validIfNotUsedFor }}
+      absoluteLifetime: {{ .Values.dex.expiry.refreshTokens.absoluteLifetime }}
   storage:
     type: kubernetes
     config:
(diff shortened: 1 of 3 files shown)


robinAwallace (approved):

helmfile/values/dex.yaml.gotmpl
```diff
@@ -28,6 +28,8 @@ volumes:
 {{- end }}
 
 config:
+  expiry:
+    idTokens: {{ .Values.dex.idTokenExpiryTime }}
```
Xartos, line 32: Should we add all the expiry times when were at it?

davidumea: Do you see any cases where the other ones could be relevant? I think some of these would just cause confusion, like `signingKeys` and `deviceRequests` https://github.com/dexidp/dex/blob/master/examples/config-dev.yaml#L78-L85

OlleLarsson: I think it is enough with just being able to configure the id token

Xartos: At least the refresh tokens and signing keys are involved in the flow. But I'd rather have them configurable, so when we get to the situation where we need to modify them we already have it instead of needing to wait for another release.

OlleLarsson: Make sense. Then please @davidumea add configuration options for those as well

Xartos (approved):

PR #1311, opened 2022-11-23
Qa 0.27

crssnd:
**What this PR does / why we need it**: fixes after the QA

**Unresolved issues:**
- there are dropped packets going to ingress-controller (both sc and wc) pods that I didn't manage to identify. Didn't notice any obvious issues caused by this.
image

- egress + ingress
image

### DIFF
--- a/helmfile/charts/networkpolicy-common/templates/falco/allow-falco.yaml
+++ b/helmfile/charts/networkpolicy-common/templates/falco/allow-falco.yaml
@@ -25,9 +25,16 @@ spec:
         {{- end }}
       ports:
         - port: {{ .Values.global.apiserver.port }}
-    - ports:
-        - port: 443
-          protocol: TCP
+    - to:
+        {{- range $ip := .Values.falco.plugins.ips }}
+        - ipBlock:
+            cidr: {{ $ip }}
+        {{- end }}
+      ports:
+        {{- range $port := .Values.falco.plugins.ports }}
+        - protocol: TCP
+          port: {{ $port }}
+        {{- end }}
     - ports:
         - port: 53
           protocol: UDP
--- a/helmfile/charts/networkpolicy-common/templates/falco/allow-falco.yaml
+++ b/helmfile/charts/networkpolicy-common/templates/falco/allow-falco.yaml
@@ -9,7 +9,7 @@ spec:
     - Egress
   podSelector:
     matchLabels:
-      app: falco
+      app.kubernetes.io/name: falco
   egress:
     - to:
       - podSelector:
@@ -25,6 +25,9 @@ spec:
         {{- end }}
       ports:
         - port: {{ .Values.global.apiserver.port }}
+    - ports:
+        - port: 443
+          protocol: TCP
     - ports:
         - port: 53
           protocol: UDP
--- a/helmfile/values/harbor.yaml.gotmpl
+++ b/helmfile/values/harbor.yaml.gotmpl
@@ -66,9 +66,9 @@ persistence:
       authurl: {{ $swift.authUrl }}
       region: {{ $swift.region }}
       container: {{ .Values.objectStorage.buckets.harbor }}
-      {{- if hasKey $swift "domainId" }}
+      {{- if getOrNil "domainId" $swift }}
       domainid: {{ $swift.domainId }}
-      {{- else if hasKey $swift "domainName" }}
+      {{- else if getOrNil "domainName" $swift }}
       domain: {{ $swift.domainName }}
       {{- end }}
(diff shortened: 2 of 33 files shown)


davidumea:
that NP issue for ingress controller doesn't seem to be an issue on safespring, this is how it looks in my cluster
image

helmfile/values/harbor.yaml.gotmpl
```diff
@@ -66,9 +66,9 @@ persistence:
       authurl: {{ $swift.authUrl }}
       region: {{ $swift.region }}
       container: {{ .Values.objectStorage.buckets.harbor }}
-      {{- if hasKey $swift "domainId" }}
+      {{- if $swift.domainId }}
       domainid: {{ $swift.domainId }}
-      {{- else if hasKey $swift "domainName" }}
+      {{- else if  $swift.domainName }}
```
Xartos, line 71: Isn't there a reason for having `hasKey`? Like if it's not set there's no error (or something like that)

helmfile/charts/networkpolicy-common/templates/falco/allow-falco.yaml
```diff
@@ -25,6 +25,9 @@ spec:
         {{- end }}
       ports:
         - port: {{ .Values.global.apiserver.port }}
+    - ports:
```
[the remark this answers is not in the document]
davidumea: IIRC I think it was trying to download some files (may have been a plugin) and I think we decided to just open the port as we couldn't be sure that the IP of the file target destination would be the same

helmfile/values/harbor.yaml.gotmpl
crssnd: maybe @aarnq can answer that
but with `haskey` it will always return true as the keys exists in the default configs with `set-me`

helmfile/charts/networkpolicy-common/templates/falco/allow-falco.yaml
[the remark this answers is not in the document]
robinAwallace: Can you add that it is possible to set IPs for port 443. But it will be `0.0.0.0/0` in common-config.yaml as you said it will be hard to know the ips

helmfile/values/harbor.yaml.gotmpl
aarnq: Hmm. The reason we can't have `hasKey` is because you should only specify either `domainId` or `domainName` and the init scripts will (for Cleura) add both into the default config so the key will always exist regardless of what you set it to.

I guess a way to have both is to use `if getOrNil $swift "domainId"`.

helmfile/charts/networkpolicy-common/templates/falco/allow-falco.yaml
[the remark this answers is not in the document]
davidumea: Pushed a fix for this, didn't add an additional note in CHANGELOG as this line was already there from a previous QA commit
- Updated the allow-falco & allow-falco-sidekick network policies to work with the new falco version

helmfile/values/networkpolicy-common.yaml.gotmpl
```diff
@@ -53,3 +53,6 @@ certManager:
     ips: {{- toYaml .Values.networkPolicies.certManager.letsencrypt.ips | nindent 6 }}
 falco:
   enabled: {{ and .Values.falco.enabled .Values.networkPolicies.falco.enabled }}
+  plugins:
+    ips: {{- toYaml .Values.networkPolicies.falco.plugins.ips | nindent 8 }}
+    ports: {{- toYaml .Values.networkPolicies.falco.plugins.ports | nindent 8 }}
```
robinAwallace, line 58: nit
```suggestion
    ips: {{- toYaml .Values.networkPolicies.falco.plugins.ips | nindent 6 }}
    ports: {{- toYaml .Values.networkPolicies.falco.plugins.ports | nindent 6 }}
```

helmfile/values/harbor.yaml.gotmpl
crssnd: should i so some tests with `getOrNil` instead?

aarnq: Yes if you have the time for it, else I can create issue to look at these things in general.

crssnd: done

Xartos (approved):

aarnq (approved):

robinAwallace (approved):

aarnq (approved):
LGTM

PR #1330, opened 2022-12-14
apps: disable rook network policies by default

crssnd:
**What this PR does / why we need it**: rook network policies should be enabled only on env that are using rook-ceph. I propose to disabled them in the default configs.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/config/config/common-config.yaml
+++ b/config/config/common-config.yaml
@@ -660,7 +660,7 @@ networkPolicies:
         - 443
 
   rookCeph:
-    enabled: true
+    enabled: false
 
   kubeSystem:
     enabled: true
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -19,8 +19,8 @@
 ### Added
 
 - Starboard resources will now be removed when running the cleanup scripts - `scripts/clean-{sc,wc}.sh`.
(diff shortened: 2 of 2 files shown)


aarnq (approved):

raviranjanelastisys (approved):

config/config/common-config.yaml
```diff
@@ -660,7 +660,7 @@ networkPolicies:
         - 443
 
   rookCeph:
-    enabled: true
+    enabled: false
```
Xartos, line 663: Would we like to add to like the init command to set this to true if it's a exoscale cluster?

crssnd: seems that with rook network policies enabled we cannot create PVCs
you can see that [here](https://github.com/elastisys/compliantkubernetes-apps/actions/runs/3702298246/jobs/6272466621)

I will add the init script update in the task opened to fix the rook network policies.

raviranjanelastisys: @crssnd Hmmm interesting , because when i tried to create manually , it worked :( !!

Xartos (approved):

Issue #1191, opened 2022-09-22
Troubleshooting scripts - Certificates and Ingresses debugging


aarnq:
**Is your feature request related to a problem? Please describe.**

Compliant Kubernetes needs more troubleshooting scripts to help operators quickly figure out faults in an environment.
We could use some more debugging checks for certificates and ingresses so we can assert that they work as they should.

**Describe the solution you'd like**

These are for when the normal checks don't cut it, so we can test with actual usage.
Check for:

**cert-manager:**
- Script to issue a certificate
- Script to force a certificate renewal
    
**ingress-nginx:**
- Perform either pings or preferably requests over time:
	- Check for dropped requests
	- Check for failed requests
	- Check for high latency
	- Check for high variability
- Check logs for all ingress-nginx pods:
    - Filter out errors and warnings
    - Identify source node or pod

And present these in some nice way.

**Describe alternatives you've considered**

Alternatives would be to check manually.

**Additional context**

Note that it could be good to have the option of doing these as targeted tests, since they will take considerably longer time, and it may not be desirable to do them with the regular tests.

**Definition of done:**

List of tests implemented under the `test` command.

PR #1331, opened 2022-12-14
tests: Add troubleshooting script for logging/certificates/ingress

### LINKED ISSUE #1193: Troubleshooting scripts - Logging

Compliant Kubernetes needs more troubleshooting scripts to help operators quickly figure out faults in an environment.
We could use some checks for the logging stack so we can quickly assert that it works as it should.

Check and present these in some nice way:

**OpenSearch:**
- Check cluster status
- Check snapshot status
- Check breakers
- Check indices
- Check aliases
- Check mappings
- *Already have an alert* - Check mappings count
- *Should also be an alert* - Check shard allocation / skew
- Check index state management
- Check users and roles
- Check object store access

**Fluentd:**
- Check connection to OpenSearch

*If time permits or create a separate issue:*

**OpenSearch Dashboards:**
- Check type conflicts in index patterns
- Check authentication issues
    - *401 loop*
    
**Fluentd:**
- Using logs from all Fluentd pods:
    - Filter logs to highlight errors or warnings and the reason
    - Check if there has been a mapping conflict

Alternatives would be to check manually and ensure that the config is correct.

**Additional context**

Note that it could be good to have the option of doing these as targeted tests, since some make take considerably longer time.

**Definition of done:**

List of tests implemented under the `test` command.

ayoubeddafali:

Compliant Kubernetes needs more troubleshooting scripts to help operators quickly figure out faults in an environment.
We could use some checks for the logging stack so we can quickly assert that it works as it should.

**Which issue this PR fixes** : fixes #1193 #1191

### DIFF
--- a/bin/test.bash
+++ b/bin/test.bash
@@ -8,6 +8,8 @@ set -eu -o pipefail
 here="$(dirname "$(readlink -f "$0")")"
 # shellcheck source=bin/common.bash
 source "${here}/common.bash"
+# shellcheck source=pipeline/test/services/service-cluster/testOpensearch.sh
+source "${pipeline_path}/test/services/service-cluster/testOpensearch.sh"
 
 test_apps_sc() {
     log_info "Testing service cluster"
@@ -23,17 +25,61 @@ test_apps_wc() {
         "${pipeline_path}/test/services/test-wc.sh" "${config[config_file_wc]}"
 }
 
-#
-# ENTRYPOINT
-#
+function sc_help() {
+    printf "%s\n" "[Usage]: test sc [target] [ARGUMENTS]"
+    printf "%s\n" "List of targets:"
+    printf "\t%-23s %s\n" "opensearch" "Open search checks"
+    printf "%s\n" "[NOTE] If no target is specified, the tests will go through all of them."
+    exit 0
+}
 
-config_load "$1"
-if [[ $1 == "wc" ]]; then
-    test_apps_wc
-elif [[ $1 == "sc" ]]; then
-    test_apps_sc
-else
-    echo "ERROR:  [$1] is an invalid argument:"
-    echo "usage:   ck8s test <wc|sc>. "
-    exit 1
-fi
+function sc() {
+    if [[ ${#} == 0 ]]; then
+        test_apps_sc
+    else
+        case ${1} in
+        opensearch)
+            sc_opensearch_checks "${@:2}"
+            ;;
+        --help)
+            sc_help
+            ;;
+        *)
+            echo "unknown command: $1"
+            sc_help 1
+            exit 1
+            ;;
+        esac
+    fi
+    exit 0
+}
+
+function main_help() {
+    echo list of commands:
+    printf "%-23s %s\n" "help" "show help menu and commands"
+    printf "%-23s %s\n" "sc" "Run sc checks"
+    printf "%-23s %s\n" "wc" "Run wc checks"
+
+    exit "${1:-0}"
+}
+
+function main() {
+    if [[ ${#} == 0 ]]; then
+        main_help 0
+    fi
+
+    case ${1} in
+    sc)
+        config_load "$1"
+        with_kubeconfig "${config[kube_config_sc]}" \
+            "$1" "${@:2}"
+        ;;
+    wc)
+        config_load "$1"
+        with_kubeconfig "${config[kube_config_wc]}" \
+            "$1" "${@:2}"
+        ;;
+    esac
+}
+
+main "$@"
--- a/bin/test.bash
+++ b/bin/test.bash
@@ -11,6 +11,15 @@ source "${here}/common.bash"
 # shellcheck source=pipeline/test/services/service-cluster/testOpensearch.sh
 source "${pipeline_path}/test/services/service-cluster/testOpensearch.sh"
 
+# shellcheck source=pipeline/test/services/service-cluster/testCertManager.sh
+source "${pipeline_path}/test/services/service-cluster/testCertManager.sh"
+# shellcheck source=pipeline/test/services/workload-cluster/testCertManager.sh
+source "${pipeline_path}/test/services/workload-cluster/testCertManager.sh"
+# shellcheck source=pipeline/test/services/service-cluster/testIngress.sh
+source "${pipeline_path}/test/services/service-cluster/testIngress.sh"
+# shellcheck source=pipeline/test/services/workload-cluster/testIngress.sh
+source "${pipeline_path}/test/services/workload-cluster/testIngress.sh"
+
 test_apps_sc() {
     log_info "Testing service cluster"
 
@@ -29,7 +38,18 @@ function sc_help() {
     printf "%s\n" "[Usage]: test sc [target] [ARGUMENTS]"
     printf "%s\n" "List of targets:"
     printf "\t%-23s %s\n" "opensearch" "Open search checks"
-    printf "%s\n" "[NOTE] If no target is specified, the tests will go through all of them."
+    printf "\t%-23s %s\n" "cert-manager" "Cert Manager checks"
+    printf "\t%-23s %s\n" "ingress" "Ingress checks"
+    printf "%s\n" "[NOTE] If no target is specified, the default sc apps tests will be executed."
+    exit 0
+}
+
+function wc_help() {
+    printf "%s\n" "[Usage]: test wc [target] [ARGUMENTS]"
+    printf "%s\n" "List of targets:"
+    printf "\t%-23s %s\n" "cert-manager" "Cert Manager checks"
+    printf "\t%-23s %s\n" "ingress" "Ingress checks"
+    printf "%s\n" "[NOTE] If no target is specified, the default wc apps tests will be executed."
     exit 0
 }
 
@@ -41,12 +61,42 @@ function sc() {
         opensearch)
             sc_opensearch_checks "${@:2}"
             ;;
+        cert-manager)
+            sc_cert_manager_checks "${@:2}"
+            ;;
+        ingress)
+            sc_ingress_checks "${@:2}"
+            ;;
         --help)
             sc_help
             ;;
         *)
             echo "unknown command: $1"
             sc_help 1
+            exit 1
+            ;;
+        esac
+    fi
+    exit 0
+}
+
+function wc() {
+    if [[ ${#} == 0 ]]; then
+        test_apps_wc
+    else
+        case ${1} in
+        cert-manager)
+            wc_cert_manager_checks "${@:2}"
+            ;;
+        ingress)
+            wc_ingress_checks "${@:2}"
+            ;;
+        --help)
+            wc_help
+            ;;
+        *)
+            echo "unknown command: $1"
+            wc_help 1
             exit 1
             ;;
         esac
--- a/pipeline/test/services/service-cluster/testOpensearch.sh
+++ b/pipeline/test/services/service-cluster/testOpensearch.sh
@@ -1,0 +1,429 @@
+#!/bin/bash
+# (return 0 2>/dev/null) && sourced=1 || sourced=0
+
+INNER_SCRIPTS_PATH="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
+# shellcheck source=bin/common.bash
+source "${INNER_SCRIPTS_PATH}/../funcs.sh"
+
+pushd "${CK8S_CONFIG_PATH}" || exit
+adminPassword=$(sops -d secrets.yaml | yq4 '.opensearch.adminPassword')
+fluentdPassword=$(sops -d secrets.yaml | yq4 '.opensearch.fluentdPassword')
+opsDomain=$(yq4 '.global.opsDomain' common-config.yaml)
+popd || exit
+
+function opensearch_check_help() {
+    printf "%s\n" "[Usage]: test sc opensearch [ARGUMENT]"
+    printf "\t%-25s %s\n" "--cluster-health" "Get cluster health"
+    printf "\t%-25s %s\n" "--snapshot-status" "Check snapshot status"
+    printf "\t%-25s %s\n" "--breakers" "Check if circuit breakers have been triggered"
+    printf "\t%-25s %s\n" "--indices" "Check if there are any missing indices"
+    printf "\t%-25s %s\n" "--aliases" "Check if each aliase has a write index"
+    printf "\t%-25s %s\n" "--mappings" "Check mappings/fields count & limit"
+    printf "\t%-25s %s\n" "--user-roles" "Check configured user roles"
+    printf "\t%-25s %s\n" "--ism" "Check ISM"
+    printf "\t%-25s %s\n" "--object-store-access" "Check object store access"
+    printf "\t%-25s %s\n" "--fluentd" "Check that fluentd can connect to opensearch"
+    printf "%s\n" "[NOTE] If no argument is specified, it will go over all of them."
+
+    exit 0
+}
+
+function sc_opensearch_checks() {
+    if [[ ${#} == 0 ]]; then
+        check_opensearch_cluster_health
+        check_opensearch_snapshots_status
+        check_opensearch_breakers
+        check_opensearch_indices
+        check_opensearch_aliases
+        check_opensearch_mappings
+        check_opensearch_user_roles
+        check_opensearch_ism
+        check_object_store_access
+        check_fluentd_connection
+        exit 0
+    fi
+    while [[ ${#} -gt 0 ]]; do
+        case ${1} in
+        --cluster-health)
+            check_opensearch_cluster_health
+            ;;
+        --snapshot-status)
+            check_opensearch_snapshots_status
+            ;;
+        --breakers)
+            check_opensearch_breakers
+            ;;
+        --indices)
+            check_opensearch_indices
+            ;;
+        --aliases)
+            check_opensearch_aliases
+            ;;
+        --mappings)
+            check_opensearch_mappings
+            ;;
+        --user-roles)
+            check_opensearch_user_roles
+            ;;
+        --ism)
+            check_opensearch_ism
+            ;;
+        --object-store-access)
+            check_object_store_access
+            ;;
+        --fluentd)
+            check_fluentd_connection
+            ;;
+        --help)
+            opensearch_check_help
+            ;;
+        esac
+        shift
+    done
+}
+
+check_opensearch_cluster_health() {
+    echo -ne "Checking if opensearch cluster is healthy ... "
+    cluster_health=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_cluster/health")
+    status=$(echo "$cluster_health" | jq -r '.status')
+    if [[ $status != "green" ]]; then
+        echo -e "failure ❌"
+        echo "$cluster_health" | jq
+    else
+        echo -e "success ✔"
+    fi
+}
+
+check_opensearch_snapshots_status() {
+    echo -ne "Checking opensearch snapshots status ... "
+    no_error=true
+    debug_msg=""
+    repo_name=$(yq4 -e '.opensearch.snapshot.repository' "${config['config_file_sc']}")
+    repo_exists_status=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_snapshot/${repo_name}" | jq "select(.error)")
+    if [[ -z "$repo_exists_status" ]]; then
+        if kubectl get "cronjob" -n "opensearch-system" "opensearch-backup" &>/dev/null; then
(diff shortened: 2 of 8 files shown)


ayoubeddafali:
Tests are targeted when specifying the command arguments, but still having a hard time making this run as part of the test_apps_sc

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,206 @@
...
+  done
+
+  echo "success ✔"
+  return 0
+
+}
+
+check_opensearch_mappings(){
+  curl -sk -u admin:${adminPassword} -X GET "https://opensearch.${opsDomain}/_mapping" | jq
+}
```
aarnq, line 174: The thing we want from this is to check so that the number of mappings/fields per index is in a healthy amount compared to the field limit configured on the indices themselves.

Your can use get on `<opensearch-url>/<index>/_field_caps?fields=*` to get a nice output of all mapped fields. And if you get the settings of the index you'll get the limit.

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,206 @@
+#!/bin/bash
+# (return 0 2>/dev/null) && sourced=1 || sourced=0
+
+INNER_SCRIPTS_PATH="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
+source "${INNER_SCRIPTS_PATH}/../funcs.sh"
+
+pushd ${CK8S_CONFIG_PATH}
+adminPassword=$(sops -d secrets.yaml | yq r  - opensearch.adminPassword)
+opsDomain=$(cat common-config.yaml | yq r - global.opsDomain)
```
aarnq, line 9: Use yq4 for these.

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,206 @@
+#!/bin/bash
+# (return 0 2>/dev/null) && sourced=1 || sourced=0
+
+INNER_SCRIPTS_PATH="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
+source "${INNER_SCRIPTS_PATH}/../funcs.sh"
```
aarnq, line 5: In that file change the check on `$PIPELINE` variable to reference it with `${PIPELINE-}` as it not set when run via this script.

aarnq (commented):
Nice to get started on the troubleshooting scripts epic!

bin/test.bash
```diff
@@ -23,17 +24,48 @@ test_apps_wc() {
...
+    echo list of commands:
+    printf "%-23s %s\n" "help" "show help menu and commands";
+    printf "%-23s %s\n" "sc" "Run sc checks";
+    printf "%-23s %s\n" "wc" "Run wc checks";
+
+    exit ${1:-0};
+}
+
+function main(){
+    if (( ${#} == 0 )); then
```
Xartos, line 58: Let's try to use the same way of doing this
```suggestion
    if [[ ${#} == 0 ]]; then
```

pipeline/test/services/service-cluster/testOpensearch.sh
ayoubeddafali: Thanks a lot, I was having a hard time finding what to test on this

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,206 @@
...
+
+    exit 0;
+}
+
+function sc_opensearch_checks(){
+    if [[ ${#} == 0 ]]; then
+        echo "Running all checks";
+        exit 0
+    fi
+    case ${1} in
```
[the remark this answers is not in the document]
ayoubeddafali: That's a good idea, I'll try that although bash is giving me a hard time 😃

ayoubeddafali:
Ready for review

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,385 @@
...
+  fi
+
+}
+
+check_opensearch_aliases(){
+  echo -ne "Checking opensearch aliase indices mapping ... "
+  no_error=true
+  debug_msg=""
+
+  curl -sk -o /tmp/response -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_cat/aliases"
```
aarnq, line 180: If you want to I think you can simplify this to just look at rightmost column of the cat output as it shows if it is a write index or not.

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,385 @@
...
+
+check_opensearch_snapshots_status(){
+  echo -ne "Checking opensearch snapshots status ... "
+  repo_name=$(yq4 -e '.opensearch.snapshot.repository' "${config['config_file_sc']}")
+  repo_exists_status=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_snapshot/${repo_name}" | jq "select(.error)")
+  if [[ -z "$repo_exists_status" ]]; then
+    if kubectl get "cronjob" -n "opensearch-system" "opensearch-backup" &> /dev/null
+    then
+        last_schedule=$(kubectl get cronjob -n opensearch-system opensearch-backup -ojsonpath='{.status.lastScheduleTime}')
+        if [[ $last_schedule != $(date +"%Y-%m-%dT12:00:00Z") ]]; then
```
aarnq, line 106: This check will not be reliable as the time may be different. Could you also consider to add a check of the current snapshots via `_cat/snapshots/$repo_name` so that the snapshots have completed and that the last is within the configured snapshot interval?

bin/test.bash
```diff
@@ -23,17 +25,58 @@ test_apps_wc() {
...
+    exit 0;
+}
+
+function sc() {
+    if [[ ${#} == 0 ]]; then
+        test_apps_sc
+    else
+        case ${1} in
+            opensearch )
+                sc_opensearch_checks "${@:2}";
```
aarnq, line 44:
```suggestion
                with_kubeconfig "${config[kube_config_sc]}" \
                    sc_opensearch_checks "${@:2}";
```
Else the kubeconfig won't be set properly. I think we need to rethink that in general here since it doesn't make much sense how it's done a bit of everywhere, but we can do it in another issue when we refactor this a bit.

ayoubeddafali: I'm not sure how it is not set properly, but in my case, It's working fine. 
I've modified in the main function so it looks like this : 
```

function main(){
    if [[ ${#} == 0 ]]; then
        main_help 0;
    fi

    case ${1} in
        sc )
            config_load "$1";
            with_kubeconfig "${config[kube_config_sc]}" \
              "$1" "${@:2}";
        ;;
        wc )
            config_load "$1";
            with_kubeconfig "${config[kube_config_wc]}" \
              "$1" "${@:2}";
        ;;
    esac
}
```

pipeline/test/services/service-cluster/testOpensearch.sh
ayoubeddafali: Yep, I saw that, but it was tricky to keep track of which alias has the missing write index

pipeline/test/services/service-cluster/testOpensearch.sh
ayoubeddafali: Sure

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,408 @@
...
+            debug_msg+="[ERROR] The snapshot ${snapshot_data[0]} terminated in an error and no data was stored.\n"
+          fi
+        done
+
+        IFS=" " read -ra last_snapshot <<< "${data[-1]}"
+
+        if [[ ${last_schedule_epoch_time} -gt ${last_snapshot[2]}  ]]; then
+          no_error=false
+          debug_msg+="[ERROR] opensearch-backup cronjob has not been scheduled\n  "
+        fi
```
aarnq, line 125: What is actually checked here, that there is a snapshot done when it was asked to?

I think we should rework it to look at simply if the latest snapshot in opensearch was created within the configured window for snapshots and not based on the cronjob.
(We could simplify and say within 24 hours.)

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,408 @@
...
+    echo "[DEBUG] None of the circuit breakers has been triggered"
+  else
+    echo "failure ❌";
+    echo -e "$debug_msg"
+  fi
+
+}
+
+check_opensearch_aliases(){
+  echo -ne "Checking opensearch aliase indices mapping ... "
```
aarnq, line 199:
```suggestion
  echo -ne "Checking opensearch aliases indices mapping ... "
```

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,408 @@
...
+
+  aliases=$(awk '{print $1}' < /tmp/response)
+  aliases_arr=("$aliases")
+
+  uniq_aliases=$(echo "${aliases_arr[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' ')
+
+  for alias in $uniq_aliases; do
+    write_index_found=false
+    if ! [[ $alias =~ ^[\.*] ]]; then
+      alias_data=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_alias/${alias}")
```
aarnq, line 213: Suggestion to reduce the loops as they take quite some time with many indices.

```suggestion
      alias_data=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_alias/${alias}")
      
      # Note: Only set to true when found, empty otherwise
      is_write_index="$(echo "$alias_data" | jq ".[].aliases.$alias.is_write_index | select(. == true)")"
```

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,408 @@
...
+}
+
+check_opensearch_mappings(){
+  echo -ne "Checking opensearch mappings/fields ... "
+  no_error=true
+  debug_msg="INDEX\t\t\t#FIELDS\tLIMIT\n"
+  debug_msg+="------------------------------------------------\n"
+
+  mappings_data=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_mapping")
+  readarray -t indices < <(jq -rc 'keys[]' <<< "$mappings_data")
```
aarnq, line 250: You can use the CAT API and awk to just pull out the names to simplify it.

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,403 @@
...
+              check_opensearch_aliases
+          ;;
+          --mappings )
+              check_opensearch_mappings
+          ;;
+          --user-roles )
+              check_opensearch_user_roles
+          ;;
+          --ism )
+              check_opensearch_ism
```
aarnq, line 70: Could you add a check to see if ISM is rolling over write indices as it should?
With `GET _plugins/_ism/policies/authlog` or from the config you should be able to get how old the index may be before it is rolled over. Then you can just inspect the creation date on the write indices to check if they are older than that.

aarnq (commented):
Overall I think it looks good, but do reindent so it is consistent, I think we run 4 spaces as our standard.

pipeline/test/services/service-cluster/testOpensearch.sh
ayoubeddafali: Sure, it's added now

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,426 @@
...
+    for policy in "${additional_policies[@]}"; do
+        policy_name=$(echo "$policy" | awk -F'.' '{print $1}' | tr '"' ' ')
+        res=$(curl -w "%{http_code}" -o /dev/null -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_plugins/_ism/policies/${policy_name}")
+        if [[ $res != "200" ]]; then
+            no_error=false
+            debug_msg+="[ERROR] Missing additional policy : ${policy_name}\n"
+        fi
+    done
+
+    indices_data=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_cat/indices" | awk '{print $3}' | tr '\n' ' ')
```
aarnq, line 335: Try to prefilter and just get what you need instead of fetching all and inspecting each as it becomes really slow with many indices. (25 seconds for an empty production-like environment.)
```suggestion
    indices_data=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_cat/aliases?format=json" | jq -r '.[] | select(.is_write_index == "true") | .index')
```
And yes `is_write_index` through CAT is apparently a string.

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,426 @@
...
+    IFS=' ' read -ra indices <<<"$indices_data"
+
+    for index in "${indices[@]}"; do
+        index_data=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/$index")
+        index_alias=$(echo "$index_data" | jq -r ".\"$index\".settings.index.plugins.index_state_management.rollover_alias")
+        is_write_index=$(echo "$index_data" | jq -r ".\"$index\".aliases.$index_alias.is_write_index")
```
aarnq, line 341: Then you can simplify this. And if you want to further optimise I'd suggest using CAT to only fetch the milliseconds creation date using `_cat/indices/<index>?h=creation.date`.

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,426 @@
...
+            rollover_age_days=$(yq4 -e '.opensearch.ism.rolloverAgeDays' "${config['config_file_sc']}")
+            epoch_now=$(date +%s)
+            index_creation_date=$(echo "$index_data" | jq -r ".\"$index\".settings.index.creation_date")
+
+            index_creation_date_seconds=$(echo "scale=3; $index_creation_date/1000" | bc)
+            rollover_diff=$(echo "scale=3; ($epoch_now-$index_creation_date_seconds)/86400" | bc | awk '{printf "%f", $0}')
+            if (($(echo "$rollover_diff > $rollover_age_days" | bc -l))); then
+                no_error=false
+                echo -e "[ERROR]The write index : $index is old and has not been rolled over as expected\n"
+            fi
```
aarnq, line 353: Is there a reason why you calculate divisions here?

I'd suggest sticking to integers and use the smallest unit if possible, because the `printf` and `bc` combo becomes locale dependant and doesn't work for me. :sweat_smile:

pipeline/test/services/service-cluster/testOpensearch.sh
ayoubeddafali: Thanks a lot, it's changed now

pipeline/test/services/service-cluster/testOpensearch.sh
ayoubeddafali: I've changed it now so I can work with milliseconds directly

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,422 @@
...
+    indices_data=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_cat/indices" | awk '{print $3}' | tr '\n' ' ')
+    IFS=' ' read -ra indices <<<"$indices_data"
+    for index in "${indices[@]}"; do
+        fields_limit=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/${index}/_settings" | jq -r ".[\"${index}\"].settings.index.mapping.total_fields.limit")
+        fields_count=$(curl -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/${index}/_field_caps?fields=*" | jq -r ".fields | keys | length")
+
+        if [[ $fields_limit != "null" ]] && [[ $fields_count -gt $fields_limit ]]; then
+            no_error=false
+        fi
+        debug_msg+="${index}\t\t\t${fields_count}\t${fields_limit}\n"
```
aarnq, line 252: Only remark I have is if you could limit the output to only show when the current mappings reaches like 50 % of the limit just to keep focus on what's important.

aarnq (approved):
Looks good to me.

pipeline/test/services/service-cluster/testOpensearch.sh
ayoubeddafali: Sure, will try to do that

ayoubeddafali: Done

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,431 @@
...
+        for policy in "${default_policies[@]}"; do
+            res=$(curl -w "%{http_code}" -o /dev/null -sk -u admin:"${adminPassword}" -X GET "https://opensearch.${opsDomain}/_plugins/_ism/policies/${policy}")
+            if [[ $res != "200" ]]; then
+                no_error=false
+                debug_msg+="[ERROR] Missing default policy : ${policy}\n"
+            fi
+        done
+    fi
+
+    # additional_policies=$(yq4 -e '.opensearch.ism.additionalPolicies | keys' "${config['config_file_sc']}")
```
robinAwallace, line 326:
```suggestion
```
Remove?

pipeline/test/services/service-cluster/testOpensearch.sh
```diff
@@ -0,0 +1,431 @@
...
+    readarray -t additional_policies < <(yq4 e -o=j -I=0 '.opensearch.ism.additionalPolicies | keys' "${config['config_file_sc']}")
+
+    additional_policies=("${additional_policies//,/ }")
+    additional_policies=("${additional_policies##[}")
+    additional_policies=("${additional_policies%]}")
+    # additional_policies=($additional_policies)
```
robinAwallace, line 332:
```suggestion
```
Same here?

robinAwallace (commented):
Looked through it but did not test it. But it looks good and seems like you have covered most or everything.

pipeline/test/services/service-cluster/testOpensearch.sh
ayoubeddafali: Yep, thanks a lot, I forgot to do the comments cleanup

pipeline/test/services/service-cluster/testOpensearch.sh
ayoubeddafali: Cleaned up, thanks 👍🏽

bin/test.bash
```diff
@@ -23,17 +25,63 @@ test_apps_wc() {
...
-    echo "ERROR:  [$1] is an invalid argument:"
-    echo "usage:   ck8s test <wc|sc>. "
-    exit 1
-fi
+function sc_help() {
+    printf "%s\n" "[Usage]: test sc [target] [ARGUMENTS]"
+    printf "%s\n" "List of targets:"
+    printf "\t%-23s %s\n" "opensearch" "Open search checks"
+    printf "\t%-23s %s\n" "fluentd" "Fluentd checks"
+    printf "\t%-23s %s\n" "opensearch-dashboards" "Opensearch dashboards checks"
```
Xartos, line 33: Maybe we can delete these since they don't exists

```suggestion
```

bin/test.bash
```diff
@@ -23,17 +25,63 @@ test_apps_wc() {
...
+    case ${1} in
+    sc)
+        config_load "$1"
+        with_kubeconfig "${config[kube_config_sc]}" \
+            "$1" "${@:2}"
+        ;;
+    wc)
+        config_load "$1"
+        with_kubeconfig "${config[kube_config_wc]}" \
+            "$1" "${@:2}"
```
Xartos, line 82: Don't we need to add the function `wc` for this to work? It doesn't seem to work for me at least

Xartos (changes requested):

bin/test.bash
ayoubeddafali: Yes, I've added it in the other PR ( certificates & ingresses ), because in this one all checks are sc related

bin/test.bash
ayoubeddafali: Sure

bin/test.bash
ayoubeddafali: Yes, that's the plan. 
The other PR has the current branch as target

Xartos: Right, good :+1:

Xartos (approved):

PR #1332, opened 2022-12-15
pipeline: update actions

aarnq:

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -22,42 +22,65 @@ jobs:
         flavor: ["prod", "dev"]
     steps:
     - name: Checkout compliantkubernetes-apps
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
+    - name: Set up Docker
+      uses: docker/setup-buildx-action@v2
+    - name: Build Docker image
+      uses: docker/build-push-action@v3
       with:
-        fetch-depth: 0 # Fetches all tags
+        cache-from: type=gha
+        context: ./pipeline
+        load: true
+        pull: false
+        push: false
+        tags: elastisys-ck8s-apps/app:latest
     - name: Init and configure config for compliantkubernetes-apps
-      uses: ./pipeline
-      with:
-        args: bash -c "./pipeline/sops-pgp-setup.bash && ./pipeline/init-unit-tests.bash"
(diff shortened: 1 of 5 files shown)


aarnq:
@crssnd please take a look and let me know what you think about the build step. If we add a similar one in e2e then we can build the cache there on the main branch so all other branches can access it.
Feels a bit less wasteful than the current method of rebuilding 850 MB worth of images in each job and we don't need to involve a registry for PRs.

crssnd:
> @crssnd please take a look and let me know what you think about the build step. If we add a similar one in e2e then we can build the cache there on the main branch so all other branches can access it. Feels a bit less wasteful than the current method of rebuilding 850 MB worth of images in each job and we don't need to involve a registry for PRs.

yes, I think it's a good idea. we should try it

aarnq:
I think I should test it for the end2end but I think I'll wait on the terraform upgrade to be merged first.

Xartos (commented):
So do I understand correctly that this basically uses the cache as registry for the image? If so that's a very neat solution

aarnq:
> So do I understand correctly that this basically uses the cache as registry for the image? If so that's a very neat solution

It does indeed. And caches are shared with subbranches so we should always have a fresh cache from the e2e.

Xartos (approved):

crssnd (approved):

.github/workflows/pipeline.yml
```diff
@@ -22,42 +22,65 @@ jobs:
...
+      run: /usr/bin/docker run --rm -e "CK8S_CONFIG_PATH" -e "CK8S_CLOUD_PROVIDER" -e "CI_CK8S_FLAVOR" -e "CK8S_PGP_FP" -e "GITHUB_RUN_ID" -e CI=true -v "/home/runner/work/compliantkubernetes-apps/compliantkubernetes-apps":"/github/workspace" -w /github/workspace elastisys-ck8s-apps/app:latest bash -c "./pipeline/sops-pgp-setup.bash && ./pipeline/init-unit-tests.bash"
       env:
         CK8S_CLOUD_PROVIDER: ${{ matrix.cloud }}
         CI_CK8S_FLAVOR: ${{ matrix.flavor }}
         CK8S_PGP_FP: 529D964DE0BBD900C4A395DA09986C297F8B7757
     - name: Run linters
-      uses: ./pipeline
-      with:
-        args: bash -c "./pipeline/sops-pgp-setup.bash && ./pipeline/lint.bash"
+      run: /usr/bin/docker run --rm -e "CK8S_CONFIG_PATH" -e "CK8S_CLOUD_PROVIDER" -e "CI_CK8S_FLAVOR" -e "CK8S_PGP_FP" -e "GITHUB_RUN_ID" -e CI=true -v "/home/runner/work/compliantkubernetes-apps/compliantkubernetes-apps":"/github/workspace" -w /github/workspace elastisys-ck8s-apps/app:latest bash -c "./pipeline/sops-pgp-setup.bash && ./pipeline/lint.bash"
```
robinAwallace, line 44: Those there exist a "uses" that takes a docker image?
My thought is just to clear up the one liner :smile:

robinAwallace: Nvm did not really find any good option

aarnq: Yeah, that is the problem with this approach.

PR #1340, opened 2022-12-22
migration: added a migration file template

crssnd:
I want to make it easier to create the migration document and also not miss updating the versions.
Also, I added some pre and post maintenance steps as they might be helpful.

I keep the script simple and it will work only for major releases.
Further improvements will be made if it's considered useful.


davidumea (commented):
The `## Steps` section should be split into two, one for steps that can be done before the maintenance window and one for disruptive steps.

I also think that the `migration.sh` script should be named to something else (something like `create-migration-document.sh`), as `migration.sh` can be interpreted as a script that does an actual migration in an environment.

migration/template/upgrade-apps.md
```diff
@@ -0,0 +1,55 @@
...
+    ```bash
+    ./bin/ck8s test sc|wc
+    kubectl get po -A
+    kubectl get nodes
+    kubectl get jobs -A
+    velero get backup
+    ```
+
+- [ ] Enable maintenance mode for the monitoring solution;
+- [ ] Notify the on-call engineers and the users before the maintenance starts.
```
OlleLarsson, line 17: I would say that 
- Check if there are any pending changes to the environment;
- Enable maintenance mode for the monitoring solution;
- Notify the on-call engineers and the users before the maintenance starts.
is too specific.

migration/template/upgrade-apps.md
```diff
@@ -0,0 +1,55 @@
...
+- [ ] Check the state of the environment, pods and nodes:
+
+    ```bash
+    ./bin/ck8s test sc|wc
+    kubectl get po -A
+    kubectl get nodes
+    ```
+
+- [ ] Disable the maintenance mode for the monitoring solution;
+- [ ] Notify the on-call engineers and the users that the maintenance was completed.
```
OlleLarsson, line 53: Same

migration/template/upgrade-apps.md
crssnd: why do you think that?
I see them quite general and applicable for every application maintenance, not only compliantkubernetes-apps

OlleLarsson: Well, 
- `Check if there are any pending changes to the environment;` more or less assumes our internal workflow with github PRs. This is definitely not a general case.
- `Enable maintenance mode for the monitoring solution;`, not every user as an external monitoring solution.
- `Notify the on-call engineers and the users before the maintenance starts.`, is very specific to us.

crssnd: `Check if there are any pending changes to the environment` - yes it's assumes that the overwrite configs are keep in a version control system. in my view a common use case
`Enable maintenance mode for the monitoring solution` - true, but also alertmanager has also a similar function ([silences](https://prometheus.io/docs/alerting/latest/alertmanager/#silences)), I can change the wording to also highlight that
`Notify the on-call engineers and the users before the maintenance starts` - agree I can replace maintenance with upgrade and also drop the on-call engineers. regarding users, as they are supported for wc and mentioned in the docs I want to keep them

migration/template/upgrade-apps.md
```diff
@@ -0,0 +1,60 @@
...
+- [ ] Check the state of the environment, pods and nodes:
+
+    ```bash
+    ./bin/ck8s test sc|wc
+    kubectl get po -A
+    kubectl get nodes
+    ```
+
+- [ ] Enable the notifications for the alerts;
+- [ ] Notify the users (if any) before the upgrade is complete;
```
davidumea, line 55:
```suggestion
- [ ] Notify the users (if any) when the upgrade is complete;
```

migration/template/upgrade-apps.md
```diff
@@ -0,0 +1,60 @@
...
+
+- [ ] Notify the users (if any) before the upgrade starts;
+- [ ] Check if there are any pending changes to the environment;
+- [ ] Check the state of the environment, pods, nodes and backup jobs:
+
+    ```bash
+    ./bin/ck8s test sc|wc
+    kubectl get po -A
+    kubectl get nodes
+    kubectl get jobs -A
```
Xartos, line 13:
```suggestion
    ./bin/ck8s ops kubectl sc get po -A
    ./bin/ck8s ops kubectl sc get nodes
    ./bin/ck8s ops kubectl sc get jobs -A
```

migration/template/upgrade-apps.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Prerequisites
+
+- [ ] Notify the users (if any) before the upgrade starts;
+- [ ] Check if there are any pending changes to the environment;
+- [ ] Check the state of the environment, pods, nodes and backup jobs:
+
+    ```bash
+    ./bin/ck8s test sc|wc
+    kubectl get po -A
```
Xartos, line 11: To make it a bit easier to read
```suggestion
    kubectl get po -A | grep -vP "(Running|Completed)"
```

davidumea: There could be pods that are "Running" but not ready that wouldn't be caught when you do this

davidumea (approved):

migration/template/upgrade-apps.md
crssnd: not perfect, but would this work?
`kubectl get pods -A -o custom-columns=NAMESPACE:metadata.namespace,POD:metadata.name,READY-false:status.containerStatuses[*].ready | grep false`

Xartos: How does this work if there are multiple containers? Will the column `READY-false` just take one or does it have multiple values?

migration/create-migration-document.sh
```diff
@@ -0,0 +1,37 @@
...
+
+usage() {
+    echo "This script must have apps old and new major versions as arguments."
+    echo "Usage: $0 [old_version] [new_version]"
+    echo "Example: ./migration.sh v0.22 v0.23"
+    }
+    if [  $# -lt 2 ]; then
+        usage
+        exit 1
+    fi
```
Xartos, line 13:
```suggestion
}
if [  $# -lt 2 ]; then
    usage
    exit 1
fi
```

migration/create-migration-document.sh
```diff
@@ -0,0 +1,37 @@
...
+
+here="$(dirname "$(readlink -f "$0")")"
+export old_version="${1}"
+export new_version="${2}"
+folder_name="${here}/v${old_version}.x-v${new_version}.x"
```
Xartos, line 18: In the usage instructions you say to include the `v` while here you expect them not to. But I would say that you could assume their there (or implement some logic to support both I guess)

```suggestion
folder_name="${here}/${old_version}.x-${new_version}.x"
```

migration/template/upgrade-apps.md
```diff
@@ -0,0 +1,60 @@
+# Upgrade v${old_version}.x to v${new_version}.x
```
Xartos, line 1: Same in this file
```suggestion
# Upgrade ${old_version}.x to ${new_version}.x
```

Xartos (commented):
Just a minor thing, otherwise I think it looks great

migration/template/upgrade-apps.md
Xartos: Yea, then that would work

Xartos (approved):

migration/template/upgrade-apps.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Prerequisites
+
+- [ ] Notify the users (if any) before the upgrade starts;
+- [ ] Check if there are any pending changes to the environment;
+- [ ] Check the state of the environment, pods, nodes and backup jobs:
+
+    ```bash
+    ./bin/ck8s test sc|wc
+    ./bin/ck8s ops kubectl sc get pods -A -o custom-columns=NAMESPACE:metadata.namespace,POD:metadata.name,READY-false:status.containerStatuses[*].ready | grep false
```
davidumea, line 11:
```suggestion
    ./bin/ck8s ops kubectl sc|wc get pods -A -o custom-columns=NAMESPACE:metadata.namespace,POD:metadata.name,READY-false:status.containerStatuses[*].ready,REASON:status.containerStatuses[*].state.terminated.reason | grep false | grep -v Completed
```

Completed pods are ready=false in the yaml/json format, so this previous command would show all completed pods as not ready. This excludes the Completed pods and also shows the reason why any pods are not ready.

migration/template/upgrade-apps.md
```diff
@@ -0,0 +1,60 @@
...
+    ./bin/ck8s ops kubectl sc get nodes
+    ./bin/ck8s ops kubectl sc get jobs -A
```
davidumea, line 13:
```suggestion
    ./bin/ck8s ops kubectl sc|wc get nodes
    ./bin/ck8s ops kubectl sc|wc get jobs -A
```

davidumea (changes requested):

davidumea (approved):

PR #1351, opened 2023-01-03
apps sc: synced timezone in all grafana dashboards

### LINKED ISSUE #1297: Standardize Grafana dashboard timezones

The default timezone in each of our Grafana dashboards are quite inconsistent, and they can't be changed through the Grafana GUI since they were created by another source. So when you update timezone in a dashboard through the top right time picker it doesn't persist and gets reset every time you refresh the page. Maybe the most frustrating part is that we have different timezones in different dashboards.

Change the default timezone in all of our dashboards to default organization timezone, which is just an empty string in the JSON model `""`

Any Grafana admin can change the default organization timezone, this means that the timezone will be the same for every user that is logged in.

An alternative was to change the timezones to "browser" timezone, but this can cause confusion if people are in different timezones.

**Additional context**

Screenshot from 2022-11-15 11-13-48

**Definition of done:**

All of our Grafana dashboards are using the same timezone.

davidumea:

**Which issue this PR fixes** : fixes #1297

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [X] I upgraded no Chart.

### DIFF
--- a/helmfile/upstream/kube-prometheus-stack/values.yaml
+++ b/helmfile/upstream/kube-prometheus-stack/values.yaml
@@ -640,7 +640,7 @@ grafana:
   ## Timezone for the default dashboards
   ## Other options are: browser or a specific timezone, i.e. Europe/Luxembourg
   ##
-  defaultDashboardsTimezone: utc
+  defaultDashboardsTimezone: ""
 
   adminPassword: prom-operator
 
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -8,6 +8,7 @@
 - Disabled all collectors for node-exporter in ciskubebench- and vulnerability-exporter except textcollector
 - Increased the default CPU limit for node-exporter in ciskubebench- and vulnerability-exporter
 - Nginx controller service annotations are now defined as a map, previously just a single string.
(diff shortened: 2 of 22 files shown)


raviranjanelastisys (approved):

helmfile/upstream/kube-prometheus-stack/values.yaml
```diff
@@ -640,7 +640,7 @@ grafana:
   ## Timezone for the default dashboards
   ## Other options are: browser or a specific timezone, i.e. Europe/Luxembourg
   ##
-  defaultDashboardsTimezone: utc
+  defaultDashboardsTimezone: ""
```
crssnd, line 643: not necessary as I see you already modified the `helmfile/values/kube-prometheus-stack-sc.yaml.gotmpl`

davidumea: Why not have it the same everywhere?

crssnd (approved):
looks good

helmfile/upstream/kube-prometheus-stack/values.yaml
crssnd: if we update the upstream chart this file will be overwritten
we use kube-prometheus-stack-sc.yaml.gotmpl to change any of the upstream chart defaults

PR #1356, opened 2023-01-09
tests: Add troubleshooting script for certificates/ingress

### LINKED ISSUE #1190: [3] Troubleshooting scripts - Certificates and Ingresses

Compliant Kubernetes needs more troubleshooting scripts to help operators quickly figure out faults in an environment.
We need checks for the status for the certificates and ingresses so we can assert that they work as they should.

Check for:

**cert-manager:**
- For ClusterIssuers:
    - Check exists
    - Check status
    - Check for custom solvers
- For Certificates:
    - Check exists
    - Check status
    - Check renewal time
- For Challenges:
    - Check for pending
    - Check for failures
    
**ingress-nginx:**
- Check targets IPs
- *Cloud dependent* - Load balancer integration:
    - Check load balancer exists
    - Check load balancer status
    - Check load balancer errors

And present these in some nice way.

Alternatives would be to check manually and ensure that the config is correct.

**Additional context**

Note that it could be good to have the option of doing these as targeted tests as well.

**Definition of done:**

List of tests implemented under the `test` command.

ayoubeddafali:
Compliant Kubernetes needs more troubleshooting scripts to help operators quickly figure out faults in an environment.
We need checks for the status for the certificates and ingresses so we can assert that they work as they should.

**Which issue this PR fixes** : fixes #1190

### DIFF
--- a/bin/test.bash
+++ b/bin/test.bash
@@ -11,6 +11,11 @@ source "${here}/common.bash"
 # shellcheck source=pipeline/test/services/service-cluster/testOpensearch.sh
 source "${pipeline_path}/test/services/service-cluster/testOpensearch.sh"
 
+# shellcheck source=pipeline/test/services/service-cluster/testCertManager.sh
+source "${pipeline_path}/test/services/service-cluster/testCertManager.sh"
+# shellcheck source=pipeline/test/services/workload-cluster/testCertManager.sh
+source "${pipeline_path}/test/services/workload-cluster/testCertManager.sh"
+
 test_apps_sc() {
     log_info "Testing service cluster"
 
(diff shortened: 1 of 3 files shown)


aarnq:
There is no need to recreate branches and open new PRs when updating branches depending on other branches, it only makes it harder to track reviews.

You can do `git fetch` to update, `git switch <this-branch>` to check out this branch, `git rebase <source-branch>` to rebase it on the updated source branch (optionally use `--interactive` if you want to control which commits to keep), and finally `git push --force-with-lease` to update this remote branch.

ayoubeddafali:
Yes, that what I usually do, but in the last commit, something seemed off when I rebased, and couldn't figure it out how to sync the two branches

bin/test.bash
```diff
@@ -31,6 +36,15 @@ function sc_help() {
     printf "\t%-23s %s\n" "opensearch" "Open search checks"
     printf "\t%-23s %s\n" "fluentd" "Fluentd checks"
     printf "\t%-23s %s\n" "opensearch-dashboards" "Opensearch dashboards checks"
+    printf "\t%-23s %s\n" "cert-manager" "Cert Manager checks"
+    printf "%s\n" "[NOTE] If no target is specified, the tests will go through all of them."
```
Xartos, line 40: Not really true if I understand it correctly. Since if you don't set TARGET it will run the old function `test_apps_sc` and not any of the new ones. And if you specify it it will run the selected new one.
Maybe you could change so that everyone was run if you didn't specify it and then you can add a new entry here like `basic-checks` or something that you can select to run the old function (`test_apps_sc`)

ayoubeddafali: Yes  you're right, seems like I forgot it. 
Sure, I will do that

ayoubeddafali: I just remembered that you can run all checks of a target like so : test sc opensearch
So I think to keep things like usual, I've changed the message so if no target is specified, it will behave like it is used to, and will run the default apps test

Issue #1396, opened 2023-02-03
clean-sc.sh leaves harbor and monitoring ns in Terminating


cristiklein:
**Describe the bug**

We recently used the `./clean-sc.sh` script against a Compliant Kubernetes on-prem environment. The `harbor` and `monitoring` namespaces were left behind in Terminating state, due to the finalizer of `challenges` being gone.

**To Reproduce**
1. Set up a SC.
2. Use `./clean-sc.sh`.

We get:

```
NamespaceContentRemaining                    True    Fri, 03 Feb 2023 12:55:34 +0000  SomeResourcesRemain   Some resources are remaining: challenges.acme.cert-manager.io has 1 resource instances
NamespaceFinalizersRemaining                 True    Fri, 03 Feb 2023 12:55:34 +0000  SomeFinalizersRemain  Some content in the namespace has finalizers remaining: finalizer.acme.cert-manager.io in 1 resource instances
```

**Expected behavior**

`./clean-sc.sh` should leave the cluster just like it was after Kubespray.

**Version (add all relevant versions):**
 - Compliant kubernetes apps version 0.28.0

**Additional context**

Perhaps we need to add `harbor` and `monitoring` [here](https://github.com/elastisys/compliantkubernetes-apps/blob/v0.28.0/scripts/clean-sc.sh#L20).

cristiklein:
@anders-elastisys Cool that you found a solution!

I believe that this script should clean the cluster and bring it to "pre-apps" state "no matter what". :smile:

Hence, if finalizers get in the way, I'd want this script to remove finalizers. See [this](https://stackoverflow.com/a/52820472) for a potential solution.

PR #1413, opened 2023-02-13
bin: Added patching of pending challenges with finalizers

anders-elastisys:

In certain situations cert-manager Challenges gets stuck in a pending state with a finalizer which would prevent certain namespaces such as monitoring and harbor from terminating after running the `clean-sc.sh` script. This PR solves this issue.

**Which issue this PR fixes:**

fixes [#1396](https://github.com/elastisys/compliantkubernetes-apps/issues/1396)

**Note** that I have not been able to fully replicate the situation described in the issue, but it has been tested that you can run the script without getting an error and that it can patch a Challenge, removing any finalizers.

The solution uses `kubectl patch` in this case, but a shorter version that should also work is turning any remaining challenges into json and replacing the finalizers with `jq` and then applying the changes. This approach is a bit shorter, e.g:
```bash
CHALLENGES="${here}/.././bin/ck8s" ops kubectl sc get challenge -A
if [ -n "$CHALLENGES" ]; then
    "${here}/.././bin/ck8s" ops kubectl sc get challenge -A -o json | jq '.items[].metadata.finalizers = null' | kubectl apply -f -
fi
```
Any feedback on preference between using the patch or the json approach is appreciated.

Also note that this PR only adds patching for the service cluster clean script. The same should apply for the workload cluster clean script.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [X] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -11,6 +11,7 @@
 
 - The update-ips script can now fetch Calico Wireguard IPs
 - The test scripts can now always access the kubeconfig
+- The `clean-sc` script now patches any pending challenges which would prevent the removal of certain namespaces
 
 ### Updated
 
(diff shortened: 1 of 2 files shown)


Xartos (approved):

raviranjanelastisys (approved):
LGTM !!

aarnq (approved):

PR #1417, opened 2023-02-15
migration: new upgrade process

### LINKED ISSUE #682: [5] Upgrade - change the upgrade and migration process
Automate the upgrade process by creating scripts instead of code snippets to be run manually.
Separate the init and config steps from the apply steps and scripts as the CD solution should only run the apply steps and scripts.

**Addition**
All steps that affect the cluster should be scripts.
There should exist a script `upgrade-apps.sh` part of the migration path folder that runs all steps affecting the cluster. 
Example:
Delete deployment sets
Directly updating config maps
Apply sc/wc

**Definition of done:**
1. a process on how and who should create this migration scripts, separated config and apply scripts
2. Separate config migration scripts and cluster changes scripts into two folders.
3. There exists a script `upgrade-apps.sh` to run all cluster changes scripts.
4. ~create templates for the migration scripts that can be reused for the main or patch releases~
5. ~template migration scripts for how to change yaml config in apps~
6. ~add the test steps in a script to run before and after the upgrade~

aarnq:

The new upgrade/migration process.

fixes #682 

I may have gone a bit overboard compared to the current issue but it was more organic to do it this way.
I still have some work to do on the documentation of the template snippets.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


aarnq:
Questions for reviewers:

- Should creating buckets be automatically done by apply snippets or done a manual task by the admin/operator?
- Should I move the upgrade prepare/apply into `bin/ck8s`?

migration/v0.29/upgrade.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
robinAwallace, line 1: As you have done such a nice script for upgrade, i would say yes you could move it to bin/apps

scripts/migration-library.sh
```diff
@@ -0,0 +1,575 @@
...
+    ;;
+  esac
+}
+
+# usage: kubectl_delete <sc|wc> <resource> <namespace> <name>
+kubectl_delete() {
+  kubectl_do "${1:-}" delete "${2:-}" --namespace "${3:-}" "${4:-}" --ignore-not-found true
+}
+
+# --- helm functions ---
```
robinAwallace, line 338: As you have done a good job separating the fucntions, i would say spit them up into separate files. So that this file does not become so long :smile:

migration/create-migration-document.sh
```diff
@@ -34,14 +34,18 @@ else
     echo "- ${folder_name} directory created"
 fi
 
-if [ -f "${folder_name}/upgrade-apps.md" ]; then
-    echo -n "- ${folder_name}/upgrade-apps.md exists. Do you want to replace it? (y/N): "
+if [ -f "${folder_name}/README.md" ]; then
```
robinAwallace, line 37: :pray:

viktor-f (commented):
This looks very nice!

Would be nice with some documentation for this in the main readme of the repo.

Make it clear somewhere that upgrades that are dependent on each other should be one single snippet. To make rollback possible.

WIP-CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 - Updated Gatekeeper Dashboard to new one from upstream
 - Renamed the ElasticSearch Dashboard to Opensearch in Grafana
 - Changed release name for `prometheus-elasticsearch-exporter` to `prometheus-opensearch-exporter`
+- The upgrade process has changed considerably, checkout the migration guide it also contains information about how to follow the new process when writing migration steps
```
OlleLarsson, line 31: Put under release notes instead?

README.md
```diff
@@ -367,6 +367,21 @@ CK8S_APPS_PATH= # fill this in
...
+The [`bin/ck8s`](bin/ck8s) script also provides commands to upgrade an environment in two steps `prepare` and `apply`.
+The former runs scripted configuration steps that may not change the state of the environment, while the latter runs scripted upgrade steps that modifies the state of the environment.
+On unexpected failures the command will try to perform a rollback when possible to ensure that the environment continues to function.
+
+```bash
+./bin/ck8s upgrade vX.Y prepare
+./bin/ck8s upgrade vX.Y apply
+```
+
+It is possible to upgrade from one minor version to the next regardless of patch releases.
```
viktor-f, line 381: You could also mention that one can upgrade to a later patch in the same minor version as well. E.g. 0.x.y -> 0.x.z

migration/template/apply/90-apply.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
Xartos, line 1: Should this file be renamed to `80-apply.sh` so all scripts that need to be after apply can be `9x-foobar`? 91-99 should be enough but it feels like it's a nice separation between before apply and after apply

scripts/migration/helmfile.sh
```diff
@@ -0,0 +1,151 @@
...
+
+    if ! helmfile_apply "${1}" "namespace=${2},name=${3}"; then
+      log_error "error: failed to upgrade ${1} ${2}/${3}"
+
+      if [ "${CK8S_ROLLBACK:-}" != "false" ]; then
+        log_warn "  - rolling back ${1} ${2}/${3}"
+
+        if ! helm_do "${1}" rollback --namespace "${2}" "${3}"; then
+          log_error "error: failed to rollback ${1} ${2}/${3}"
+        fi
```
Xartos, line 145: Do we want to signal when this happens so that we can take action? (like send some alert)

```suggestion
      log_error "error: failed to upgrade ${1} ${2}/${3}"

      if [ "${CK8S_ROLLBACK:-}" != "false" ]; then
        log_warn "  - rolling back ${1} ${2}/${3}"

        if ! helm_do "${1}" rollback --namespace "${2}" "${3}"; then
          log_error "error: failed to rollback ${1} ${2}/${3}"
          return 2 # Failed to rollback
        fi
        return 1 # Successful rollback
```

viktor-f (approved):
LGTM
I have not tested it and have not looked in detail at all of the helper functions.

scripts/migration/lib.sh
```diff
@@ -0,0 +1,286 @@
...
+    done
+    ;;
+
+  *)
+    log_fatal "usage: config_validate <secrets|sc|wc>"
+    ;;
+  esac
+
+  if [[ "${pass}" = "false" ]]; then
+    exit 1
```
Xartos, line 140: Do we always want to do a hard stop here or do we want to prompt (as the bin variant does)?

scripts/migration/lib.sh
```diff
@@ -0,0 +1,286 @@
...
+    ;;
+  esac
+}
+
+check_sops() {
+  grep -qs "sops:\\|\"sops\":\\|\\[sops\\]\\|sops_version=" "${1:-/dev/null}"
+}
+
+check_config() {
+  if [ -z "${THIS:-}" ]; then
```
Xartos, line 170: Do we really need this since we have `set -u`? Then it should error out when expanding those

bin/upgrade.bash
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
+
+ROOT="$(readlink -f "$(dirname "${0}")/../")"
+
+CK8S_STACK="$(basename "$0")"
+export CK8S_STACK
```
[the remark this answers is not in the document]
aarnq: It is that way because of shellcheck.
```
In bin/upgrade.bash line 5:
export CK8S_STACK="$(basename "$0")"
       ^--------^ SC2155 (warning): Declare and assign separately to avoid masking return values.
```

migration/v0.29/apply/00-template.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
aarnq: Convenience I guess, I can remove it if you want to.

scripts/migration/helmfile.sh
aarnq: I can add it but it will be highly dependant on how other tools work to capture this special case. I think we will need to work more on these sort of things in general when we continue on this path.

And good that you pointed it out since I totally missed to terminate here on failure after I moved it to change dispatch. :sweat_smile:

scripts/migration/lib.sh
aarnq: I can add some optional prompt, but we have to be careful to disable that in automation.

scripts/migration/lib.sh
aarnq: No, but it looks nicer when things do go wrong.

bin/upgrade.bash
```diff
@@ -0,0 +1,152 @@
...
+    log_info "apply snippet \"${snippet##"${ROOT}/migration/"}\":"
+    if "${snippet}" execute; then
+      log_info "apply snippet success\n---"
+    else
+      log_error "apply snippet failure"
+      if "${snippet}" rollback; then
+        log_warn "apply snippet rollback success"
+        exit 1
+      else
+        log_fatal "apply snippet rollback failure"
```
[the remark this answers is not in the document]
aarnq: It already does but now I have to pass the return value.

scripts/migration/lib.sh
aarnq: Think this should suffice, though I don't know if it will trigger in automation tools. It shouldn't hopefully.

bin/upgrade.bash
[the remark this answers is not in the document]
aarnq: Now it does +1 on failure of execute apply, +1 on success of execute internal rollback and +2 on success of rollback

robinAwallace (approved):

scripts/migration/helmfile.sh
```diff
@@ -0,0 +1,155 @@
...
+  if [[ "${#}" -lt 3 ]] || [[ ! "${2}" =~ ^(sc|wc)$ ]]; then
+    log_fatal "usage: helmfile_change_dispatch <function> <sc|wc> <selectors>..."
+  fi
+
+  local list
+  if ! list="$(helmfile_list "${2}" "${@:3}" 2> /dev/null)"; then
+      log_warn "warning: ${2} ${*:3} had no matching releases"
+      return
+  fi
+  # list="$(yq4 -P '[.[] | select(.enabled and .installed)]' <<< "${list}")"
```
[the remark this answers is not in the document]
aarnq: Pending question:
> Should the removal of old releases be a required migration step?
> 
> - So we don't rely on having them left in the helmfile state, instead we just do a snippet to uninstall it via helm.

PR #1435, opened 2023-02-28
apps wc: Move rolebinding to user-rbac chart

aarnq:

We can't have it where it is right now as it tries to create it before the namespaces exists, and I like to keep everything in helm charts so I came up with a different solution.

It will only try to install it on install/upgrade if it doesn't already exists and it has the annotation to kept so in subsequent runs it's just ignored.

We could employ this trick more to remove the need for the deploy script completely.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


raviranjanelastisys (approved):
LGTM

PR #1443, opened 2023-03-03
Add alerts for dropped packets from/to workloads

### LINKED ISSUE #1307: [1] Alert for blocked network traffic
We have added a lot of network policies recently. If they block traffic then it is likely either
-  something that is miss-configured and should be fixed
- or the environment is compromised and we should investigate.
This is especially true in sc. In wc it might also be that the users are trying to configure their own network policies or are figuring out how to use the platform.

We should add an alert that triggers whenever network policies are blocking traffic. Something similar to `rate(no_policy_drop_counter[1m]) > 1`.

**Additional context**

**Definition of done:**
- There is an alert for dropped traffic.

ayoubeddafali:

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1307 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.


ayoubeddafali:
I didn't find a way to preserve the identifier of the netpol, as the exported metric ( **no_policy_drop_counter** ) doesn't provide that 

Exported metrics:
```go
var dropDesc = prometheus.NewDesc("no_policy_drop_counter", "Number of packets dropped to/from a workload because no policies matched them", []string{
	"pod", // the name of the Kubernetes pod, if any
	"app", // the value of the "app" label of the pod, if any
	"ip", // a comma separated list of ips or subnets associated with the workload
	"type", // fw = from workload, tw = to workload
}, nil)
var acceptDesc = prometheus.NewDesc("policy_accept_counter", "Number of packets accepted by a policy on a workload", []string{
	"pod",
	"app",
	"ip",
	"type",
	"policy", // full name of the accepting calico policy, eg default/knp.default.foo
}, nil)
```

helmfile/charts/prometheus-alerts/values.yaml
```diff
@@ -62,6 +62,7 @@ defaultRules:
     kubeProxy: true
     kubeControllerManager: true
     configReloaders: true
+    networkpolicies: false
```
aarnq, line 65: It would be great if we had a way to enable and disable them via the config.

aarnq (commented):
Add an entry to the wip changelog.

helmfile/charts/prometheus-alerts/templates/alerts/packets-dropped.yaml
```diff
@@ -0,0 +1,42 @@
...
+  labels:
+    app: {{ template "prometheus-alerts.name" . }}
+{{ include "prometheus-alerts.labels" . | indent 4 }}
+{{- if .Values.defaultRules.alertLabels }}
+{{ toYaml .Values.defaultRules.alertLabels | indent 4 }}
+{{- end }}
+{{- if .Values.defaultRules.annotations }}
+  annotations:
+{{ toYaml .Values.defaultRules.annotations | indent 4 }}
+{{- end }}
```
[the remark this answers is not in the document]
ayoubeddafali: @aarnq Are you sure about those indentations ? because with the suggested change I get pasring errors  when running helm template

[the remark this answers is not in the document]
ayoubeddafali: Done. 
I had to use the `nindent` for the `{{- include "prometheus-alerts.labels" . | nindent 4 }}`, otherwise the labels  get mixed up

helmfile/charts/prometheus-alerts/values.yaml
ayoubeddafali: sure, I'll look into that and see if there is a way

helmfile/charts/prometheus-alerts/templates/alerts/packets-dropped.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: {{ printf "%s-%s" (include "prometheus-alerts.fullname" .) "packets-dropped" | trunc 63 | trimSuffix "-" }}
+  labels:
+    app: {{ template "prometheus-alerts.name" . }}
+    {{- include "prometheus-alerts.labels" . | nindent 4 }}
+    {{- if .Values.defaultRules.alertLabels }}
+    {{- toYaml .Values.defaultRules.alertLabels | indent 4 }}
+    {{- end }}
+  {{- if .Values.defaultRules.annotations }}
+  annotations:
+    {{- toYaml .Values.defaultRules.annotations | indent 4 }}
```
aarnq, line 14:
```suggestion
    {{- toYaml .Values.defaultRules.alertLabels | nindent 4 }}
    {{- end }}
  {{- if .Values.defaultRules.annotations }}
  annotations:
    {{- toYaml .Values.defaultRules.annotations | nindent 4 }}
```

I assume these needs to be changed then as well.

helmfile/charts/prometheus-alerts/templates/alerts/packets-dropped.yaml
```diff
@@ -0,0 +1,42 @@
...
+spec:
+  groups:
+  - name: packets-dropped
+    rules:
+    - alert: PacketsDroppedToWorkload
+      annotations:
+        description: Number of packets dropped to a workload because no policies matched them
+        summary: Number of packets dropped to a workload
+      expr: |
+        rate(no_policy_drop_counter{type="tw", cluster=~".*", pod_namespace=~".*", exported_pod=~".*"}[1m]) > 0
```
Xartos, line 25: Since this is a discrete value I like to use `increase` instead of `rate` since that returns the number of increases within that time range and makes a bit more sense in my mind.
```suggestion
        increase(no_policy_drop_counter{type="tw", cluster=~".*", pod_namespace=~".*", exported_pod=~".*"}[1m]) > 0
```

Same for the other one as well.

helmfile/charts/prometheus-alerts/templates/alerts/packets-dropped.yaml
```diff
@@ -0,0 +1,42 @@
...
+      for: 5m
```
Xartos, line 26: I think we want to trigger instantly here (or at least less than the range used in the alert `1m`)

```suggestion
      for: 0m
```

Since otherwise we would not catch the situation where we just have a single package drop each 10 minutes.

ayoubeddafali: Sure thing

robinAwallace: As calico-account will report package drop for unavailable pods as well, i would like this alert to notify us when there are drop packages over a longer period of time. And not instantly.

Xartos: Should we have a compromise with 1 policy matching for a longer period (lets say more than 20 drops over 3 hours with a `for` of 6h) and one that's shorter (more than 0 over 1m (with a `for` of 5m)
Then we would quickly catch big spikes and lower rates drops.

robinAwallace: I like that idea :+1:

ayoubeddafali: @Xartos @robinAwallace I've updated the alerts, can you check if that's what you meant ? Thanks

Xartos: Perfection :ok_hand:

helmfile/charts/prometheus-alerts/templates/alerts/packets-dropped.yaml
```diff
@@ -0,0 +1,60 @@
...
+        summary: Number of packets dropped from a workload
+      expr: |
+        increase(no_policy_drop_counter{type="fw", cluster=~".*", pod_namespace=~".*", exported_pod=~".*"}[3h]) > 20
+      for: 6h
+      labels:
+        severity: info
+
+{{- if .Values.defaultRules.additionalRuleLabels }}
+{{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
+{{- end }}
```
Xartos, line 59: I think this should be for each rule since it's the global extra rule label thing:

```suggestion
      labels:
        severity: info
{{- if .Values.defaultRules.additionalRuleLabels }}
{{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
{{- end }}
```

Like this one https://github.com/elastisys/compliantkubernetes-apps/blob/aeddafali/alert-for-blocked-network-traffic/helmfile/charts/prometheus-alerts/templates/alerts/kubernetes-resources.yaml#L30-L65

ayoubeddafali: Done, thanks a lot 👍🏽

Xartos (approved):

helmfile/charts/prometheus-alerts/values.yaml
ayoubeddafali: Not really, I didn't manage to find a way to do that through the config. 
Do you have any ideas that I can try ?

Xartos: You should be able to do it like we have for some here https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/prometheus-alerts-sc.yaml.gotmpl#L19

ayoubeddafali: It's updated now to read from the config

config/config/common-config.yaml
```diff
@@ -757,3 +757,6 @@ networkPolicies:
 
   dnsAutoscaler:
     enabled: true
+
+  alerts:
+    enabled: false
```
Xartos, line 762: I think we can have it enabled by default

```suggestion
    enabled: true
```

Xartos (approved):

helmfile/charts/prometheus-alerts/templates/alerts/packets-dropped.yaml
```diff
@@ -0,0 +1,68 @@
...
+spec:
+  groups:
+  - name: packets-dropped
+    rules:
+    - alert: FrequentPacketsDroppedToWorkload
+      annotations:
+        description: Number of packets dropped to a workload because no policies matched them
+        summary: Number of packets dropped to a workload
+      expr: |
+        increase(no_policy_drop_counter{type="tw", cluster=~".*", pod_namespace=~".*", exported_pod=~".*"}[1m]) > 0
```
crssnd, line 25: how would increase work if we have a constant number of packets dropped over a long period of time?

crssnd, line 25: if we want everything do we still need `cluster=~".*", pod_namespace=~".*", exported_pod=~".*"`?
what advantage would that bring us?

config/config/common-config.yaml
```diff
@@ -757,3 +757,6 @@ networkPolicies:
 
   dnsAutoscaler:
     enabled: true
+
+  alerts:
```
crssnd, line 761: I find this a litter confusing, as we do not actually deploy netpol to alerts
my proposal:
```
networkPolicies:
  enableAlerting: true
  # Default deny requires Calico as network plugin
  defaultDeny: false
  allowedNameSpaces: []
  additionalEgressPolicies: []
  additionalIngressPolicies: []
```

helmfile/charts/prometheus-alerts/templates/alerts/packets-dropped.yaml
ayoubeddafali: I tried to simulate this case by running curl on an infinite loop with a delay between each request, and it seems like the increase function stays constant also if we have a constant number of dropped packets 

image

ayoubeddafali: Agree, I don't think it necessarily bring any advantages, just I wanted the expr to be clear when reading it. 
I'll change it

helmfile/charts/prometheus-alerts/templates/alerts/packets-dropped.yaml
```diff
@@ -0,0 +1,68 @@
...
+    {{- toYaml .Values.defaultRules.annotations | nindent 4 }}
+  {{- end }}
+spec:
+  groups:
+  - name: packets-dropped
+    rules:
+    - alert: FrequentPacketsDroppedToWorkload
+      annotations:
+        description: Number of packets dropped to a workload because no policies matched them
+        summary: Number of packets dropped to a workload
```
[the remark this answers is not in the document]
ayoubeddafali: Sure

helmfile/charts/prometheus-alerts/templates/alerts/packets-dropped.yaml
ayoubeddafali: @crssnd It should behave as **rate** would, in that case as well 
image

crssnd (approved):

PR #1444, opened 2023-03-03
Apps: opa gatekeeper psp

robinAwallace:
What this PR does / why we need it**:

This is just a WIP/Draft, just to show the general structure

Using OPA gatekeeper PSP to replace kubernetes psp
- Implemented for ingress-nginx and monitoring

Issues:

The current psp constraint from gatekeeper for hostports only allows for one min/max range. This is an issue as ingress-nginx require multiple ports far away from each other. And creating multiple psps for hostports will block each other. 

I had to add gatekeepers mutation for some missing security constraints. But i wished that for supplementary  security constraints such as supplementaryGroups and FSGroups should be allowed to be missing. 

**Which issue this PR fixes**: fixes #

**Checklist:**

    - [x] will change any APIs.
         - Removes kubernetes PodSecurityContext
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


robinAwallace:
I have now created the necessary gatekeeper psps and mutations for ingress nginx and would like to hear your input before continuing with monitoring. 
As you notice it becomes a lot of gatekeeper psp files per kubernetes psp.

robinAwallace:
I am now done with ingress and monitoring. Most stuff could use the restricted psp + mutation and some needed extra permissions.

So the new setup is use restricted and then disable restricted parts for what is needed.

aarnq:
This looks like it will become a pain to maintain with so many specifics than not only are specific templates but also helmfile values.

Would it make more sense to have the helm chart templates generate constraints and then use the helmfile values to define individual "PSP"s that render down to override constraints? Then the helmfile values will become the single source of truth, making it much easier to audit and maintain.

helmfile/values/psp/psp-common.gotmpl
```diff
@@ -0,0 +1,101 @@
...
+    - monitoring
+
+  mutation:
+    labelSelector:
+      matchExpressions:
+        - key: app
+          operator: NotIn
+          values:
+            - ciskubebench-exporter
+            - vulnerability-exporter
```
aarnq, line 24: Why do they need to be exempt?

aarnq:
BTW, does the effective "PSP" become the most or the least restrictive? (Else my suggestion might become a bit harder to realise.)

robinAwallace:
> This looks like it will become a pain to maintain with so many specifics than not only are specific templates but also helmfile values.
> 
> Would it make more sense to have the helm chart templates generate constraints and then use the helmfile values to define individual "PSP"s that render down to override constraints? Then the helmfile values will become the single source of truth, making it much easier to audit and maintain.

After my latest rework i think your idea would be a good next step. Letting the helm values generate the constraints.

> BTW, does the effective "PSP" become the most or the least restrictive? (Else my suggestion might become a bit harder to realise.)

They are the most restrictive and PSA the least restrictive. If that was what you where asking for.

helmfile/values/psp/psp-common.gotmpl
robinAwallace: They are atm running as privileged and required at least root. So one thing that i want to do is to change that they do not run as root etc before doing gatekeeper constraints.

aarnq:
> > BTW, does the effective "PSP" become the most or the least restrictive? (Else my suggestion might become a bit harder to realise.)
> 
> They are the most restrictive and PSA the least restrictive. If that was what you where asking for.

I was think about when you add so multiple constraints apply, as I see you have excluded certain pods from the base restrictive one. With PSAs only one can be applied to a given namespace since they use the same key.

robinAwallace:
Ah, multiple constraints for the same rule does not work. That is why anything that needs more permissions then the restricted constraints needs to be exempted from the restricted constraints.

And the reason why it does not work is that all constraints that apply to a object must be fulfilled.

bootstrap/namespaces/helmfile/values/namespaces-sc.yaml.gotmpl
```diff
@@ -1,27 +1,106 @@
 namespaces:
 - name: cert-manager
+  labels:
+    pod-security.kubernetes.io/audit: baseline
+    pod-security.kubernetes.io/audit-version: v1.25
```
aarnq, line 5: Would it be beneficial to be able to configure the PSA versions? I'm thinking we need these changes before we upgrade from v1.24.

helmfile/charts/gatekeeper/podsecuritypolicies/templates/generic/privilege.yaml
```diff
@@ -0,0 +1,21 @@
+{{- range $namespace, $services := .Values.constraints }}
+  {{- range $name, $value := $services }}
+    {{- if hasKey $value.allow "privilege" }}
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: K8sPSPPrivilegedContainer
+metadata:
+  name: privilege-{{ $name }}
```
aarnq, line 7:
```suggestion
metadata:
  name: privileged-{{ $name }}
```
In order to keep this the same with current PSPs and how SecurityContexts are defined.

helmfile/charts/gatekeeper/podsecuritypolicies/values.yaml
```diff
@@ -0,0 +1,45 @@
...
+#   <service-name>:
+#     labelsSelector:
+#       foo: bar
+#     # Every "allow" is an override, and is optional.
+#     # If a override is NOT part of "allow", the restricted constraint is applied
+#     allow:
+#       allowedCapabilities:
+#         - NET_BIND_SERVICE
+#         - ...
+#       hostPort: <true/false>
```
aarnq, line 12:
```suggestion
#       hostNetwork: <true/false>
```
Should we also include the range for `hostPorts`?

helmfile/charts/gatekeeper/podsecuritypolicies/values.yaml
```diff
@@ -0,0 +1,45 @@
+# Add pods that should override the restricted constraints
+# <namespace>:
+#   <service-name>:
+#     labelsSelector:
```
aarnq, line 4:
```suggestion
#     labelSelector:
```
or
```suggestion
#     selectorLabels:
```
or even
```suggestion
#     podSelectorLabels:
```
These are selected when any match right?

helmfile/charts/gatekeeper/podsecuritypolicies/templates/restricted/mutation/run-as-group.yaml
```diff
@@ -0,0 +1,69 @@
...
+          operator: NotIn
+          values:
+            - {{ $labelvalue | quote }}
+            {{- end }}
+          {{- end }}
+        {{- end }}
+  location: "spec.containers[name: *].securityContext.runAsGroup"
+  parameters:
+    assign:
+      value: 1000
```
aarnq, line 31:
```suggestion
      value: 10000
```
Seems to be a recommendation from certain sources since 1000 could easily exist on the host. Don't know if that would break something thou.

helmfile/values/gatekeeper/podsecuritypolicies.gotmpl
```diff
@@ -0,0 +1,62 @@
...
+      labelsSelector:
+          app.kubernetes.io/component: controller
+      allow:
+        allowedCapabilities:
+          - NET_BIND_SERVICE
+        {{- if .Values.ingressNginx.controller.useHostPort }}
+        hostPort: true
+        {{- end }}
+        hostNamespace: {}
+        privilege: {}
```
aarnq, line 13: I'm thinking just for style that we should set `true` even though it's enough that the key exists.

aarnq (commented):
Really cool that the same style works as for netpolgen. This will make it super simple to setup and maintain new PSPs.

helmfile/charts/gatekeeper/podsecuritypolicies/templates/restricted/mutation/run-as-group.yaml
robinAwallace: Sure, i can try it

helmfile/charts/gatekeeper/podsecuritypolicies/values.yaml
robinAwallace: One issue with gatekeepers HostNetworkPorts constraint is it only allows for one range. 
But we decided that it is okay to open up for 1-65535 as we have network policies.
But yeah maybe it is better to call it `hostNetwork`

bootstrap/namespaces/helmfile/values/namespaces-sc.yaml.gotmpl
robinAwallace: I have not looked that much how these labels work. But it sounds reasonable

helmfile/charts/gatekeeper/podsecuritypolicies/values.yaml
aarnq: Right, I got confused by mixing up the `matchLabels` and `matchExpressions`.

bootstrap/namespaces/helmfile/values/namespaces-sc.yaml.gotmpl
robinAwallace: @robinelastisys as you did some testing with these labels you maybe have some input?

helmfile/charts/gatekeeper/podsecuritypolicies/templates/restricted/mutation/run-as-group.yaml
robinAwallace: I had issue with the exporters i updated `your UID isn't in the passwd file.`
But i can just specify runAsUser: 1000 for it.

bootstrap/namespaces/helmfile/values/namespaces-sc.yaml.gotmpl
```diff
@@ -1,27 +1,106 @@
...
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/audit-version: v1.25
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/enforce-version: v1.25
+    pod-security.kubernetes.io/warn: privileged
+    pod-security.kubernetes.io/warn-version: v1.25
 {{- if .Values.fluentd.enabled }}
 - name: fluentd-system
 {{- end }}
+- name: fluentd-system
```
aarnq, line 30:
```suggestion
{{- if .Values.fluentd.enabled }}
- name: fluentd-system
{{- end }}
```

bootstrap/namespaces/helmfile/values/namespaces-sc.yaml.gotmpl
```diff
@@ -1,27 +1,106 @@
...
 {{ end }}
 {{ if .Values.kured.enabled }}
 - name: kured
+  labels:
+    pod-security.kubernetes.io/audit: baseline
+    pod-security.kubernetes.io/audit-version: v1.25
+    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce-version: v1.25
+    pod-security.kubernetes.io/warn: baseline
+    pod-security.kubernetes.io/warn-version: v1.25
```
aarnq, line 98:
```suggestion
  labels:
    pod-security.kubernetes.io/audit: privileged
    pod-security.kubernetes.io/audit-version: v1.25
    pod-security.kubernetes.io/enforce: privileged
    pod-security.kubernetes.io/enforce-version: v1.25
    pod-security.kubernetes.io/warn: privileged
    pod-security.kubernetes.io/warn-version: v1.25
```

bootstrap/namespaces/helmfile/values/namespaces-sc.yaml.gotmpl
aarnq: Apparently v1.25 works on v1.24. :thinking:

robinAwallace: The version label is apparently optional, defualt is the current kubernetes verison. I think we can remove it. 
Also i think we want to keep the labels `warn` and `audit`
```
Pods are often created indirectly, by creating a workload object such as a Deployment or Job. The workload object defines a Pod template and a controller for the workload resource creates Pods based on that template. To help catch violations early, both the audit and warning modes are applied to the workload resources. However, enforce mode is not applied to workload resources, only to the resulting pod objects.
```

helmfile/14-podsecuritypolicies.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - "{{ requiredEnv "CK8S_CONFIG_PATH" }}/common-config.yaml"
+      - "{{ requiredEnv "CK8S_CONFIG_PATH" }}/sc-config.yaml"
+    secrets:
+      - "{{ requiredEnv "CK8S_CONFIG_PATH" }}/secrets.yaml"
+
+# The desired state of Helm releases.
+releases:
+
+# Gatekeeper PodSecurityPolices
+- name: psp
```
aarnq, line 30:
```suggestion
- name: podsecuritypolicies
```
I'd like this for clarity.

helmfile/50-applications.yaml
```diff
@@ -158,7 +158,8 @@ releases:
     group: starboard
   chart: ./charts/starboard/starboard-operator-psp-rbac
   version: 0.1.0
-  installed: {{ .Values.starboard.enabled }}
+  installed: false
```
aarnq, line 161: The procedure to remove a helm release nowadays is to explicitly remove it via a migration snippet.

aarnq: You should also remove Falco PSP RBAC.

aarnq (commented):
Do we think it would be more clear if we used `custom` and `default` instead of `generic` and `restricted` since some applications can actually use a more restricted PSP than the restricted profile.

At least `generic` to `custom` feels like a better term. And `mutations` in plural since it's more than one.

helmfile/charts/gatekeeper/podsecuritypolicies/templates/generic/volumes.yaml
```diff
@@ -0,0 +1,34 @@
...
+    namespaces:
+      - {{ $namespace }}
+    labelSelector:
+      matchLabels:
+        {{- toYaml $value.labelsSelector | nindent 8 }}
+  parameters:
+    volumes:
+      {{- if $value.volumes }}
+      {{- toYaml $value.allow.volumes | nindent 6 }}
+      {{- else }}
```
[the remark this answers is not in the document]
robinAwallace: Agreed :+1:

helmfile/values/gatekeeper/podsecuritypolicies.gotmpl
```diff
@@ -0,0 +1,62 @@
...
+      labelsSelector:
+          app.kubernetes.io/component: controller
+      allow:
+        allowedCapabilities:
+          - NET_BIND_SERVICE
+        {{- if .Values.ingressNginx.controller.useHostPort }}
+        hostNetwork: true
+        {{- end }}
+        hostNamespace: true
+        privileged: true
```
aarnq, line 13: Are these really privileged? And this renders a constraint against privileged so I don't think it works as you intended. :no_mouth:

helmfile/charts/gatekeeper/podsecuritypolicies/templates/restricted/mutation/run-as-group.yaml
aarnq: I think I have to back-pedal on this due to upstream charts, many assume 1000, won't set it properly and won't allows us to modify it. :no_mouth:

helmfile/values/gatekeeper/podsecuritypolicies.gotmpl
robinAwallace: You are right, I got confused when looking at ingress nginx :sweat_smile: I will fix that

bootstrap/namespaces/helmfile/values/namespaces-sc.yaml.gotmpl
robinAwallace: I removed that version label

helmfile/charts/gatekeeper/podsecuritypolicies/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+#           rule: MustRunAs
+#           ranges:
+#             - min: 1
+#               max: 65535
+#       sysctl:
+#         - ...
+constraints: {}
+
+# List of pod labels that mutation should be exempted from
+mutation: []
```
viktor-f, line 47: Could you make the name a bit more obvious that this excludes things from the mutation? Something like this:
```suggestion
mutationExceptions: []
```

viktor-f (commented):
I have not tested this or carefully checked all of the details, but I think this looks very nice!

robinAwallace:
> Any reason why you set PSA as privileged for all namespaces that should be done separately?

That is the commit i took from a previous task. 
Thought to drop them when this is done.

helmfile/charts/gatekeeper/podsecuritypolicies/values.yaml
robinAwallace: Can do :+1:

config/config/common-config.yaml
```diff
@@ -562,6 +562,62 @@ clusterAdmin:
...
+        memory: 100Mi
+
+  audit:
+    resources:
+      requests:
+        cpu: 200m
+        memory: 100Mi
+      limits:
+        cpu: 500m
+        memory: 200Mi
```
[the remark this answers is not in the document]
robinAwallace: Yes, this could become a problem. But i have not seen it yet.

helmfile/values/gatekeeper/podsecuritypolicies.gotmpl
```diff
@@ -0,0 +1,76 @@
...
+      labelsSelector:
+          app.kubernetes.io/component: controller
+      allow:
+        allowedCapabilities:
+          - NET_BIND_SERVICE
+        {{- if .Values.ingressNginx.controller.useHostPort }}
+        hostNetwork: true
+        {{- end }}
+        hostNamespace: true
+        privilegedEscalation: true
```
aarnq, line 13: Can we finalise how we want the names on keys? Do we want to drop all `allowed` and `allow` given that we are in that map already? Else we should add it back to `allowPrivilegeEscalation`, and can we drop the `d` in that one so it matches what's set in security contexts?

robinAwallace: I will go through them and match them to the security contexts. So adding `allow` drop the `d`

OlleLarsson (approved):
Approving, @aarnq will have the final saying.

migration/v0.30/README.md
```diff
@@ -54,6 +54,33 @@ As with all scripts in this repository `CK8S_CONFIG_PATH` is expected to be set.
...
+    ```bash
+    ./migration/v0.30/apply/12-gatekeeper-psp.sh
+    ```
+
+1. Disable PodSecurityContext for the clusters
+
+    For compliantkubernetes-kubespray you can follow these [steps](https://github.com/elastisys/compliantkubernetes-kubespray/blob/main/migration/v2.20.0-ck8sx-v2.21.0-ck8s1/upgrade-cluster.md) to disable PodSecurityContext.
+
+1. Continue with the `disruptive` step of the upgrade.
+
```
aarnq, line 83:
```suggestion
### Important - PodSecurityPolicies

Apps v0.30 uninstalls the kubernetes PodSecurityPolicies for services running in the clusters.
This release of apps includes PodSecurityAdmission labels and OPA Gatekeepers PodSecurityPolicies and removes the installed PodSecurityPolicies for services.

This means there are some pre-prerequisites that needs to be done before doing the final upgrade apply.

After doing the `non-disruptive` for either the automatic method or manual you should follow these steps before running the `disruptive` step.

1. Add the PSA labels for each namespace

    ```bash
    ./migration/v0.30/apply/10-bootstrap.sh
    ```

1. Install Gatekeeper components

    ```bash
    ./migration/v0.30/apply/12-gatekeeper-psp.sh
    ```

1. Disable PodSecurityPolicies for the clusters

    For compliantkubernetes-kubespray you can follow these [steps](https://github.com/elastisys/compliantkubernetes-kubespray/blob/main/migration/v2.20.0-ck8sx-v2.21.0-ck8s1/upgrade-cluster.md) to disable PodSecurityPolicies.

1. Continue with the `disruptive` step of the upgrade.

```

aarnq (approved):

robinAwallace:
@aarnq @viktor-f @OlleLarsson Please check the new migration step

helmfile/charts/starboard/vulnerability-exporter/templates/deployment.yaml
```diff
@@ -17,14 +17,16 @@ spec:
       serviceAccountName: vulnerability-exporter
       containers:
       - name: metrics-collector
-        image: elastisys/curl-jq:cronjob
+        image: ghcr.io/elastisys/user-crontab:v0.1.0
+        securityContext:
+          allowPrivilegeEscalation: true
+          runAsUser: 1000
+          runAsGroup: 1000
```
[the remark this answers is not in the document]
robinAwallace: I think @viktor-f can answer the falco question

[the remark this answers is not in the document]
viktor-f: Not sure, but you can check if falco is triggering any alerts for the pod.

robinAwallace:
:bell:  Last call before i merge this :bell:

aarnq (approved):
LGTM

PR #1475, opened 2023-03-21
apps: use stageTimestamp as timekey for audit logs

OlleLarsson:
I think it might make more sense to base time from the timestamp for a request based on when it enters a stage.
Previously requests for all stages got the same timestamp.

Let me know what you think

Before
image

After (ignore the requestReceived stage)
image

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/WIP-CHANGELOG.md
+++ b/WIP-CHANGELOG.md
@@ -26,5 +26,6 @@
 - grafana to v9.3.8
 
 ### Changed
+- Changed timekey to `stageTimestamp` for Kubernetes audit logs. Use `auditID` to correlate stages of the same request.
 
 ### Removed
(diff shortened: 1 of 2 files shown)


Xartos (approved):
Yea this makes sense

cristiklein:
@OlleLarsson Awesome PR. The screenshots make it dead simple to understand why we want this PR.

My only concern is: How would we correlate stages belonging to the same request? Before this PR, we could simply look at timestamp. How do we do that after this PR?

WIP-CHANGELOG.md
```diff
@@ -26,5 +26,6 @@
 - grafana to v9.3.8
 
 ### Changed
+- Changed timekey to `stageTimestamp` for Kubernetes audit logs
```
cristiklein, line 29:
```suggestion
- Changed timekey to `stageTimestamp` for Kubernetes audit logs. Use `auditID` to correlate stages of the same request.
```

cristiklein (approved):
LGTM!

PR #1499, opened 2023-04-05
apps wc: Make user-rbac extra-workload-admins idempotent

aarnq:
Else diff will report it as new all the time.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.

### DIFF
--- a/helmfile/charts/user-rbac/templates/rolebindings/workload-admin.yaml
+++ b/helmfile/charts/user-rbac/templates/rolebindings/workload-admin.yaml
@@ -20,8 +20,9 @@ subjects:
   kind: Group
   name: {{ $group }}
 {{- end }}
+{{- $known := lookup "v1" "Service" "default" "kubernetes" }}
 {{- $value := lookup "rbac.authorization.k8s.io/v1" "RoleBinding" $namespace "extra-workload-admins" }}
-{{- if and (or $.Release.IsInstall $.Release.IsUpgrade) (not $value) }}
+{{- if and $known (not $value) (or $.Release.IsInstall $.Release.IsUpgrade) }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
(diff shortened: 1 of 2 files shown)


helmfile/charts/user-rbac/templates/rolebindings/workload-admin.yaml
```diff
@@ -20,8 +20,9 @@ subjects:
   kind: Group
   name: {{ $group }}
 {{- end }}
+{{- $known := lookup "v1" "Service" "default" "kubernetes" }}
 {{- $value := lookup "rbac.authorization.k8s.io/v1" "RoleBinding" $namespace "extra-workload-admins" }}
-{{- if and (or $.Release.IsInstall $.Release.IsUpgrade) (not $value) }}
+{{- if or $.Release.IsInstall $.Release.IsUpgrade | and $known (not $value) }}
```
Xartos, line 25: How does the `|` operator work in this situation? Like does the `and` function have input?

aarnq: Well let me tell you cause I've finally read up on how the pipe operator works in helm :sweat_smile:

When you use a pipe the result of the expression will be used as the **last** argument for the next expression so this one would be equivalent to `and $known (not $value) (or $.Release.IsInstall $.Release.IsUpgrade)`. And this applies to all functions.

Xartos: Righ, yea that kinda makes sense. But would it maybe be a bit easier to read if we changed it to the equivalent expression in that case? At least for me that's more readable.

OlleLarsson (approved):

raviranjanelastisys (approved):

Xartos (approved):

PR #1511, opened 2023-04-12
apps: fix kube-prometheus-stack migration script

### LINKED ISSUE #1496: [1] Snippet remove kube-prometheus-components is not idempotent.
**Describe the bug**
[This snippet](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration/v0.30/apply/20-remove-kube-prometheus-components.sh) is not idempotent in the way new upgrades should be and can potentially leave environment with unmanaged resources.

**To Reproduce**
1. Run upgrade
2. Run upgrade again
3. Missing daemonset and deployment

**Expected behavior**
It should only delete the daemonset and deployment if it isn't the new version.

 - Compliant Kubernetes apps version main and what will be v0.30.

crssnd:
**What this PR does / why we need it**: the script to remove monitoring components should run only when needed

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1496 

- the `app=prometheus-node-exporter` label does not exists in the new version of node-exporter, so the daemonset will not be removed if we re-run the upgrade scripts

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/migration/v0.30/apply/20-remove-kube-prometheus-components.sh
+++ b/migration/v0.30/apply/20-remove-kube-prometheus-components.sh
@@ -9,11 +9,27 @@ run() {
   case "${1:-}" in
   execute)
     for CLUSTER in sc wc; do
-      log_info "  - removing kube-state-metrics from $CLUSTER"
-      kubectl_do $CLUSTER delete deployments.apps -l app.kubernetes.io/instance=kube-prometheus-stack,app.kubernetes.io/name=kube-state-metrics --cascade=orphan -n monitoring
-      log_info "  - removing prometheus-node-exporter from $CLUSTER"
-      kubectl_do $CLUSTER delete daemonset -l app=prometheus-node-exporter -n monitoring
+      old_state_metrics=$(kubectl get deployments.apps -n monitoring -l app.kubernetes.io/instance=kube-prometheus-stack,app.kubernetes.io/name=kube-state-metrics,app.kubernetes.io/version!=2.8.0 --output=jsonpath={.items..metadata.name})
+      old_node_exporter=$(kubectl get daemonset -n monitoring -l app=prometheus-node-exporter --output=jsonpath={.items..metadata.name})
+
+      if [[ -n "${old_state_metrics}" ]]; then
+        log_info "- removing kube-state-metrics from $CLUSTER"
+        kubectl_do $CLUSTER delete deployments.apps -l app.kubernetes.io/instance=kube-prometheus-stack,app.kubernetes.io/name=kube-state-metrics,app.kubernetes.io/version!=2.8.0 --cascade=orphan -n monitoring
+      fi
+
+      if [[ -n "${old_node_exporter}" ]]; then
+        log_info "- removing prometheus-node-exporter from $CLUSTER"
+        kubectl_do $CLUSTER delete daemonset -l app=prometheus-node-exporter -n monitoring
+      fi
     done
+
+    # --- sc ---
+    helmfile_upgrade sc app=kube-prometheus-stack
+    helmfile_upgrade sc app=thanos
+
+    # --- wc ---
+    helmfile_upgrade wc app=kube-prometheus-stack
+    helmfile_upgrade wc app=user-alertmanager
     ;;
   rollback)
     log_warn "rollback not implemented"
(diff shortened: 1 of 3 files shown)


migration/v0.30/apply/20-remove-kube-prometheus-components.sh
```diff
@@ -10,7 +10,7 @@ run() {
   execute)
     for CLUSTER in sc wc; do
       log_info "  - removing kube-state-metrics from $CLUSTER"
-      kubectl_do $CLUSTER delete deployments.apps -l app.kubernetes.io/instance=kube-prometheus-stack,app.kubernetes.io/name=kube-state-metrics --cascade=orphan -n monitoring
+      kubectl_do $CLUSTER delete deployments.apps -l app.kubernetes.io/instance=kube-prometheus-stack,app.kubernetes.io/name=kube-state-metrics,app.kubernetes.io/version!=2.8.0 --cascade=orphan -n monitoring
       log_info "  - removing prometheus-node-exporter from $CLUSTER"
       kubectl_do $CLUSTER delete daemonset -l app=prometheus-node-exporter -n monitoring
```
aarnq, line 15: This one as well.

I think I rather would like to see it do a check and then delete, so that the script doesn't print out that it removes something when it doesn't.

crssnd: for the DS the label `app=prometheus-node-exporter` does not exists in the newer version, so I think it is safe to keep it as it is

aarnq: I don't like this as the approach for snippets, as such things are not clear.

I think I would prefer that we have a way to pull the version of installed charts and take actions accordingly. I'm fine with letting it be like this for this release and I'll add functions in the scripts library to do so for next release.

Also the upgrade of kube-prometheus-stack should really all be contained here.

crssnd: I think I fixed everything we discussed. let me know if I missed anything

migration/v0.30/apply/20-remove-kube-prometheus-components.sh
```diff
@@ -2,18 +2,35 @@
...
     for CLUSTER in sc wc; do
-      log_info "  - removing kube-state-metrics from $CLUSTER"
-      kubectl_do $CLUSTER delete deployments.apps -l app.kubernetes.io/instance=kube-prometheus-stack,app.kubernetes.io/name=kube-state-metrics --cascade=orphan -n monitoring
-      log_info "  - removing prometheus-node-exporter from $CLUSTER"
-      kubectl_do $CLUSTER delete daemonset -l app=prometheus-node-exporter -n monitoring
+      old_state_metrics=$(kubectl get deployments.apps -n monitoring -l app.kubernetes.io/instance=kube-prometheus-stack,app.kubernetes.io/name=kube-state-metrics,app.kubernetes.io/version!=2.8.0 --output=jsonpath={.items..metadata.name})
+      old_node_exporter=$(kubectl get daemonset -n monitoring -l app=prometheus-node-exporter --output=jsonpath={.items..metadata.name})
+
+      if [[ -n "${old_state_metrics}" ]]; then
+        log_info "  - removing kube-state-metrics from $CLUSTER"
```
aarnq, line 17:
```suggestion
        log_info "- removing kube-state-metrics from $CLUSTER"
```

migration/v0.30/apply/20-remove-kube-prometheus-components.sh
```diff
@@ -2,18 +2,35 @@
...
+        kubectl_do $CLUSTER delete deployments.apps -l app.kubernetes.io/instance=kube-prometheus-stack,app.kubernetes.io/name=kube-state-metrics,app.kubernetes.io/version!=2.8.0 --cascade=orphan -n monitoring
+      fi
+
+      if [[ -n "${old_node_exporter}" ]]; then
+        log_info "  - removing prometheus-node-exporter from $CLUSTER"
```
aarnq, line 22:
```suggestion
        log_info "- removing prometheus-node-exporter from $CLUSTER"
```

aarnq (approved):

lunkan93 (approved):

PR #1532, opened 2023-04-19
apps: add gatekeeper psps for rook-ceph

### LINKED ISSUE #1451: [2] Replace PSPs for rook-ceph
We have now investigated both PSA and gatekeeper as a replacement for PSPs. We have decided that we want to try to replace it with a combination of both.

The idea is to use PSA for coarse grained control and gatekeeper for fine grained control. The PSA used in a namespace have to match the most privileged pod in that namespace, then we use gatekeeper to lock everything down as much as possible.

Use https://github.com/elastisys/compliantkubernetes-apps/pull/1444 as a basis.

**Additional context**

**Definition of done:**

- We have replaced all PSPs for rook-ceph.

lunkan93:
add gatekeeper psps for rook-ceph
**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1451 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/config/config/common-config.yaml
+++ b/config/config/common-config.yaml
@@ -810,3 +810,7 @@ networkPolicies:
 
   dnsAutoscaler:
     enabled: true
+
+# Gatekeeper PSP for rook-ceph enabled/disabled
+rookCeph:
+  enabled: false
--- a/helmfile/charts/gatekeeper/podsecuritypolicies/templates/default/mutations/privilege-escalation.yaml
+++ b/helmfile/charts/gatekeeper/podsecuritypolicies/templates/default/mutations/privilege-escalation.yaml
@@ -1,6 +1,15 @@
 {{- $namespaceSelectorLabels := .Values.mutations.namespaceSelectorLabels }}
 {{- $namespaces := keys .Values.constraints | sortAlpha }}
 {{- $exceptions := .Values.mutations.exceptions }}
+
+{{- range $namespace, $services := .Values.constraints }}
+{{- range $name, $value := $services }}
+{{- if $value | dig "allow" "allowPrivilegeEscalation" false }}
+{{- $exceptions = append $exceptions $value.podSelectorLabels }}
+{{- end }}
+{{- end }}
+{{- end }}
+---
 ---
 apiVersion: mutations.gatekeeper.sh/v1
 kind: Assign
--- a/helmfile/values/podsecuritypolicies/common/rook-ceph.yaml.gotmpl
+++ b/helmfile/values/podsecuritypolicies/common/rook-ceph.yaml.gotmpl
@@ -1,0 +1,237 @@
+{{- if .Values.rookCeph.enabled }}
+constraints:
+  rook-ceph:
+
+    operator:
+      podSelectorLabels:
+        app: rook-ceph-operator
+      allow:
+        privileged: true
+        allowPrivilegeEscalation: true
+        volumes:
+          - emptyDir
+          - projected
+
+    csi-rbdplugin-provisioner:
+      podSelectorLabels:
+        app: csi-rbdplugin-provisioner
+      allow:
+        privileged: true
+        allowPrivilegeEscalation: true
+        volumes:
+          - emptyDir
+          - projected
+          - hostPath
+        allowedHostPaths:
+          - pathPrefix: "/dev"
+          - pathPrefix: "/sys"
+          - pathPrefix: "/lib/modules"
+        runAsUser:
+          rule: RunAsAny
+        runAsGroup:
+          rule: RunAsAny
+        supplementalGroups:
+          rule: RunAsAny
+        fsGroup:
+          rule: RunAsAny
+      mutation:
+        dropAllCapabilities: false
+
+    csi-rbdplugin:
+      podSelectorLabels:
+        app: csi-rbdplugin
+      allow:
+        privileged: true
+        allowPrivilegeEscalation: true
+        hostNetworkPorts: true
+        hostNamespace: true
+        volumes:
+          - emptyDir
+          - projected
+          - hostPath
+        allowedHostPaths:
+          - pathPrefix: "/run/udev"
+          - pathPrefix: "/dev"
+          - pathPrefix: "/var/lib/rook"
+          - pathPrefix: "/var/lib/kubelet/plugins_registry"
+          - pathPrefix: "/var/lib/kubelet/plugins"
+          - pathPrefix: "/var/lib/kubelet/pods"
+          - pathPrefix: "/sys"
+          - pathPrefix: "/run/mount"
+          - pathPrefix: "/lib/modules"
+        runAsUser:
+          rule: RunAsAny
+        runAsGroup:
+          rule: RunAsAny
+        supplementalGroups:
+          rule: RunAsAny
+        fsGroup:
+          rule: RunAsAny
+      mutation:
+        dropAllCapabilities: false
+
+    crashcollector:
+      podSelectorLabels:
+        app: rook-ceph-crashcollector
+      allow:
+        allowPrivilegeEscalation: true
+        volumes:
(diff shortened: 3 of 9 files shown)


helmfile/values/podsecuritypolicies/common/rook-ceph.yaml.gotmpl
```diff
@@ -0,0 +1,234 @@
...
+        runAsUser:
+          rule: RunAsAny
+        runAsGroup:
+          rule: RunAsAny
+        supplementalGroups:
+          rule: RunAsAny
+        fsGroup:
+          rule: RunAsAny
+        allowedCapabilities:
+          - SYS_ADMIN
```
OlleLarsson, line 36: Do we need to touch anything with capabilities as we are running a privileged container?
Privileged should lift any restrictions there are on capabilities.

lunkan93: This was mostly to avoid violating the constraint `capabilities-restricted` which requires containers to drop all capabilities if none are specified under allowedCapabilities.

helmfile/values/podsecuritypolicies/common/rook-ceph.yaml.gotmpl
```diff
@@ -0,0 +1,234 @@
...
+mutationExceptions:
+  - app: csi-rbdplugin
+  - app: csi-rbdplugin-provisioner
+  - app: rook-ceph-mgr
+  - app: rook-ceph-mon
+  - app: rook-ceph-osd
+  - app: rook-ceph-detect-version
+  - app: rook-ceph-csi-detect-version
+  - app: rook-ceph-osd-prepare
```
robinAwallace, line 233: What mutation was causing the issue?

robinAwallace: Wonder if it is adding a override for that mutation, so not all mutations are overridden.

robinAwallace (commented):
Its only the mutation exception im wondering about, other then that it LGTM

helmfile/values/podsecuritypolicies/common/rook-ceph.yaml.gotmpl
lunkan93: I did not find much success when trying to override mutations for these pods, they would either be stuck crashing, or for at least some of the pods there were issues with supplemental group in terms of violating constraints.

robinAwallace (approved):

Xartos (approved):
LGTM!

config/config/common-config.yaml
```diff
@@ -834,5 +834,9 @@ networkPolicies:
   dnsAutoscaler:
     enabled: true
 
+# Gatekeeper PSP for rook-ceph enabled/disabled
+rookCeph:
```
crssnd, line 838: we should enable this we we run init, if the cloud provider is exoscale
similar with what we do [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/init.bash#L152)

lunkan93: Added this

helmfile/values/podsecuritypolicies/common/rook-ceph.yaml.gotmpl
OlleLarsson: Should be able to change this once [these](https://github.com/elastisys/compliantkubernetes-apps/pull/1559/files) changes are merged

helmfile/values/podsecuritypolicies/common/rook-ceph.yaml.gotmpl
```diff
@@ -0,0 +1,242 @@
...
+          rule: RunAsAny
+        supplementalGroups:
+          rule: RunAsAny
+        fsGroup:
+          rule: RunAsAny
+        allowedCapabilities: []
+      mutation:
+        dropAllCapabilities: false
+mutationExceptions:
+  - app: csi-rbdplugin
```
robinAwallace, line 234: I think you should take an extra look at these, to see if you could fix them.

helmfile/values/podsecuritypolicies/common/rook-ceph.yaml.gotmpl
```diff
@@ -0,0 +1,226 @@
+{{- if .Values.rookCeph.enabled }}
+constraints:
+  rook-ceph:
+    rook-ceph-operator:
```
robinAwallace, line 4: You can drop all rook-ceph for services, as they are under the rook-ceph namespace
```suggestion
    operator:
```

robinAwallace (approved):
LGTM :+1: Just remember to squash the commits

helmfile/charts/gatekeeper/podsecuritypolicies/templates/default/mutations/privilege-escalation.yaml
```diff
@@ -1,6 +1,15 @@
 {{- $namespaceSelectorLabels := .Values.mutations.namespaceSelectorLabels }}
 {{- $namespaces := keys .Values.constraints | sortAlpha }}
 {{- $exceptions := .Values.mutations.exceptions }}
+
+{{- range $namespace, $services := .Values.constraints }}
+{{- range $name, $value := $services }}
+{{- if $value | dig "allow" "allowPrivilegeEscalation" "true" }}
```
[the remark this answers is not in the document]
robinAwallace: It should be true right? As this mutation should not run if `allow.allowPrivilegeEscalation` is set to true. As we do not want it to add `false`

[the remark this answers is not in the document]
aarnq: No, since the `false` here is the default value if `allow.allowPrivilegeEscalation` is unset.
So having it as `true` will always skip the mutation unless it is explicitly set to `false`.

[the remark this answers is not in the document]
aarnq: Though come to think of it, the `"false"` should be `false` else it is will still evaluate to true.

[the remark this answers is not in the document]
lunkan93: My understanding is the following:
1. If `allowPrivilegeEscalation` is set to `true`, add it as an exception so that it does not get mutated to false by this `assign`. (This was causing issues because the custom mutation would not set it to `true` as it was already set to `false` by this one from what I could tell. This was the reason for this change in the first place.) 
2. Otherwise, do not add this as an exception and let this `assign` set it to `false` regardless if you had it unset or explicitly set to `false`.

Please correct me if I'm completely missing something :smile:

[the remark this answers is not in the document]
aarnq: The problem here is that `$value | dig "allow" "allowPrivilegeEscalation" "true"` will evaluate to `"true"` if `.allow.allowPrivilegeEscalation` is unset. (The last arg passed to `dig`, not counting piped `$value`, is the default value it will evaluate to if it can't find a value.)
So that exception will always be added if `allowPrivilegeEscalation` isn't explicitly set to `false`, and that is not what you intend right?.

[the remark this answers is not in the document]
lunkan93: Aaah now I see what you're saying, thanks for clarifying

aarnq (approved):
nit, else it looks good to me

PR #1536, opened 2023-04-20
apps: Expose all `fluentd-aggregator` buffer settings

aarnq:

So we can tune buffer settings to cope with higher log volumes.

Should I do this for forwarder as well?

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.

### DIFF
--- a/config/config/common-config.yaml
+++ b/config/config/common-config.yaml
@@ -175,7 +175,23 @@ fluentd:
       #   </match>
 
   aggregator:
-    buffer: {}
+    # All buffer settings are available, keys are transformed to snake_case as required by fluentd config
+    buffer:
+      chunkLimitSize: 50MB
+      totalLimitSize: 9GB
+
+      flushMode: interval
+      flushInterval: 15m
+      flushThreadCount: 4
+      flushThreadBurstInterval: 0.2
+
+      retryType: exponential_backoff
+      retryForever: true
+      retryMaxInterval: 30
+
+      timekey: 10m
+      timekeyWait: 1m
+      timekeyUseUtc: true
 
     persistence:
       storage: 10Gi
@@ -185,7 +201,7 @@ fluentd:
         cpu: 300m
         memory: 500Mi
       limits:
-        cpu: 500m
+        cpu: 750m
         memory: 1000Mi
 
     tolerations: []
--- a/helmfile/values/fluentd/aggregator-common.yaml.gotmpl
+++ b/helmfile/values/fluentd/aggregator-common.yaml.gotmpl
@@ -132,13 +132,9 @@ aggregator:
     buffer.prop: |
       @type file
 
-      overflow_action {{ .Values.fluentd.aggregator.buffer | get "overflowAction" "block" }}
-      chunk_limit_size {{ .Values.fluentd.aggregator.buffer | get "chunkLimitSize" "50MB" }}
-      total_limit_size {{ .Values.fluentd.aggregator.buffer | get "totalLimitSize" "9GB" }}
-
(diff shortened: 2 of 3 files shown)


aarnq:
I will test a bit with less frequent interval as it seems like it doesn't care if the chunk *should* be sent.

config/config/common-config.yaml
```diff
@@ -175,7 +175,22 @@ fluentd:
...
 
   aggregator:
-    buffer: {}
+    # All buffer settings are available, keys are transformed to snake_case as required by fluentd config
+    buffer:
+      chunkLimitSize: 50MB
+      totalLimitSize: 9GB
+
+      flushMode: interval
+      flushInterval: 5s
```
OlleLarsson, line 184: So we can at most loose 5s worth of logs if fluentd were to crash?

helmfile/values/fluentd/aggregator-common.yaml.gotmpl
```diff
@@ -127,13 +127,9 @@ aggregator:
...
 
-      overflow_action {{ .Values.fluentd.aggregator.buffer | get "overflowAction" "block" }}
-      chunk_limit_size {{ .Values.fluentd.aggregator.buffer | get "chunkLimitSize" "50MB" }}
-      total_limit_size {{ .Values.fluentd.aggregator.buffer | get "totalLimitSize" "9GB" }}
-
-      timekey {{ .Values.fluentd.aggregator.buffer | get "timekey" "10m" }}
-      timekey_wait {{ .Values.fluentd.aggregator.buffer | get "timekeyWait" "1m" }}
-      timekey_use_utc true
+      {{- range $key, $value := .Values.fluentd.aggregator.buffer }}
+      {{ $key | snakecase }} {{ $value }}
```
OlleLarsson, line 131: Will this indent properly?

config/config/common-config.yaml
aarnq: This is when it pushes to S3, so at most 5s if the PVC is lost, so this is way to low given how frequent it would push to S3.

helmfile/values/fluentd/aggregator-common.yaml.gotmpl
aarnq: Yes

From:
```yaml

fluentd:
  aggregator:
    buffer:
      chunkLimitSize: 100MB
      totalLimitSize: 9GB
      flushMode: interval
      flushInterval: 5s
      flushThreadCount: 4
      retryType: exponential_backoff
      retryForever: true
      retryMaxInterval: 30
      timekey: 10m
      timekeyWait: 1m
      timekeyUseUtc: true
```

To:
```yaml
data:
  buffer.prop: |
    @type file
    chunk_limit_size 100MB
    flush_interval 5s
    flush_mode interval
    flush_thread_count 4
    retry_forever true
    retry_max_interval 30
    retry_type exponential_backoff
    timekey 10m
    timekey_use_utc true
    timekey_wait 1m
    total_limit_size 9GB
```

config/config/common-config.yaml
aarnq: Set it to 15m now, it will still push when the chunk is read after ~10m, so this is in case that doesn't happen.

OlleLarsson: But isn't flushing just for persisting the buffer on disk in our case?

aarnq: No the flushing here is done to the output.

raviranjanelastisys (approved):

davidumea (approved):

OlleLarsson (approved):

PR #1549, opened 2023-04-26
apps wc: add gatekeeper psps for user fluentd

OlleLarsson:

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.

### DIFF
--- a/helmfile/values/podsecuritypolicies/common/fluentd.yaml.gotmpl
+++ b/helmfile/values/podsecuritypolicies/common/fluentd.yaml.gotmpl
@@ -28,7 +28,7 @@ constraints:
   fluentd:
     fluentd-forwarder-user:
       podSelectorLabels:
-        app.kubernetes.io/instance: fluentd-forwarder
+        app.kubernetes.io/instance: fluentd
       allow:
         allowPrivilegeEscalation: true
         allowedCapabilities:
@@ -50,4 +50,8 @@ constraints:
           - hostPath
           - projected
   {{- end }}
+mutations:
+  exceptions:
+    - app.kubernetes.io/instance: fluentd
+    - app.kubernetes.io/instance: fluentd-forwarder
 {{- end }}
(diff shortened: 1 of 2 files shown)


OlleLarsson:
need to add so that ALL capabilities are dropped.

aarnq:
You can just correct the selector label of the one above.

helmfile/values/podsecuritypolicies/common/fluentd.yaml.gotmpl
```diff
@@ -50,4 +50,8 @@ constraints:
           - hostPath
           - projected
   {{- end }}
+mutations:
+  exceptions:
+    - app.kubernetes.io/instance: fluentd
+    - app.kubernetes.io/instance: fluentd-forwarder
```
aarnq, line 56: I've been testing this a bit and do you think it would be better to have an implicit removal of `drop: [ALL]` if we set allowed capabilities, or to add a per service mutation exception to skip adding `drop: [ALL]`? 
(I would want to avoid setting wide exceptions, though it is better now that these are limited to certain namespaces.)

OlleLarsson: I first thought of implicit removal if one has configured for example `allow.allowedCapabilities`, but now I'm leaning towards the option of disabling mutations per service level.

helmfile/values/podsecuritypolicies/common/velero.yaml.gotmpl
```diff
@@ -17,4 +17,12 @@ constraints:
...
         privileged: true
         allowPrivilegeEscalation: true
+    velero:
+      podSelectorLabels:
+        name: velero
+      allow:
+        runAsUser:
+          rule: MustRunAsNonRoot
+      mutation:
+        runAsUser: 1000
```
aarnq, line 27: Why not set the proper config on Veleros security context?

PR #1566, opened 2023-05-02
Linus/log manager ephemeral volumes

### LINKED ISSUE #1537: Allow log-manager compaction to use ephemeral volumes
Log-managers compaction fetches, merges, and compresses per tag per day, and one of those days may be quite large which cannot be served by nodes with restrictive local storage.

Allow log-manager compaction to use [ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/) to alleviate the problem.

Limit block size, but it will complicate the process.

**Additional context**
- There might be some consideration if all jobs should fall under the same option or if separate config options should be set.
- There might be some consideration to scheduling so there won't be high storage pressure all at the same time.

**Definition of done:**
Log-manager compaction can run with large blocks on nodes with limited storage.

linus-astrom:
This adds so log-manager compaction can use ephemeral volumes. / So that log-manager compaction effectively can run on nodes with limited storage.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1537

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/helmfile/charts/log-manager/templates/cronjob-compaction.yaml
+++ b/helmfile/charts/log-manager/templates/cronjob-compaction.yaml
@@ -28,6 +28,9 @@ spec:
             - name: compaction
               imagePullPolicy: {{ $.Values.image.pullPolicy }}
               image: {{ $.Values.image.repository }}:{{ default $.Chart.AppVersion $.Values.image.tag }}
+              {{- if .compaction.ephemeralVolumes.enabled }}
+              workingDir: /scratch
+              {{- end }}
               command: [/scripts/compaction.sh]
               env:
                 - name: S3_CONFIG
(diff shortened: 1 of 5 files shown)


aarnq (commented):
One thing, else I think it looks good.

helmfile/charts/log-manager/templates/cronjob-compaction.yaml
```diff
@@ -28,6 +28,7 @@ spec:
             - name: compaction
               imagePullPolicy: {{ $.Values.image.pullPolicy }}
               image: {{ $.Values.image.repository }}:{{ default $.Chart.AppVersion $.Values.image.tag }}
+              workingDir: /scratch
```
aarnq, line 31: This might need to be conditional as well, or does it work properly without the ephemeral volume enabled?

aarnq (approved):

robinAwallace (approved):

PR #1542, opened 2023-04-24
apps sc: Add Prometheus Blackbox Exporter dashboard (fixes #1437)

Zash:
Adapted from https://grafana.com/grafana/dashboards/7587-prometheus-blackbox-exporter/

**What this PR does / why we need it**: #1437 

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1437

**Public facing documentation PR** *(if applicable)*

**Special notes for reviewer**:

**Add a screenshot or an example to illustrate the proposed solution:**

image

**Checklist:**

- [x] Added relevant notes to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md)
- [x] Proper commit message prefix on all commits
- [ ] Updated the [public facing documentation](https://github.com/elastisys/compliantkubernetes)
- Is this changeset backwards compatible for existing clusters? Applying:
    - [x] is completely transparent, will not impact the workload in any way.
    - [ ] requires running a migration script.
    - [ ] will create noticeable cluster degradation.
          E.g. logs or metrics are not being collected or Kubernetes API server
          will not be responding while upgrading.
    - [ ] requires draining and/or replacing nodes.
    - [ ] will change any APIs.
          E.g. removes or changes any CK8S config options or Kubernetes APIs.
    - [ ] will break the cluster.
          I.e. full cluster migration is required.
- Chart checklist (pick exactly one):
    - [ ] I upgraded no Chart.
    - [x] I upgraded a Chart and determined that no migration steps are needed.
    - [ ] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).

**Pipeline config** *(if applicable)*
If you change some config options (e.g. add/rename variable or change the default value) you may need to update the config used by the pipeline in `pipeline/config`.

PR #1572, opened 2023-05-04
Fix Fluentd probe

Zash:

See screenshots and https://github.com/elastisys/compliantkubernetes-apps/pull/1542#issuecomment-1521382910

Related to #1542 

**Before**

image

**After**

image

Further investigation needed to find out whether the probe is correct about fluentd being down. Could also be more wrong with the probe URL.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.

### DIFF
--- a/helmfile/values/prometheus-blackbox-exporter-sc.yaml.gotmpl
+++ b/helmfile/values/prometheus-blackbox-exporter-sc.yaml.gotmpl
@@ -45,7 +45,7 @@ serviceMonitor:
     {{- end }}
     {{- if .Values.prometheusBlackboxExporter.targets.fluentd }}
     - name: fluentd
-      url: fluentd-aggregator-aggregator.fluentd-system.svc.cluster.local:9880/fluentd.healthcheck?json={"ping","pong"}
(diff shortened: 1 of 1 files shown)


aarnq (approved):

Zash:
> Further investigation needed to find out whether the probe is correct about fluentd being down.

It's down, the probe is correct. Good to have that part of the dashboard tested :)

Zash:
Furthermore, is Fluent running in the service cluster at all? It seems to be in the workload cluster, so a probe on the cluster-internal name from the *sc* seems unlikely to work. Finally, I'm not sure if the URL is even correct, running the same probe in the *wc* reports **DOWN** as well.

aarnq:
> Furthermore, is Fluent running in the service cluster at all? It seems to be in the workload cluster, so a probe on the cluster-internal name from the _sc_ seems unlikely to work. Finally, I'm not sure if the URL is even correct, running the same probe in the _wc_ reports **DOWN** as well.

Yes, if you have fluentd enabled in the SC cluster and at least one of audit or scLogs enabled.

Zash:
> Yes, if you have fluentd enabled in the SC cluster and at least one of audit or scLogs enabled.

Turns out I had (via defaults) `fluentd.enabled = false`...

crssnd (approved):

PR #1576, opened 2023-05-08
apps sc: Run log-manager compact more frequently

aarnq:

Since the defaults for fluentd will cause a lot of chunks we should compact them more often so it won't need several hours each time.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -998,11 +998,11 @@ fluentd:
       enabled: true
       ephemeralVolumes:
         enabled: false
-      schedule: 0 22 * * *
+      schedule: 0 1/6 * * *
       days: 10
     retention:
       enabled: true
-      schedule: 0 23 * * *
+      schedule: 0 20 * * *
       days: 30 # For HIPAA/PDL 1825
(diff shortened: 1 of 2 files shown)


Pavan-Gunda (approved):

config/config/sc-config.yaml
```diff
@@ -998,11 +998,11 @@ fluentd:
       enabled: true
       ephemeralVolumes:
         enabled: false
-      schedule: 0 22 * * *
+      schedule: 0 21/6 * * *
```
crssnd, line 1001: will this run every 6th hour between 21 to 23?

aarnq: Yeah, I will change it as it didn't work as I expected.

aarnq: Now it's changed, also moved the window for when it does it so it is well before the default sync schedule, and set so the retention runs one hour after the last run each day.

crssnd (approved):

PR #1582, opened 2023-05-09
apps sc & wc: private subnet as node-ips

### LINKED ISSUE #1453: [2] Update the update-ips script to know about Octavia and to allow anchors/overrides
It is confusing to get warnings about IPs that should exist in the IP lists managed by the update-ips script, since they may be there for a reason but you will not see it unless you manually check either comments in the file or why the got added.
One such point of confusion is the use of Octavia load balancers, since it will want to remove those IPs for every check that is done.

I would like it to be aware of when Octavia is in use so it can prompt to add the private subnet instead of adding specific IPs.

Alternatives would be to perhaps preserve comments to show why certain IPs differ or to ignore values with comments when considering the diff, perhaps both for added visibility.
Third alternative would be to just keep override lists, but then it feels like we need to improve our handling of netpol configs in general because it would be too much to maintain in values templates.

(I'm almost more in favour of the second alternative as it would allow us to anchor additions we know we want to keep.)

**Definition of done:**
The update-ips script is more friendly to use when you need to override IPs.

### LINKED ISSUE #1677: Rework/refactor update-ips script

The update-ips script have quickly grown quite large and complex, which makes it difficult to add new features to it.

We should rework it so we can more easily update it, and so that the swift endpoint lookup via the keystone endpoint works for anything that needs it (like rclone-sync).

Additionally we should rework it so for each check the process will be 1. fetch, 2. filter (integrating the subnet check worked on [here](https://github.com/elastisys/compliantkubernetes-apps/pull/1582)), then either 3a diff (if dry-run), or 3b apply (if update).
Currently the path for most checks means that we have fetch+diff and fetch+apply as separate functions, so the diff might not generate the same results as apply.

**Additional context**

I would also vote to change the `update-ips <target> update` to be `update-ips <target> apply` to make it more inline with other commands.

**Definition of done:**

Script reworked and simplified.

robinelastisys:
https://github.com/elastisys/compliantkubernetes-apps/issues/1453 https://github.com/elastisys/compliantkubernetes-apps/issues/1677

fixes #1453 
fixes #1677

Ready for review.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.

### DIFF
--- a/bin/ck8s
+++ b/bin/ck8s
@@ -36,7 +36,7 @@ usage() {
     echo "  validate <wc|sc>                                  validates config files" 1>&2
     echo "  providers                                         lists supported cloud providers" 1>&2
     echo "  flavors                                           lists supported configuration flavors" 1>&2
-    echo "  update-ips <wc|sc|both> <update|dry-run>          Automatically fetches and updates the IPs for network policies" 1>&2
+    echo "  update-ips <wc|sc|both> <apply|dry-run>           Automatically fetches and applies the IPs for network policies" 1>&2
     exit 1
 }
 
@@ -152,7 +152,7 @@ case "${1}" in
     flavors) echo "${ck8s_flavors[@]}" ;;
     update-ips)
         [[ "${2}" =~ ^(wc|sc|both)$ ]] || usage
-        [[ "${3}" =~ ^(update|dry-run)$ ]] || usage
+        [[ "${3}" =~ ^(apply|dry-run)$ ]] || usage
         "${here}/update-ips.bash" "${2}" "${3}"
         ;;
     *) usage ;;
--- a/bin/update-ips.bash
+++ b/bin/update-ips.bash
@@ -33,7 +33,7 @@ diffIPs() {
     IPS=("$@")
     tmp_file=$(mktemp --suffix=.yaml)
 
-    yq4 -n '. = []' > "${tmp_file}"
+    yq4 -n '. = []' >"${tmp_file}"
     for ip in "${IPS[@]}"; do
         yq4 -i '. |= . + ["'"${ip}"'/32"]' "${tmp_file}"
     done
@@ -45,63 +45,11 @@ diffIPs() {
     fi
     diff -U3 --color=always \
         --label "${file//${CK8S_CONFIG_PATH}\//}" <(yq4 -P "${yaml_path}"' // [] | sort_by(.)' "${file}") \
-        --label expected <(yq4 -P '. | sort_by(.)' "${tmp_file}") > "${out_file}"
+        --label expected <(yq4 -P '. | sort_by(.)' "${tmp_file}") >"${out_file}"
     DIFF_RETURN=$?
     rm "${tmp_file}"
+
     return ${DIFF_RETURN}
-}
-
-# Get all non /32 IPs in given field and config file
-getConfigIPS() {
-    local source_field="$1"
-    local config_file="$2"
-
-    local configIPS=()
-
-    for ip in $(yq4 e "$source_field | .[]" "$config_file"); do
-        if [[ $ip != */32 ]]; then
-            configIPS+=("${ip}")
-        fi
-    done
-
-    echo "${configIPS[@]}"
-}
-
-# Insert processes IPs
-updateConfigFile() {
-    local source_field="$1"
-    local config_file="$2"
-    local ip="$3"
-
-    if [[ $ip =~ "/" ]]; then
-        yq4 -i "${source_field}"' |= . + ["'"${ip}"'"]' "${config_file}"
-    else
-        yq4 -i "${source_field}"' |= . + ["'"${ip}"'/32"]' "${config_file}"
-    fi
-}
-
-# Check if IPs fits into subnets
-performIPCheckAndRemoveFromIPSList() {
-    local IPToCheck="$1"
-    local networkToCompareTo="$2"
-
-    # Try to see if ip belongs to subnet - if not, exit(1)
-    if python3 -c "import ipaddress; exit(0) if ipaddress.ip_address('${IPToCheck}') in ipaddress.ip_network('${networkToCompareTo}') else exit(1)"; then
-            # Filter out the matching string for IPS.
-            filtered_array=()
-            for ip in "${IPS[@]}"; do
-                if [[ "$ip" != "$IPToCheck" ]]; then
-                    filtered_array+=("$ip")
-                fi
-            done
-
-            IPS=("${filtered_array[@]}")
-
-        # Add working networkToCompareTo to the list of working subnets
-        if [[ " ${working_subnet[*]} " != *" $networkToCompareTo "* ]]; then
-            working_subnet+=("$networkToCompareTo")
-        fi
-    fi
 }
 
 # Fetches the IPs from a specified address
@@ -118,22 +66,41 @@ getDNSIPs() {
 
 # Usage: diffDNSIPs <dns_record> <yaml_path> <file>
 diffDNSIPs() {
-    local IPS
-    read -r -a IPS <<< "$(getDNSIPs "${1}")"
-    diffIPs "${2}" "${3}" "${IPS[@]}"
+    diffIPs "${2}" "${3}" "${@:4}"
     return $?
 }
 
 # Updates the list from the file and yaml path specified with IPs fetched from the domain
 # Usage: updateDNSIPs <dns_record> <yaml_path> <file>
 updateDNSIPs() {
-    endpoint="${1}"
-    configKey="${2}"
-    configFile="$3"
+    local yaml_path="${2}"
+    local file="${3}"
 
-    read -r -a IPS <<<"$(getDNSIPs "${endpoint}")"
+    yq4 -i "${yaml_path}"' = []' "${file}"
 
-    processIPRanges "$configKey" "$configFile"
+    IFS=' ' read -ra ip_array <<<"${@:4}"
+
+    for ip in "${ip_array[@]}"; do
+        yq4 -i "${yaml_path}"' |= . + ["'"${ip}"'"]' "${file}"
+    done
+}
+
+# Usage: updateIPs <yaml_path> <file> <IP 1> <IP ..>
+updateIPs() {
+    local yaml_path="${1}"
+    local file="${2}"
+    shift 2
+    local IPS=("$@")
+
+    yq4 -i "${yaml_path}"' = []' "${file}"
+
+
+
+    IFS=' ' read -ra ip_array <<<"${@:2}"
+
+    for ip in "${ip_array[@]}"; do
+        yq4 -i "${yaml_path}"' |= . + ["'"${ip}"'"]' "${file}"
+    done
 }
 
 # Usage: updateIPs <yaml_path> <file> <IP 1> <IP ..>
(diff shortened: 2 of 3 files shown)


bin/update-ips.bash
```diff
@@ -235,12 +235,30 @@ fi
 
 ## Add sc nodes ips to sc config
 if [[ "${CHECK_CLUSTER}" =~ ^(sc|both)$ ]]; then
-    checkIfDiffAndUpdateKubectlIPs "sc" "" ".networkPolicies.global.scNodes.ips" "${config["override_sc"]}"
+    if [[ "${CK8S_CLOUD_PROVIDER}" =~ ^(citycloud|elastx)$ ]]; then
+        echo "Since your provider is either citycloud or elastx, you can choose to add the internal network(172.16.0.0/32) instead of specific IPs."
+        echo -n "- run? [y/N]: "
+        read -r reply
+        if [[ "${reply}" != "N" ]]; then
+            checkIfDiffAndUpdateKubectlIPs "sc" "" "172.16.0.0/32" "${config["override_sc"]}"
```
aarnq, line 243: You need to check which IPs the nodes have and infer the subnet, since it is not a given that they all use `172.16.0.0`. Also to get the entire subnet it should be `/24` for `172.16.0.x` else it will only be the specific IP ending in 0.

Additionally, it would be nice if it checked that node IPs fall into the set subnet CIDR, and won't prompt if that is the case.

robinelastisys: https://www.arin.net/reference/research/statistics/address_filters/ .. Apparently 172.16.0.0/12 could do the trick as far as coverage is concerned.

viktor-f:
I think that this issue called for a more generic solution that allows us to preserve any given ip somehow. Something like this suggestion from the issue:

> Alternatives would be to perhaps preserve comments to show why certain IPs differ or to ignore values with comments when considering the diff, perhaps both for added visibility.

I especially like the idea that it will not remove anything with a comment.

In addition to this I still want to keep your idea of always adding the private subnet (although as André said I'm not sure we should hard code it to 172.16.0.0/12).

aarnq:
> I think that this issue called for a more generic solution that allows us to preserve any given ip somehow. ...

Yeah we had a discussion on Slack regarding it, and the new idea is that it will treat subnet CIDRs differently compared to individual IPs. And if it finds a CIDR it should keep it and ignore any IPs it finds that matches those CIDRs.

bin/update-ips.bash
```diff
@@ -71,14 +71,102 @@ diffDNSIPs() {
...
+    working_subnet=()
+
+    # Clear config-values
+    yq4 -i "$inputSource"' = []' "$inputConfig"
+
+    # IF config-value is not empty, go ahread with the following:
+    # 1. Check if any got kubectlIPs fits inside of the IPs existing in the config-files.
+    # 2. IF True = Put the working subnet-address and kube-address in assigned arrays for comparison and remove copies.
+    # 3. Merge the working subnet-adresses and kube-addresses into a finalized list.
+    if [ "${configIPS[*]}" != '[]' ]; then
```
Xartos, line 99: Will this ever be `[]`? I think that it will just be `""` if the list is empty. And if you want to check if it's empty, you can do 

```suggestion
    if [ "${#configIPS[@]}" -gt 0 ]; then
```

bin/update-ips.bash
```diff
@@ -51,6 +51,61 @@ diffIPs() {
...
+        yq4 -i "${source_field}"' |= . + ["'"${ip}"'"]' "${config_file}"
+    else
+        yq4 -i "${source_field}"' |= . + ["'"${ip}"'/32"]' "${config_file}"
+    fi
+}
+
+# Check if IPs fits into subnets
+performIPCheck() {
+    local kubectlIP="$1"
+    local configIP="$2"
```
Xartos, line 86: Feels like these could be a little more descriptive as well

```suggestion
    local IPToCheck="$1"
    local networkToCompareTo="$2"
```

bin/update-ips.bash
```diff
@@ -51,6 +51,61 @@ diffIPs() {
...
+
+    # Try to see if ip belongs to subnet - if not, exit(1)
+    if python3 -c "import ipaddress; exit(0) if ipaddress.ip_address('${kubectlIP}') in ipaddress.ip_network('${configIP}') else exit(1)"
+    then
+            # Filter out the matching string for IPS.
+            filtered_array=()
+            for kubectl_ip in "${IPS[@]}"; do
```
Xartos, line 93:
```suggestion
            for ip in "${IPS[@]}"; do
```

bin/update-ips.bash
```diff
@@ -51,6 +51,61 @@ diffIPs() {
     return ${DIFF_RETURN}
 }
 
+# Get IPs in given field and config file
+getConfigIPS() {
```
Xartos, line 55: Isn't this only returning no /32 ranges?

```suggestion
# Get all non /32 IPs in given field and config file
getConfigIPRanges() {
```

bin/update-ips.bash
```diff
@@ -71,12 +126,13 @@ diffDNSIPs() {
 
 # Updates the list from the file and yaml path specified with IPs fetched from the domain
 updateDNSIPs() {
-    read -r -a IPS <<< "$(getDNSIPs "${1}")"
+    endpoint="${1}"
+    inputSource="${2}"
+    inputConfig="$3"
```
Xartos, line 131: Maybe it would make more sense with

```suggestion
    configKey="${2}"
    configFile="$3"
```

bin/update-ips.bash
```diff
@@ -51,6 +51,61 @@ diffIPs() {
...
+    fi
+}
+
+# Check if IPs fits into subnets
+performIPCheck() {
+    local kubectlIP="$1"
+    local configIP="$2"
+
+    # Try to see if ip belongs to subnet - if not, exit(1)
+    if python3 -c "import ipaddress; exit(0) if ipaddress.ip_address('${kubectlIP}') in ipaddress.ip_network('${configIP}') else exit(1)"
```
Xartos, line 89: What happens if the `ipaddress.ip_address('${kubectlIP}')` part gets a range bigger than /32? Will it return only if it's a strict subset or if there's any match?

bin/update-ips.bash
```diff
@@ -109,13 +165,53 @@ diffKubectlIPs() {
...
+
+# Process ip ranges with subnet masks in consideration.
+processIPRanges(){
+    local inputSource="$1"
+    local inputConfig="$2"
+
+    local multiIPRanges=()
+    read -r -a multiIPRanges <<<"$(getConfigIPS "$inputSource" "$inputConfig")"
+
+    local working_subnet=()
```
Xartos, line 187: Is this actually modified in the other function even if this is a `local` variable?

bin/update-ips.bash
robinelastisys: ${kubectlIP} needs to be a valid IP(as in no extras). If not:

"ValueError: '172.16.0.43/32' does not appear to be an IPv4 or IPv6 address" for example.

bin/update-ips.bash
robinelastisys: You could say, yes!

aarnq:
The `update` command seems to well now, though I think we need to work on the dry-run part as well. :sweat_smile:

aarnq:
I propose that we try to hook into the process earlier by running this as a filter on the list of ips (rather than modify how the ips are written into the file, as then we don't need to do a similar thing for how we diff as well). And with that we should rework the update-ips script in general so we can have some common path for how we update and diff ips.

bin/update-ips.bash
```diff
@@ -15,231 +15,351 @@ source "${here}/common.bash"
...
-        yq4 -i "${2}"' |= . + ["'"${ip}"'/32"]' "${3}"
-    done
+  yq4 -i "${yaml_path}"' = []' "${file}"
+
+  IFS=' ' read -ra ip_array <<<"${@:4}"
+
+  for ip in "${ip_array[@]}"; do
+    yq4 -i "${yaml_path}"' |= . + ["'"${ip}"'"]' "${file}"
+  done
 }
```
aarnq, line 87: Redundant.

bin/update-ips.bash
```diff
@@ -15,231 +15,351 @@ source "${here}/common.bash"
...
-    for ip in "${IPS[@]}"; do
-        yq4 -i "${3}"' |= . + ["'"${ip}"'/32"]' "${4}"
-    done
+  yq4 -i "${yaml_path}"' = []' "${file}"
+  IFS=' ' read -ra ip_array <<<"${@:5}"
+
+  for ip in "${ip_array[@]}"; do
+    yq4 -i "${yaml_path}"' |= . + ["'"${ip}"'"]' "${file}"
+  done
 }
```
aarnq, line 144: Redundant.

bin/update-ips.bash
```diff
@@ -15,231 +15,351 @@ source "${here}/common.bash"
...
-    yq4 -i "${yaml_path}"' = []' "${file}"
-    for ip in "${IPS[@]}"; do
-        yq4 -i "${yaml_path}"' |= . + ["'"${ip}"'/32"]' "${file}"
-    done
+  IFS=' ' read -ra ip_array <<<"${@:2}"
+
+  for ip in "${ip_array[@]}"; do
+    yq4 -i "${yaml_path}"' |= . + ["'"${ip}"'"]' "${file}"
+  done
 }
```
aarnq, line 102: We can change this to just write the whole array at once without clearing and adding each entry individually.

bin/update-ips.bash
```diff
@@ -15,231 +15,351 @@ source "${here}/common.bash"
...
+    out_file=/dev/null
+  fi
+  diff -U3 --color=always \
+    --label "${file//${CK8S_CONFIG_PATH}\//}" <(yq4 -P "${yaml_path}"' // [] | sort_by(.)' "${file}") \
+    --label expected <(yq4 -P '. | sort_by(.)' "${tmp_file}") >"${out_file}"
+  DIFF_RETURN=$?
+  rm "${tmp_file}"
+
+  return ${DIFF_RETURN}
 }
```
aarnq, line 51: Simplification

```bash
diff_cidrs() {
  local config_key="${1}"
  local config_file="${2}"
  local cidrs=("${@:3}")

  if $DRY_RUN; then
    out_file=/dev/stdout
  else
    out_file=/dev/null
  fi

  # use diff return implicitly
  diff -U3 --color=always \
    --label "${config_file//${CK8S_CONFIG_PATH}\//}" <(yq4 -P "${config_key}"' // [] | sort' "${config_file}") \
    --label expected <(yq4 -P 'split(" ") | sort' <<< "${cidrs[@]}") > "${out_file}"
}
```

We should not modify anything in the input at this stage, since the input to the diff should reflect the input into update else the diff will be misleading.

bin/ck8s
```diff
@@ -36,7 +36,7 @@ usage() {
     echo "  validate <wc|sc>                                  validates config files" 1>&2
     echo "  providers                                         lists supported cloud providers" 1>&2
     echo "  flavors                                           lists supported configuration flavors" 1>&2
-    echo "  update-ips <wc|sc|both> <update|dry-run>          Automatically fetches and updates the IPs for network policies" 1>&2
+    echo "  update-ips <wc|sc|both> <apply|dry-run>           Automatically fetches and applies the IPs for network policies" 1>&2
```
aarnq, line 39:
```suggestion
    echo "  update-ips <wc|sc|both> <apply|dry-run>           Automatically fetches and updates the IPs for network policies" 1>&2
```

aarnq (commented):
I haven't looked through the main flow but I will once these fundamentals are done.

bin/ck8s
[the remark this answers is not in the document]
robinelastisys: Yes it was! - https://github.com/elastisys/ck8s-ops/pull/4701

Xartos (approved):

PR #1610, opened 2023-05-30
pre-commit: Fix markdown lint

aarnq:

Noticed it didn't work as intended.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.

### DIFF
--- a/.github/ISSUE_TEMPLATE/chart_upgrade.md
+++ b/.github/ISSUE_TEMPLATE/chart_upgrade.md
@@ -7,20 +7,20 @@ assignees: ''
 
 ---
 
-**What Chart should be upgraded?**
+**What Chart should be upgraded?**:
 
 /e.g., dexidp/dex 2.15/
 
-**Why should this Chart be upgraded?**
+**Why should this Chart be upgraded?**:
 
 - [ ] We need to perform a security upgrade.
 - [ ] We need it for a new feature: *which feature (link to blocked issue)*.
 - [ ] We need to keep up to prevent future issues.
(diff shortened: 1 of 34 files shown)


OlleLarsson (approved):

.github/ISSUE_TEMPLATE/chart_upgrade.md
```diff
@@ -7,20 +7,20 @@ assignees: ''
 
 ---
 
-**What Chart should be upgraded?**
+**What Chart should be upgraded?**:
```
crssnd, line 10:
```suggestion
**What Chart should be upgraded?**
```

.github/ISSUE_TEMPLATE/chart_upgrade.md
```diff
@@ -7,20 +7,20 @@ assignees: ''
...
 
 /e.g., dexidp/dex 2.15/
 
-**Why should this Chart be upgraded?**
+**Why should this Chart be upgraded?**:
```
crssnd, line 14:
```suggestion
**Why should this Chart be upgraded?**
```

crssnd (approved):

.github/ISSUE_TEMPLATE/chart_upgrade.md
aarnq: I can't remove it else it doesn't comply with MD036: Emphasis used instead of a header.
So I think this is a good enough circumvention here.

I'll see if I can do some other type of rework when I get around to update the templates.

PR #1612, opened 2023-05-31
apps sc: Harbor GC

### LINKED ISSUE #1602: Enable Harbor GC schedule

To ensure that harbor does not get overfilled with old images, layers, etc. We should enable harbors GC schedule. 
However it is not possible to enable it via the [helm-chart](https://github.com/goharbor/harbor-helm/issues/694) :cry: 
So we need to enable it via the API, as we do when adding [OIDC](https://github.com/elastisys/compliantkubernetes-apps/tree/main/helmfile/charts/harbor/init-harbor). 

https://goharbor.io/docs/edge/administration/garbage-collection/

Good to know :+1: 

```
To avoid damaging the uploading artifact, the garbage collection introduces a time windows(2 hours) to reserve 
the recent uploaded layers. Garbage collection does not sweep the manifest & blob files that have a timestamp 
in the time window. Harbor runs garbage collection without interrupting your ability to continue use Harbor, for 
example you are able to push, pull, or delete artifacts while garbage collection is running.
```

I want to be able to enable harbors GC from config and setting a schedule.

Not implementing this. 

**Additional context**

No

**Definition of done:**

When it is possible to enable harbors GC from config and setting a schedule.

robinAwallace:

Enable harbor GC
Did some other refactoring aswell.

**Which issue this PR fixes**: fixes #1602 

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [x] I upgraded no Chart.


helmfile/charts/harbor/init-harbor/files/init.sh
```diff
@@ -0,0 +1,101 @@
...
+              \"schedule\": {
+                \"cron\": \"${GC_SCHEDULE}\",
+                \"type\": \"Custom\"
+              }
+            }")
+    if [ -n "$err" ]; then
+        echo "ERROR when configuring GC: $err"
+        exit 1
+    fi
+}
```
viktor-f, line 93: Should we have some check to see if it already enabled? I'm thinking that we might otherwise override any schedule that the users might have set if this init-script re-runs later.

config/config/sc-config.yaml
```diff
@@ -397,6 +397,9 @@ harbor:
     retentionDays: 7
   s3:
     multipartcopythresholdsize: "536870912"
+  gc:
+    enabled: false
```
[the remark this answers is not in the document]
robinAwallace: Yeah it should be enabled by default. Just missed it.

helmfile/charts/harbor/init-harbor/files/init.sh
robinAwallace: First thought, yes that makes sense. On seconded thought, then when we want to change it via the init-script it becomes a bit harder to check. 

I do not think any one will change it via the GUI once it is already been set. But i can add a check and a force update variable if we want to enforce updating it from the init-script. What do you think, should i add a check?

robinAwallace: Added the check

helmfile/charts/harbor/init-harbor/files/init.sh
```diff
@@ -76,6 +76,32 @@ configure_OIDC() {
...
+            body=""
+        else
+            # shellcheck disable=SC3057
+            body="${res:0:${#res}-3}"
+        fi
+
+        if [ -n "$body" ]; then
+            echo "GC already configured"
+            return
+        fi
```
viktor-f, line 102: Nit: Couldn't you compact it to something like this?
```suggestion
        if [ ${#res} -ne 3 ]; then
            echo "GC already configured"
            return
        fi
```

davidumea: @robinAwallace I think you forgot to change the `-eq` operator to `-ne`

viktor-f (approved):
LGTM

davidumea (approved):

PR #1615, opened 2023-06-02
apps wc: Fix HNC migration and race condition

aarnq:

Since it was causing failures when upgrading.
I haven't tested to install with this change but I don't see why there would be an issue with it.
However this requires that whenever we update managed labels we must update HNC first then user-rbac. And we need to change that dependency in helmfile though I need to better test that first before I'm willing to make that change.

**Checklist:**

    - [ ] will change any APIs.
    - [ ] will break the cluster.
    - [ ] I upgraded no Chart.


robinAwallace (approved):
:+1:

OlleLarsson (approved):

crssnd (approved):

PR #1640, opened 2023-06-19
Anders elastisys/add multiple kubeapiservers support

anders-elastisys:
Currently, if we configure different global `baseDomain` for `wc` and `sc` we get `EndpointDown` alerts since `sc` will probe the wrong domain. This PR allows us to override the default configuration for kube-apiserver blackbox monitoring, which is useful if we want to be able to configure multiple kube-apiserver domains (e.g. in the case of having multiple workload clusters with only one service cluster).

One issue with this templating is that it is possible to override other existing blackbox endpoints by adding them to the [`kubeApiservers`](https://github.com/elastisys/compliantkubernetes-apps/compare/anders-elastisys/add-multiple-kubeapiservers-support?expand=1#diff-32ea2de1ac9797b598bd4286c8dde58bfe56155405e078907360a224a977085aR1084) list (e.g. `harbor` or `nginx`), so one has to be careful when configuring it.

I am open for suggestions on possible improvements or other ways to solve this problem.

The uptime dashboard could be left as it is and it would not be a problem for normal CK8s setups, but in the case of having multiple workload clusters it would be nice to view them in the uptime dashboard somehow. In the screenshot below it can be seen that the kube-apiserver panels fills the screen due to them being configured with the "Repeat by variable" option, [which does not seem to support having a set width?](https://github.com/grafana/grafana/pull/8199)

The following screenshot shows how the Uptime dashboard looks like with the new changes.
uptime-dashboard

**Checklist:**

  - [ ] will change any APIs.
  - [ ] will break the cluster.
  - [ ] I upgraded no Chart.


config/config/sc-config.yaml
```diff
@@ -1081,6 +1081,8 @@ prometheusBlackboxExporter:
     thanosQuery: true
     thanosReceiver: true
 
+kubeapiServers: []
```
OlleLarsson, line 1084: Could we instead have something like
```yaml
prometheusBlackboxExporter:
  # Informative comment..
  customKubeapiTargets:
    - name: ..
      domain: ..
```
?

anders-elastisys: I like it. [Changed it](https://github.com/elastisys/compliantkubernetes-apps/blob/anders-elastisys/add-multiple-kubeapiservers-support/config/config/sc-config.yaml#L1081-L1086), better?

crssnd (approved):

OlleLarsson (approved):
Any comments/thoughts on this @aarnq / @davidumea ?

davidumea (approved):

helmfile/charts/grafana-dashboards/dashboards/uptime-dashboard.json
```diff
@@ -901,7 +906,37 @@
...
+          ],
+          "value": [
+            "user-api-server"
+          ]
+        },
+        "cacheTimeout": null,
+        "datasource": "prometheus-sc",
+        "definition": "label_values(probe_success{instance=~\".*kube-apiserver.*\"}, target)",
+        "hide": 0,
+        "includeAll": false,
```
aarnq, line 924:
```suggestion
        "includeAll": true,
```
And possibly have it default to all.

anders-elastisys: Changed it so default is now "all"

aarnq (approved):

Issue #1405, opened 2023-02-09
Upgrade falco to version v0.34.1


raviranjanelastisys:
**What Chart should be upgraded?**
upstream/falco

Make sure to also update the [falco-sidekick](https://github.com/falcosecurity/charts/blob/master/falcosidekick/Chart.yaml)

**Why should this Chart be upgraded?**
- [ ] We need to perform a security upgrade.
- [ ] We need it for a new feature: *which feature (link to blocked issue)*.
- [x] We need to keep up to prevent future issues.
- [ ] Other: *Write the reason here*
- I checked the migration of the new Chart:
    - [ ] I upgraded a Chart and determined that no migration steps are needed.
    - [ ] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).
- [ ] I tested the functionality provided by the new Chart (e.g., Auth flow, Grafana dashboards, etc.)

PR #1662, opened 2023-07-04
Anders elastisys/upgrade falco chart

anders-elastisys:
[Falco supports three different types of syscall sources](https://falco.org/docs/event-sources/kernel/), _kernel module_ (default), _classic eBPF probe_ and the more recent _modern eBPF probe_. The issue with the default type is that it is tightly coupled with the host kernel, and is reliant on loading kernel modules that sometimes may not exists (and if they don't, Falco pods will end up in a CrashLoopBackOff state until a module that supports the current host kernel version is released). 

The other two options are [supported by most of the recent Linux kernels](https://falco.org/docs/event-sources/kernel/#requirements) and don't need to download any external modules. These options can also be [run with less privileges](https://falco.org/docs/event-sources/kernel/#least-privileged-mode-2) than with the default _kernel module_ configuration.

Also, the Falco Helm chart used in CK8s was behind latest release with one major version, and the Falco apps version was behind with 2 minor versions.

This PR upgrades the Falco Helm chart and apps version, as well as adding templating to be able to configure a different syscall source for Falco, as currently there is no good way to change the syscall source for Falco in CK8s.

**Which issue this PR fixes**: fixes https://github.com/elastisys/compliantkubernetes-apps/issues/1405

Worth noting that from version 3.0.0 and onwards of the Falco Helm chart, the chart deploys [falcoctl](https://github.com/falcosecurity/charts/blob/master/falco/BREAKING-CHANGES.md#falcoctl-support) by default. `falcoctl` is useful to manage plugins and updating them continuously (e.g. _rulesfiles_) without having to wait on new releases of Falco. However, `falcoctl` is not required and is **disabled**  with this PR as the goal of this PR was to upgrade the Falco version and make it possible to use the `modern-bpf` syscall source.

**Checklist:**

  - [ ] will change any APIs.
  - [ ] will break the cluster.
  - [ ] I upgraded no Chart.


helmfile/values/falco.yaml.gotmpl
```diff
@@ -12,12 +12,26 @@ falco:
...
+    path: {{ .Values.falco.driver.ebpf.path }}
+    hostNetwork: {{ .Values.falco.driver.ebpf.hostNetwork }}
+    leastPrivileged: true
+  {{- else if eq .Values.falco.driver.kind "modern-bpf" }}
+  modern_bpf:
+    leastPrivileged: true
+  {{- end }}
   loader:
     initContainer:
       securityContext:
```
viktor-f, line 37: Is this loader used for all three kinds of drivers? Looks like it is at least used for "module" and "ebpf".

helmfile/values/falco.yaml.gotmpl
```diff
@@ -30,6 +44,13 @@ nodeSelector: {{- toYaml .Values.falco.nodeSelector | nindent 2  }}
 affinity: {{- toYaml .Values.falco.affinity | nindent 2  }}
 tolerations: {{- toYaml .Values.falco.tolerations | nindent 2  }}
 
+falcoctl:
+  artifact:
+    install:
+      enabled: false
+    follow:
+      enabled: false
+
```
viktor-f, line 53: Would you like to write an issue about enabling this? Otherwise I'll do it.

helmfile/values/podsecuritypolicies/common/falco.yaml.gotmpl
```diff
@@ -12,7 +12,11 @@ constraints:
         - pathPrefix: /usr
         - pathPrefix: /etc
         - pathPrefix: /dev
+        {{- if eq .Values.falco.driver.kind "ebpf" }}
+        - pathPrefix: /sys/kernel/debug
+        {{- else if eq .Values.falco.driver.kind "module" }}
         - pathPrefix: /sys/module/falco
+        {{- end }}
```
viktor-f, line 19: Nothing needed here for the "modern-bpf" driver?

anders-elastisys: No, I have tested `modern-bpf` in a cluster that supported it and using that templating seems to have worked fine. Looking in the [Falco Helm chart,](https://github.com/falcosecurity/charts/blob/master/falco/templates/pod-template.tpl#L165-L175) there doesn't seem to be any extra mounts for `modern-bpf`.

helmfile/values/falco.yaml.gotmpl
anders-elastisys: The loader init container is disabled when running running with `modern-bpf`, you can see the [templating in the Helm chart here](https://github.com/falcosecurity/charts/blob/falco-3.3.0/falco/templates/pod-template.tpl#L229-L233). It is also mentioned [here](https://github.com/falcosecurity/charts/tree/falco-3.3.0/falco#about-drivers) that `modern-bpf` does not require the loader. So the Helmfile templating here should just be ignored when running with `kind: modern-bpf`

helmfile/values/falco.yaml.gotmpl
anders-elastisys: I can do that!

anders-elastisys: There is now an issue on enabling `falcoctl` here: https://github.com/elastisys/compliantkubernetes-apps/issues/1669

viktor-f (approved):
LGTM, you can squash the commits as well.

aarnq (approved):
Seems to work well, unfortunately I'm not able to test the eBPF backend.

davidumea (approved):

PR #1697, opened 2023-07-31
apps: configure trailing dots(grafana) and ndots for pods

### LINKED ISSUE #1671: [2] Configure DNS resolve to prevent resolve issues

Some containers in some environments may experience dns resolve issues, in particular when resolving external names.

This in particularly affect Grafana a lot in some cases so we've added trailing dots to make the domain names absolute.
However this has turned out to cause compatibility issues with certain browser preventing users from login in.

I would like to see us properly configuring the dnsConfig in some way so that this can be corrected without the trailing dot.

Alternative would be to change the cluster wide resolve config to fix this, however that can potentially lead to other issues with how DNS is expected to work in Kubernetes.

**Definition of done:**

Resolve issues dealt with without trailing dots.

linus-astrom:
We can now choose if grafana needs to have trailing dots and a gatekeeper mutation that makes pods have the specified amount of ndots if enabled.

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1671

**Checklist:**

  - [ ] will change any APIs.
  - [ ] will break the cluster.
  - [x] I upgraded no Chart.

### DIFF
--- a/helmfile/charts/gatekeeper/mutations/templates/ndot-amount-dns-policy.yaml
+++ b/helmfile/charts/gatekeeper/mutations/templates/ndot-amount-dns-policy.yaml
@@ -1,0 +1,42 @@
+{{- define "ndots-mutation" }}
+---
+apiVersion: mutations.gatekeeper.sh/v1
+kind: Assign
+metadata:
+  {{- if .mutationLabelValue }}
+  name: ndot-amount-dns-policy-{{ .mutationLabelValue }}
+  {{- else }}
+  name: ndot-amount-dns-policy
+  {{- end }}
+spec:
+  applyTo:
+    - groups: [""]
+      kinds: ["Pod"]
+      versions: ["v1"]
+  match:
+    scope: Namespaced
+    kinds:
+      - apiGroups: ["*"]
+        kinds: ["Pod"]
+    {{- if .mutationLabelValue }}
+    labelSelector:
+      matchLabels:
+        {{ .mutationLabelKey }}: {{ .mutationLabelValue }}
+    {{- end }}
+  location: "spec.dnsConfig"
+  parameters:
+    assign:
+      value:
+        options:
+        - name: ndots
+          value: "{{ .Values.ndots.ndotAmount }}"
+{{- end }}
+{{- if $.Values.ndots.enabled }}
+{{- if gt (len ($.Values.ndots | dig "labelSelector" "matchLabels" dict)) 0 }}
+{{- range $key, $val := .Values.ndots.labelSelector.matchLabels }}
+{{- include "ndots-mutation" (mustMergeOverwrite (dict "mutationLabelKey" $key "mutationLabelValue" $val) $) }}
+{{- end }}
+{{- else }}
+{{- include "ndots-mutation" $ }}
+{{- end }}
+{{- end }}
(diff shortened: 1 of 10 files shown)


helmfile/charts/gatekeeper/mutations/templates/ndot-amount-dns-policy.yaml
```diff
@@ -0,0 +1,35 @@
+{{- if $.Values.ndots.enabled }}
+{{- $path:= "" }}
+{{- if $.Values.ndots.labelSelector.enabled }}
+  {{- $path = .Values.ndots.labelSelector.matchLabels }}
+{{- end }}
+{{- range $key, $val := $path | default .Values.ndots.clusterWide }}
```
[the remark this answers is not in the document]
linus-astrom: yes its a string but its used for the name and making the range only cycle once. Because its only active when labelselector is disabled.

helmfile/charts/gatekeeper/mutations/templates/ndot-amount-dns-policy.yaml
```diff
@@ -0,0 +1,35 @@
...
+    {{- end }}
+  location: "spec.dnsConfig"
+  parameters:
+    assign:
+      value:
+        options:
+        - name: ndots
+          value: "{{ $.Values.ndots.ndotAmount }}"
+{{- end }}
+{{- end }}
```
Xartos, line 35: What you could do instead is 

```suggestion
{{- if $.Values.ndots.enabled }}
{{ define "ndots-mutation" }}
---
apiVersion: mutations.gatekeeper.sh/v1
kind: Assign
metadata:
  {{- if .mutationLabelValue }}
  name: ndot-amount-dns-policy-{{ .mutationLabelValue }}
  {{- else }}
  name: ndot-amount-dns-policy
  {{- end }}
spec:
  applyTo:
    - groups: [""]
      kinds: ["Pod"]
      versions: ["v1"]
  match:
    scope: Namespaced
    kinds:
      - apiGroups: ["*"]
        kinds: ["Pod"]
    {{- if .mutationLabelValue }}
    labelSelector:
      matchLabels:
        {{ .mutationLabelKey }}: {{ .mutationLabelValue }}
    {{- end }}
  location: "spec.dnsConfig"
  parameters:
    assign:
      value:
        options:
        - name: ndots
          value: "{{ .Values.ndots.ndotAmount }}"
{{- end }}

{{- if $.Values.ndots.labelSelector.enabled }}
{{- range $key, $val := .Values.ndots.labelSelector.matchLabels }}
{{- include "ndots-mutation" (mustMergeOverwrite (dict "mutationLabelKey" $key "mutationLabelValue" $val) $) }}
{{- end }}
{{- else }}
{{- include "ndots-mutation" $ }}
{{- end }}
{{- end }}
```

Then you don't have to have this magic variable that has to be a particular length for it to work. And we still only get one thing to manage. I might've missed something but I think this should work.

linus-astrom: Fixed thou i had to change where the define and if statement in the beginning was because define does not like being under a if statement.

Xartos (commented):
Other than the last thing I have now I think it looks good

lunkan93 (approved):

PR #1700, opened 2023-08-01
Probe public SC endpoints from WC

### LINKED ISSUE #1437: Add blackbox exporter dashboard and add probes from WC to SC
It would be of great interest to add a dashboard for blackbox exporter to easily see the performance of probes within an environment, both in terms of DNS resolve and TCP connect.

Furthermore it would be nice with probes going from the WC to public SC endpoints to monitor how well it can communicate with it.

- Added dashboard for blackbox exporter
- Added probes in WC against public target in SC

**Definition of done:**
Both points added above.

Zash:
**What this PR does / why we need it**: to monitor how well WC can communicate with SC

**Which issue this PR fixes**: fixes #1437

**Special notes for reviewer**: Depends on Blackbox Exporter dashboard added in https://github.com/elastisys/compliantkubernetes-apps/pull/1542

image

**Checklist:**

  - [ ] will change any APIs.
  - [ ] will break the cluster.
  - [ ] I upgraded no Chart.
  - [x] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).

### DIFF
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -87,7 +87,7 @@ falco:
 
 grafana:
   ops:
-    subdomain: grafana
+    # subdomain moved to common-config
     resources:
       requests:
         cpu: 50m
@@ -119,7 +119,7 @@ grafana:
           memory: 100Mi
   user:
     enabled: true
-    subdomain: grafana
+    # subdomain moved to common-config
     resources:
       limits:
         cpu: 100m
@@ -151,8 +151,8 @@ grafana:
           memory: 100Mi
 
 harbor:
-  enabled: true
-  subdomain: harbor
+  # the enabled property moved to common-config
+  # subdomain moved to common-config
   # The tolerations, affinity, and nodeSelector are applied to all harbor pods.
   tolerations: []
   nodeSelector: {}
@@ -315,7 +315,7 @@ harbor:
 
   notary:
     replicas: 1
-    subdomain: notary.harbor
+    # subdomain moved to common-config
     resources:
       requests:
         cpu: 20m
@@ -479,7 +479,7 @@ prometheus:
 
 dex:
   replicaCount: 2
-  subdomain: dex
+  # subdomain moved to common-config
   additionalKubeloginRedirects: []
   enableStaticLogin: true
   serviceMonitor:
@@ -708,7 +708,7 @@ opensearch:
   createIndices: true
 
   dashboards:
-    subdomain: opensearch
+    # subdomain moved to common-config
     # Note SSO is enabled via `opensearch.sso.enabled`
     resources:
       requests:
--- a/helmfile/values/prometheus-blackbox-exporter-wc.yaml.gotmpl
+++ b/helmfile/values/prometheus-blackbox-exporter-wc.yaml.gotmpl
@@ -66,3 +66,33 @@ serviceMonitor:
       scrapeTimeout: 30s
       module: http_2xx
     {{- end }}
+
+    - name: sc-dex
+      url: https://{{ .Values.dex.subdomain }}.{{ .Values.global.baseDomain }}/healthz
+      interval: 60s
+      scrapeTimeout: 30s
+      module: http_2xx
+    {{- if .Values.harbor.enabled }}
+    - name: sc-harbor
+      url: https://{{ .Values.harbor.subdomain }}.{{ .Values.global.baseDomain }}/api/v2.0/ping
+      interval: 60s
+      scrapeTimeout: 30s
+      module: http_2xx
+    {{- end }}
+    - name: sc-user-grafana
+      url: https://{{ .Values.grafana.user.subdomain }}.{{ .Values.global.baseDomain }}/api/health
+      interval: 60s
+      scrapeTimeout: 30s
+      module: http_2xx
+    - name: sc-opensearch-ops
+      url: https://{{ .Values.opensearch.subdomain }}.{{ .Values.global.opsDomain }}
+      interval: 60s
+      scrapeTimeout: 30s
+      module: http_401
+    {{- if and .Values.thanos.enabled .Values.thanos.receiver.enabled }}
+    - name: sc-thanos-recv-ops
+      url: https://{{ .Values.thanos.receiver.subdomain }}.{{ .Values.global.opsDomain }}
+      interval: 60s
+      scrapeTimeout: 30s
+      module: http_401
+    {{- end }}
--- a/scripts/migration/yq.sh
+++ b/scripts/migration/yq.sh
@@ -38,6 +38,19 @@ yq_move() {
   fi
 }
 
+yq_move_to_file() {
+  if [[ "${#}" -lt 4 ]] || [[ ! "${1}" =~ ^(common|sc|wc)$ ]] || [[ ! "${3}" =~ ^(common|sc|wc)$ ]]; then
+    log_fatal "usage: yq_move_to_file <common|sc|wc> <source> <common|sc|wc> <destination>"
+  fi
+
+  if ! yq_null "${1}" "${2}"; then
+    log_info "  - move: ${1} ${2} to ${4} ${3}"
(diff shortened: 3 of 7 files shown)


Xartos (commented):
Looks good. Only thing I'm missing is some migration scripts for the config change. Something like this one https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration/v0.31/prepare/02-move-grafana-values.sh

config/config/sc-config.yaml
```diff
@@ -472,7 +468,6 @@ prometheus:
 
 dex:
   replicaCount: 2
-  subdomain: dex
```
crssnd, line 475: can you add some comments saying that the `subdomain` key can be found in the common config?

crssnd:
I am not sure that we need to ping this 3 endpoints, seems redundant and not useful:
```
sc-harbor-notary
sc-grafana-ops
sc-opensearch
```
I would like more opinions about this

crssnd:
did you check that the network policies allow this probing?

Zash:
> did you check that the network policies allow this probing?

Tested with apps v0.30.4, all the added probes succeed in my test cluster.

config/config/sc-config.yaml
Zash: done in https://github.com/elastisys/compliantkubernetes-apps/pull/1700/commits/d0df25d4e6f47a209cae55b835f7f2e4d13eba44

scripts/migration/yq.sh
```diff
@@ -38,6 +38,19 @@ yq_move() {
   fi
 }
 
+yq_xmove() {
```
Xartos, line 41: I think I would rather like
```suggestion
yq_move_file() {
```

To make it a bit more descriptive

crssnd:
> > I am not sure that we need to ping this 3 endpoints, seems redundant and not useful:
> > ```
> > sc-harbor-notary
> > sc-grafana-ops
> > sc-opensearch
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I would like more opinions about this
> 
> I don't see why we wouldn't. It could help in a scenario where the services isn't accessable externally but this probe works. That would tell us that the issue lies between the external network and the internal network. While if these probes would fail it's probably something internally in the service cluster (some firewall rule or whatnot).

Maybe I am wrong, but if we use the FQDN would this also go through the external network and the sc LB IP (also used by user-grafana and ops opensearch, harbor portal etc)?

At leat I feel that the current external and internal monitoring that we have for sc should cover the example you gave and in this issue we only want to see the latency between the clusters (I think mainly useful for thanos and opensearch endpoints)

scripts/migration/yq.sh
Zash: `yq_move_to_file` as to not imply that it moves files themselves?

Xartos: Yea, that sounds good

Xartos:
After a discussion with @crssnd I'm on his side, so I also think we can remove the three he suggested

Zash:
> in /github/workspace/helmfile/50-applications.yaml: failed to render values files "values/prometheus-blackbox-exporter-wc.yaml.gotmpl": failed to render [values/prometheus-blackbox-exporter-wc.yaml.gotmpl], because of template: stringTemplate:80:18: executing "stringTemplate" at <.Values.harbor.enabled>: map has no entry for key "enabled"

Hm, does `.harbor.enabled` need moving from `sc-config.yaml` to `common-config.yaml` as well?

crssnd:
> > I am not sure that we need to ping this 3 endpoints, seems redundant and not useful:
> > ```
> > sc-harbor-notary
> > sc-grafana-ops
> > sc-opensearch
> > ```
> 
> Keep `sc-opensearch-ops` ?

yes

crssnd:
> > in /github/workspace/helmfile/50-applications.yaml: failed to render values files "values/prometheus-blackbox-exporter-wc.yaml.gotmpl": failed to render [values/prometheus-blackbox-exporter-wc.yaml.gotmpl], because of template: stringTemplate:80:18: executing "stringTemplate" at <.Values.harbor.enabled>: map has no entry for key "enabled"
> 
> Hm, does `.harbor.enabled` need moving from `sc-config.yaml` to `common-config.yaml` as well?

probably if it's now use for both cluster

migration/v0.31/prepare/07-move-subdomains.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
crssnd, line 1: you will need to move this scripts to v0.33
maybe also merge them, as I do not see a reason to keep the separately

migration/v0.31/prepare/07-move-subdomains.sh
```diff
@@ -0,0 +1,15 @@
...
+
+HERE="$(dirname "$(readlink -f "${0}")")"
+ROOT="$(readlink -f "${HERE}/../../../")"
+
+# shellcheck source=scripts/migration/lib.sh
+source "${ROOT}/scripts/migration/lib.sh"
+
+log_info "- move subdomain values to common"
```
crssnd, line 9: also add them into the migration README files in the manual method section

helmfile/values/prometheus-blackbox-exporter-wc.yaml.gotmpl
```diff
@@ -66,3 +66,38 @@ serviceMonitor:
       scrapeTimeout: 30s
       module: http_2xx
     {{- end }}
+
+    - name: sc-api-server
+      url: https://kube-apiserver.{{ .Values.global.baseDomain }}:/healthz
```
crssnd, line 71: I do not think we are using this subdomain for sc

helmfile/values/prometheus-blackbox-exporter-wc.yaml.gotmpl
```diff
@@ -66,3 +66,38 @@ serviceMonitor:
...
+      scrapeTimeout: 30s
+      module: http_2xx
+    {{- if .Values.harbor.enabled }}
+    - name: sc-harbor
+      url: https://{{ .Values.harbor.subdomain }}.{{ .Values.global.baseDomain }}/api/v2.0/ping
+      interval: 60s
+      scrapeTimeout: 30s
+      module: http_2xx
+    {{- end }}
+    - name: sc-grafana
```
crssnd, line 87:
```suggestion
    - name: sc-user-grafana
```

migration/v0.31/prepare/07-move-subdomains.sh
Zash: done

helmfile/values/prometheus-blackbox-exporter-wc.yaml.gotmpl
Zash: Hm, I think I just copied that from

https://github.com/elastisys/compliantkubernetes-apps/blob/1651f30df8a1d1eb4a98c576cf91a321c06e5d17/helmfile/values/prometheus-blackbox-exporter-sc.yaml.gotmpl#L106

No idea if this is actually relevant but it looks hard-coded here:

https://github.com/elastisys/compliantkubernetes-apps/blob/1651f30df8a1d1eb4a98c576cf91a321c06e5d17/helmfile/charts/kubeapi-metrics/templates/ingress.yaml#L3

crssnd: I need to check and return with an answer

Xartos (approved):

helmfile/values/prometheus-blackbox-exporter-wc.yaml.gotmpl
crssnd: this one is used for user-apiserver only, we do not create and ingress for apiserver in sc so I think you can remove this check

crssnd (approved):
please do not forget to squash

PR #1755, opened 2023-09-05
New test harness

aarnq:
### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           

> [!important]

- [ ] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

This paves the way for more structured unit, integration, and end-to-end testing by using bats.

#### Additional information to reviewers

This has a long way to go before it can be used in the pipeline, ~~and I want to integrate Cypress first so I know what requirements we have.~~ ~~I also aim to implement helpers for the current main test script and see if there is any way to generate tests from a matrix.~~

I've already did so for testing deployments and statefulsets, though we might want to rework it into a fetch-once-check-multiple solution as the old one did to improve the speed. Though with bats we can run things in parallel.

Cypress support is done! It is integrated with generators into the bats suite, but can be run manually as well. A support file is setup so we can add common functions if we need to, and a few config ones have been implemented.

There is also now a sort of generator for testing matrices, or at least the building blocks for doing so.

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,0 +1,62 @@
+FROM ubuntu:rolling
+LABEL org.opencontainers.image.source https://github.com/elastisys/compliantkubernetes-apps
+
+ARG DEBIAN_FRONTEND=noninteractive
+ARG TZ=Etc/UTC
+
+RUN apt-get update && \
+    apt-get install -y apache2-utils bats curl dnsutils gettext-base git gpg iputils-ping jq make net-tools npm pwgen s3cmd ssh unzip wget \
+        libasound2 libgbm-dev libgconf-2-4 libgtk2.0-0 libgtk-3-0 libnotify-dev libnss3 libxss1 libxtst6 xauth xvfb && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+ENV SOPS_VERSION="3.7.3"
+RUN wget --progress=dot:giga "https://github.com/mozilla/sops/releases/download/v${SOPS_VERSION}/sops-v${SOPS_VERSION}.linux" && \
+    install -Tm 755 "sops-v${SOPS_VERSION}.linux" /usr/local/bin/sops && \
+    rm "sops-v${SOPS_VERSION}.linux"
+
+ENV KUBECTL_VERSION "1.24.4"
+RUN wget --progress=dot:giga "https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl" && \
+    install -Tm 755 kubectl /usr/local/bin/kubectl && \
+    rm kubectl
+
+ENV KUBELOGIN_VERSION="1.24.0"
+RUN wget --progress=dot:giga "https://github.com/int128/kubelogin/releases/download/v${KUBELOGIN_VERSION}/kubelogin_linux_amd64.zip" && \
+    unzip kubelogin_linux_amd64.zip && \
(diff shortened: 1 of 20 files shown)


aarnq:
@robinAwallace that is not in the scope of this PR but in the future for sure.
I am working on methods of generating tests, and with that in place I think we could easily do loads of such tests. Because then you could just configure target cluster, resource, config key and config values and let it do the work.

robinAwallace:
Perfect, understandably that it is not part of this PR :+1:

tests/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+RUN chown -R ubuntu:ubuntu /tests
+
+WORKDIR /tests
+
+USER ubuntu
+
+RUN ls -l
+
+RUN make dep
```
simonklb, line 60: Would it be possible to only copy the necessary files for `make dep` and then the rest to reduce build time when updating tests?

simonklb: I just tried this and it _seems_ to be working. The update test -> build -> run round trip time is reduced a lot by this so it would be nice to see something to this effect.

```suggestion
COPY ./Makefile /tests/Makefile
COPY ./common /tests/common

RUN chown -R ubuntu:ubuntu /tests

WORKDIR /tests

USER ubuntu

RUN ls -l

RUN make dep

COPY ./ /tests
```

aarnq: Yes, I was going to start fixing that. :+1:

aarnq: Also fixed so it generates tests during build.

simonklb: Sweet!

crssnd (approved):

Xartos (commented):
I think I would like to request a "meeting PR review" on this one.

tests/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    install -Tm 755 kubelogin /usr/local/bin/kubectl-oidc_login && \
+    rm kubelogin kubelogin_linux_amd64.zip LICENSE README.md
+
+ENV HELM_VERSION "3.8.0"
+RUN wget --progress=dot:giga "https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz" && \
+    tar -zxvf "helm-v${HELM_VERSION}-linux-amd64.tar.gz" && \
+    install -Tm 755 linux-amd64/helm /usr/local/bin/helm && \
+    rm -rf linux-amd64 "helm-v${HELM_VERSION}-linux-amd64.tar.gz"
+
+ENV XDG_DATA_HOME=/root/.config
```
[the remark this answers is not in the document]
aarnq: Yes, since the env will be saved within the image it will reuse that path when you enter it.

simonklb (approved):
Nice work!

I think we should consider adding the generate tests to the git repository though. I find it nice to track diffs when I update generated code. Interested to hear your take on it. (not blocking the merge btw)

aarnq:
> Nice work!
> 
> I think we should consider adding the generate tests to the git repository though. I find it nice to track diffs when I update generated code. Interested to hear your take on it. (not blocking the merge btw)

Maybe, I don't think I want to add it right now to keep the structure simple.
But if we start to implement more features into the generators (and add the output change/tracking tests) then we will definitely go in that direction.

PR #1771, opened 2023-09-12
apps wc: Skip installing fluentd example resources if disabled

simonklb:
> [!warning]
> - [ ] personal data beyond what is necessary for interacting with this pull request
> - [ ] business confidential information, such as customer names

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

The `fluentd` namespace is not created during bootstrap if `fluentd.enabled = false` but the Fluentd "example resources" are still created which errors. This fixes that.

Alternatively I could remove the "example resources" as they seem pretty out of place. Does anyone know if they are being used somehow?

- Fixes #

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:

### DIFF
--- a/scripts/deploy-wc.sh
+++ b/scripts/deploy-wc.sh
@@ -19,21 +19,25 @@ INTERACTIVE=${1:-""}
 # Add example resources.
 # We use `create` here instead of `apply` to avoid overwriting any changes the
 # user may have done.
-if [ "$(kubectl get configmap fluentd-extra-config -n fluentd)" ] ; then
-  echo "fluentd-extra-config ConfigMap already in place. Ignoring."
+if ! kubectl get ns fluentd > /dev/null; then
+  echo "fluentd namespace missing, skipping installing fluentd example resources."
 else
-  echo "Creating fluentd-extra-config ConfigMap"
-  kubectl create -f "${SCRIPTS_PATH}/../manifests/examples/fluentd/fluentd-extra-config.yaml"
+  if kubectl get configmap fluentd-extra-config -n fluentd > /dev/null; then
+    echo "fluentd-extra-config ConfigMap already in place. Ignoring."
+  else
+    echo "Creating fluentd-extra-config ConfigMap"
+    kubectl create -f "${SCRIPTS_PATH}/../manifests/examples/fluentd/fluentd-extra-config.yaml"
+  fi
+
+  if kubectl get configmap fluentd-extra-plugins -n fluentd > /dev/null ; then
+    echo "fluentd-extra-plugins ConfigMap already in place. Ignoring."
+  else
+    echo "Creating fluentd-extra-plugins ConfigMap"
+    kubectl create -f "${SCRIPTS_PATH}/../manifests/examples/fluentd/fluentd-extra-plugins.yaml"
+  fi
 fi
 
-if [ "$(kubectl get configmap fluentd-extra-plugins -n fluentd)" ] ; then
-  echo "fluentd-extra-plugins ConfigMap already in place. Ignoring."
-else
-  echo "Creating fluentd-extra-plugins ConfigMap"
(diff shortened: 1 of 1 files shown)


scripts/deploy-wc.sh
```diff
@@ -19,18 +19,22 @@ INTERACTIVE=${1:-""}
...
+    echo "Creating fluentd-extra-config ConfigMap"
+    kubectl create -f "${SCRIPTS_PATH}/../manifests/examples/fluentd/fluentd-extra-config.yaml"
+  fi
+
+  if [ "$(kubectl get configmap fluentd-extra-plugins -n fluentd)" ] ; then
+    echo "fluentd-extra-plugins ConfigMap already in place. Ignoring."
+  else
+    echo "Creating fluentd-extra-plugins ConfigMap"
+    kubectl create -f "${SCRIPTS_PATH}/../manifests/examples/fluentd/fluentd-extra-plugins.yaml"
+  fi
```
OlleLarsson, line 37: Maybe we can update this to follow the format of your introduced check?

```suggestion
  if kubectl get configmap fluentd-extra-config -n fluentd > /dev/null 2>&1 ; then
    echo "fluentd-extra-config ConfigMap already in place. Ignoring."
  else
    echo "Creating fluentd-extra-config ConfigMap"
    kubectl create -f "${SCRIPTS_PATH}/../manifests/examples/fluentd/fluentd-extra-config.yaml"
  fi

  if kubectl get configmap fluentd-extra-plugins -n fluentd > /dev/null 2>&1; then
    echo "fluentd-extra-plugins ConfigMap already in place. Ignoring."
  else
    echo "Creating fluentd-extra-plugins ConfigMap"
    kubectl create -f "${SCRIPTS_PATH}/../manifests/examples/fluentd/fluentd-extra-plugins.yaml"
  fi
```

OlleLarsson (approved):

scripts/deploy-wc.sh
simonklb: It was basically like that before but was changed here: https://github.com/elastisys/compliantkubernetes-apps/commit/f0c9eca57f1193c42a3acc7f1871178dfcf4379b Looks intentional to not suppress error messages? @viktor-f could you elaborate?

viktor-f: I don't really remember why this was changed, it was a bit too long ago.
But my guess is that I wanted the error message to show so that you could see why it failed (was the kubeconfig not set? was there some permission issues?).
I could be ok with either keeping the error messages or suppressing them. But would like all of the calls here to be treated in the same way. So if you don't suppress the error messages then it should be the same for the line above this with `if ! kubectl get ns fluentd > /dev/null 2>&1; then`

simonklb: Before fixing this: Are these "fluentd examples" actually still used or are they actually just examples?

OlleLarsson: They are used as they are mounted by the fluentd-user pods.

simonklb: Stopped suppressing error messages: https://github.com/elastisys/compliantkubernetes-apps/pull/1771/commits/9ff49376ff1bb17c7cecbb429a7264458e6e6ebc
Made things use the same syntax: https://github.com/elastisys/compliantkubernetes-apps/pull/1771/commits/3d116c8a570835fa46be37c45c434981b4b201b7

@OlleLarsson @viktor-f PTAL!

viktor-f (approved):

OlleLarsson (approved):

aarnq (approved):

PR #1823, opened 2023-10-13
End-to-end tests for Harbor API

aarnq:
> [!warning]

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

I'm not entirely happy with the execution of these tests, and I'm thinking of reworking them a bit.

I would like the setup of the user-demo image to be a one liner and that practically all that needs can create it's own project/repository/robot to ensure that each test suite has a fresh and predictable environment, and that each test suite can be run independently.
And having it as a before all files may prevent the specific harbor tests to tell what goes wrong.

The clone of the docs repo may be allowed to be a before all files since it will be relatively static.

- Fixes https://github.com/elastisys/ck8s-ops/issues/1849

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:

### DIFF
--- a/tests/common/lib.bash
+++ b/tests/common/lib.bash
@@ -20,6 +20,10 @@ common_setup() {
   load "${ROOT}/tests/common/bats/detik/lib/linter.bash"
   load "${ROOT}/tests/common/bats/detik/lib/utils.bash"
   load "${ROOT}/tests/common/bats/support/load.bash"
+
+  if [[ -z "${CK8S_CONFIG_PATH:-}" ]]; then
+    fail "CK8S_CONFIG_PATH is unset!"
+  fi
 }
 
 # note: not intended for direct use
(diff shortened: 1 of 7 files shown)


tests/common/lib.bash
```diff
@@ -129,6 +129,20 @@ yq_dig() {
   fi
 }
 
+# usage: yq_secret <config-key> <default>
+yq_secret() {
+  if [[ -n "${1:-}" ]]; then
```
Zash, line 134: suggestion:

```suggestion
  if [[ 0 == $# ]]; then
    fail "missing config key argument"
    # assuming this also returns from the function?
```

inverting the condition allows an early return and dropping one level of indentation

tests/common/lib/harbor.bash
```diff
@@ -0,0 +1,141 @@
...
+
+harbor_teardown_user_demo_image() {
+  run harbor_delete_repository "${project}" user-demo
+
+  refute_output
+  assert_success
+}
+
+harbor_get_current_user() {
+  curl -s -u "${harbor_username}:${harbor_password}" -X GET "https://${harbor_endpoint}/api/v2.0/users/current" \
```
Zash, line 99: The author of curl has [opinions](https://daniel.haxx.se/blog/2015/09/11/unnecessary-use-of-curl-x/) about `curl -X` :)

tests/end-to-end/harbor-api.bats
```diff
@@ -0,0 +1,126 @@
...
+  harbor_load_env
+
+  export harbor_endpoint
+
+  export project="end-to-end-test-project"
+  export robot="end-to-end-test-robot"
+
+  export robot_id_path="/tmp/compliantkubernetes-apps-end-to-end-test-robot-id"
+  export robot_name_path="/tmp/compliantkubernetes-apps-end-to-end-test-robot-name"
+  export robot_secret_path="/tmp/compliantkubernetes-apps-end-to-end-test-robot-secret"
```
Zash, line 18: Worth using `tempfile` to pick these names? Trades risk of multiple things using the same files for a risk of a failed test leaving them behind if the tests is somehow interrupted before they're cleaned up.

tests/common/lib/harbor.bash
```diff
@@ -0,0 +1,141 @@
...
+  export docs_path="/tmp/compliantkubernetes"
+
+  if [[ -d "${docs_path}/.git" ]]; then
+    pushd "${docs_path}" || exit 1
+    git checkout -- .
+    git switch main
+    git pull
+    popd || exit 1
+  else
+    git clone "${docs_repo}" "${docs_path}"
```
Zash, line 81:
```suggestion
    git clone --depth=1 "${docs_repo}" "${docs_path}"
```
could reduce space requirements and maybe clone performance as well, assuming full history is not required

tests/common/lib/harbor.bash
```diff
@@ -0,0 +1,141 @@
...
+}
+
+harbor_delete_repository() {
+  curl -s -u "${harbor_username}:${harbor_password}" -X DELETE "https://${harbor_endpoint}/api/v2.0/projects/${1}/repositories/${2}" \
+    -H "accept: application/json"
+}
+
+harbor_create_robot() {
+  curl -s -u "${harbor_username}:${harbor_password}" -X POST "https://${harbor_endpoint}/api/v2.0/robots" \
+    -H "accept: application/json" \
```
Zash, line 122: minor nit: inconsistent capitalization of HTTP headers

tests/common/lib/harbor.bash
```diff
@@ -0,0 +1,141 @@
...
+harbor_get_current_user() {
+  curl -s -u "${harbor_username}:${harbor_password}" -X GET "https://${harbor_endpoint}/api/v2.0/users/current" \
+    -H "accept: application/json"
+}
+
+harbor_create_project() {
+  curl -s -u "${harbor_username}:${harbor_password}" -X POST "https://${harbor_endpoint}/api/v2.0/projects" \
+    -H "accept: application/json" \
+    -H "Content-Type: application/json" \
+    -d "{\"project_name\": \"${1}\"}"
```
Zash, line 107:
```suggestion
    -d "$(jq --arg project_name "${1}" --null-input '{project_name: $project_name}')"
```
This kind of thing should make JSON literal arguments safer to build, tho in tests you could count on nobody ever deciding to put quotes or `\` in these variables...

tests/common/lib/harbor.bash
aarnq: Sure, but it feels like that mainly applies to regular sites when you have redirects.
And setting the method explicitly makes easier to see what is expected of the command, at least in my opinion.

tests/end-to-end/harbor-api.bats
aarnq: That is why I'm using the slug thing, I should perhaps move that to the env load and require it there to make it safer.

The reason why I wanted it to be the same across runs is that then if the process gets killed for some reason it may still successfully clean it up the next run.

Perhaps I should just rework the clean up function do a recursive deletion instead. (The setup will still fail early as it will have conflict if anything is already setup.)

tests/common/lib/harbor.bash
aarnq: :thinking: I did so to allow it do do a simple fetch+pull to keep it up to date and fresh, since it will be used for multiple tests and saved between tests.

Maybe I should just do it in the makefile and it becomes a part of the common setup. Then one would have more control over clearing/updating.

tests/end-to-end/harbor-api.bats
Zash: Unsure how this interacts with bats own cleanup but using [bash traps](https://stackoverflow.com/a/41619743) is one way:
```bash
cleanup () {
rm "/tmp/some-file"
}
trap cleanup EXIT
```

crssnd:
I tried to run `bats end-to-end/harbor-api.bats`, but I get quite a lot of errors for the `assert_line`

> failed with status 6
> lines: parameter not set

Am I missing some steps?

aarnq:
> I tried to run `bats end-to-end/harbor-api.bats`, but I get quite a lot of errors for the `assert_line`
> 
> > failed with status 6
> > lines: parameter not set
> 
> Am I missing some steps?

I see it too, I'm investigating.

aarnq:
I will not fix so the test container can run the build/push/pull tests here. 
It will require some investigations to run either `buildah` or `docker` in the container, especially as we want to run as non root inside the container.

crssnd (approved):

davidumea (approved):

Issue #1807, opened 2023-10-06
Upgrade falco chart to v3.7 and app to v0.36


viktor-f:
**What Chart should be upgraded?**:

falco/falco 3.7.x with app version 0.36.x

If newer versions exists, then you should probably use that instead.

**Why should this Chart be upgraded?**:

- [ ] We need to perform a security upgrade.
- [ ] We need it for a new feature: *which feature (link to blocked issue)*.
- [x] We need to keep up to prevent future issues.
- [ ] Other: *Write the reason here*

As part of the falco v0.36.0 release they did a major rework of how they distribute the rules provided upstream. Instead of having one single file with all available rules they have split it into a few different categories:

- stable (enabled by default): rules that are well maintained and are good at detecting broader threats. Meant to provide a good baseline.
- incubating (disabled by default): more specific use cases, including more application-specific rules.
- sandbox (disabled by default): experimental rules.
- deprecated (disabled by default): rules that, upon re-assessment, were deemed not as applicable to the broader community and are now kept as examples/inspiration. Not actively supported.

We will take this opportunity to reset our falco rules usage. Currently there are plenty of rules that sends a lot of alerts that might not be relevant and that are hard for an operator to look through. With this update we will reduce the number of rules. The goal is to get a subset of rules that will not produce any alarms in a "normal/stable" compliant kubernetes environment (disregarding any user application and actual intrusion attempts). 

To do this, during this task, you should 
1. start with just the rules in the stable category and remove any rule that is creating alerts.
2. Then some combination of
    a. spending some time to modify some of the rules you removed from stable so that they better fit our platform and do not produce alerts
    b. spending some time to look at the alerts in incubating and sandbox and take some that seem very important and does not create alerts after some very minor modifications

Then create new issues to add more rules back into our platform. At least one issue should be focused on alerts that you think are important but that needed more modifications to fit the platform.

**Acceptance criteria**:

- I checked the migration of the new Chart:
  - [ ] I upgraded a Chart and determined that no migration steps are needed.
  - [ ] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).
- [ ] I tested the functionality provided by the new Chart (e.g., Auth flow, Grafana dashboards, etc.)
- [ ] New issues are created to add more rules back into the platform.

PR #1843, opened 2023-10-26
apps: upgrade falco chart and rework exceptions

Eliastisys:
> [!warning]

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

**Which issue this PR fixes**: fixes https://github.com/elastisys/compliantkubernetes-apps/issues/1807

Upgrades falcosecurity/falco to 3.8.3 with app version 0.36.1.
Upgrades falcosecurity/falco-exporter to 0.9.7.
Adds an option to enable/disable upstream rule files in apps config.
Tests falco on compliant with all AMS installed.
Filters out false positives with exceptions while default, sandbox and incubating rules are enabled.

- Disables falco incubating and sandbox [rules](https://github.com/falcosecurity/rules/blob/main/rules/).

#### Checklist

- I checked the migration of the new Chart:
- [x] I tested the functionality provided by the new Chart (e.g., Auth flow, Grafana dashboards, etc.)
- [ ] New issues are created to add more rules back into the platform.

## Note!
New issues should be created to add more rules back into our platform before this is merged.


helmfile/values/falco/falco-common.yaml.gotmpl
```diff
@@ -11,7 +11,10 @@ falco:
 
   grpc_output:
     enabled: true
-
+  config:
+    artifact:
+      install:
+        refs:
```
viktor-f, line 17: Why does `refs` not have a value here? Is that intentional and if so does that evaluate to an empty list? If you just want an empty list then I'd prefer to explicitly write `refs: []`.
Is this what controls which upstream rules files the we install?

helmfile/values/falco/falco-common.yaml.gotmpl
```diff
@@ -54,263 +57,36 @@ falcoctl:
...
-      condition: >
-        ((evt.type = connect and evt.dir=<) or
-          (evt.type in (sendto,sendmsg))) and ssh_port
-        and not (k8s.ns.name = argocd-system and k8s.pod.name startswith argocd-repo-server)
-      output: >
-        Outbound SSH connection (user=%user.name server_ip=%fd.sip server_port=%fd.sport client_ip=%fd.cip)
-      priority: WARNING
-      tags: [network]
   overwrites.yaml: |-
     - macro: k8s_containers
```
viktor-f, line 61: This macro looks like it is not appending to the existing macro and you do not seem to include all of the images that are in the upstream macro. Is that intentional? I think I'd prefer if we appended to the upstream macro unless there are things there that you do not want included.

helmfile/values/falco/falco-common.yaml.gotmpl
```diff
@@ -54,263 +57,36 @@ falcoctl:
...
-    - list: known_binaries_to_read_environment_variables_from_proc_files
-      append: true
-      items: [systemd-run, rook, udevadm]
-    - list: known_drop_and_execute_containers
+
+    - rule: Run shell untrusted
       append: true
-      items:
-        - ghcr.io/elastisys/calico-accountant # when .calicoAccountant.backend = nftables
+      condition: and not k8s_containers
```
viktor-f, line 88: Instead of modifying the rule I would like you to use the macro `user_shell_container_exclusions` that is already part of the rule.
See https://github.com/falcosecurity/rules/blob/e206c1a3f12c310c38afcf6af9c311c312791804/rules/falco_rules.yaml#L601C13-L601C44

helmfile/values/falco/falco-common.yaml.gotmpl
Eliastisys: I think the only reason is that it wasn't appended [previously](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile/values/falco/falco-common.yaml.gotmpl#L80-L83). But i think it makes sense to append it, so we'll do that instead.

helmfile/values/falco/falco-common.yaml.gotmpl
```diff
@@ -54,263 +54,45 @@ falcoctl:
...
-      condition: >
-        ((evt.type = connect and evt.dir=<) or
-          (evt.type in (sendto,sendmsg))) and ssh_port
-        and not (k8s.ns.name = argocd-system and k8s.pod.name startswith argocd-repo-server)
-      output: >
-        Outbound SSH connection (user=%user.name server_ip=%fd.sip server_port=%fd.sport client_ip=%fd.cip)
-      priority: WARNING
-      tags: [network]
   overwrites.yaml: |-
+    - list: trusted_image_repositories
```
viktor-f, line 58: nit: maybe rename this to something that makes it a little more clear that it belongs to the `k8s_containers` macro.

viktor-f (commented):
Which rules are included with this? Is it everything in the default rules file but nothing from the incubating or sandbox files? Please note this in the PR message as well so that it will be part of the changelog.

helmfile/values/falco/falco-common.yaml.gotmpl
LundqvistAndre: This was intentionally left a bit agnostic since it may be used in other exceptions later. (When working with the old rules this list was used in multiple exceptions, since we are looking to enable them later we chose to leave it as is)

Eliastisys:
> Which rules are included with this? Is it everything in the default rules file but nothing from the incubating or sandbox files? Please note this in the PR message as well so that it will be part of the changelog.

Only default rules are used. Noted in PR message. @viktor-f

helmfile/values/falco/falco-common.yaml.gotmpl
viktor-f: Ah, ok. Then I guess it's fine.
But then I think I would like a comment just above that list which informs about which rules this should apply to. So that you know what happens if you add an image to the list.

helmfile/values/falco/falco-common.yaml.gotmpl
```diff
@@ -12,6 +12,12 @@ falco:
   grpc_output:
     enabled: true
 
+  rules_file:
+    - /etc/falco/falco_rules.yaml
+    - /etc/falco/falco-incubating_rules.yaml
+    - /etc/falco/falco-sandbox_rules.yaml
+    - /etc/falco/falco_rules.local.yaml
+    - /etc/falco/rules.d
```
viktor-f, line 20: Is it fine to specify all of these even when you have disabled some of them?

Eliastisys: It is fine in the sense that nothing breaks. It will create empty files in the containers, so maybe we want some checks here as well

Eliastisys: https://github.com/elastisys/compliantkubernetes-apps/pull/1843/commits/83a19273b9f21e2076c7981e35a5d9a0b240080f#diff-e91a3a06c409a58ed8e2eb4e6dbe9e8931306dedb3d4c09756344e8636dacf8bR15-R26

viktor-f: If it doesn't break anything then it's fine, but it does look better with the checks :+1:

config/config/common-config.yaml
```diff
@@ -120,9 +131,7 @@ falco:
   ## additional falco rules
   ## ref: https://falco.org/docs/rules/
   customRules:
-    non-sudo-setuid.yaml: |-
-      - rule: Non sudo setuid
-        enabled: false
```
viktor-f, line 125: Just to make this more clear I think I'd want an example here in a comment. Something like this, but feel free to update it:
```
customRules: {}
  # my-rules-example.yaml : |-
  # - macro: <name of macro to add/change> # possible to use rules and lists as well
  #   condition: <add condition here>
  #   ...
```

viktor-f (commented):
I think this looks good.
Since this disables the incubating and sandbox rules (that previously were enabled as part of the old single rules file), I think that should be noted in the PR and it should be noted as a `release note` or `admin-change`.

viktor-f:
You should add the release note or admin change in the PR description instead (there are some headers that you can uncomment from the template) and you can remove the text in the wip-changelog (that files is deprecated).

Eliastisys:
> You should add the release note or admin change in the PR description instead (there are some headers that you can uncomment from the template) and you can remove the text in the wip-changelog (that files is deprecated).

@viktor-f Does it look ok now?

viktor-f (approved):
LGTM (note that I have not tested it)

Pavan-Gunda (approved):

PR #1778, opened 2023-09-18
Allow per cluster upgrades

anders-elastisys:
> [!warning]
> **This is public repository, ensure not to disclose:**
> - [ ] personal data beyond what is necessary for interacting with this pull request
> - [ ] business confidential information, such as customer names

### What kind of PR is this?

**Required**: Mark one of the following that is applicable:

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

*Optional*: Mark one or more of the following that are applicable:

> [!important]
> Breaking changes should be marked `kind/admin-change` or `kind/dev-change` depending on type
> Critical security fixes should be marked with `kind/security`

- [x] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

### Platform Administrator notice

This PR changes the number of arguments required to run the `upgrade` command, as you now have to specify the specific cluster, `sc`, `wc`, or `both` to apply to both clusters as the command did previously. The ck8s usage has been updated to clarify this:
```bash
➜ ./bin/ck8s                      
COMMANDS:
  ...
  upgrade <wc|sc|both> <vX.Y> prepare               runs all prepare steps upgrading the configuration
  upgrade <wc|sc|both> <vX.Y> apply                 runs all apply steps upgrading the environment
```

Migration templating have also been updated to support and clarify these changes.

### What does this PR do / why do we need this PR?

This PR allows for running the `upgrade` command for each cluster (`wc` and `sc`) separately, which is particularly useful for multi-workload cluster environments, or environments without workload clusters. Previously the `upgrade` command always applies for both `wc` and `sc`.

- Fixes #1472 

#### Additional information to reviewers
Note that this PR only fixes this for `upgrade apply`, as `upgrade prepare` requires more work as it uses the `init` command which currently does not support per cluster initialization (see [this issue](https://github.com/elastisys/compliantkubernetes-apps/issues/1777)).

#### Checklist

- [x] Proper commit message prefix on all commits
  
- Change checks:
  - [ ] The change is transparent
  - [ ] The change is disruptive
  - [ ] The change requires no migration steps
  - [ ] The change requires migration steps
- Metrics checks:
  - [ ] The metrics are still exposed and present in Grafana after the change
  - [x] The metrics names didn't change (Grafana dashboards and Prometheus alerts are not affected)
[...]

PR #1845, opened 2023-10-27
Allow per cluster init

### LINKED ISSUE #1777: Allow init command to run per cluster
Similar to [this](https://github.com/elastisys/compliantkubernetes-apps/issues/1472), we would like to be able to run the `init` script for each cluster separately.

Should be able to specify `wc` or `sc` when running `init` command to only initialize for the specified cluster, i.e:
```
./bin/ck8s init wc
./bin/ck8s init sc
```

**Additional context**

**Definition of done:**
It is possible to run `init` command per cluster (`sc` or `wc`).

anders-elastisys:
> [!warning]

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

The `ck8s init` command now requires a cluster argument which can be either of `<wc|sc|both>`.

Allow for running the `init` command per cluster. This also fixes issue with running the `fix-psp-violations` command which got introduced with [this PR](https://github.com/elastisys/compliantkubernetes-apps/pull/1778).

- Fixes #1777 

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:
- Bug checks:

### DIFF
--- a/bin/ck8s
+++ b/bin/ck8s
@@ -10,7 +10,7 @@ source "${here}/common.bash"
 
 usage() {
     echo "COMMANDS:" 1>&2
-    echo "  init [--generate-new-secrets]                     initialize the config path" 1>&2
+    echo "  init <wc|sc|both> [--generate-new-secrets]        initialize the config path" 1>&2
     echo "  bootstrap <wc|sc>                                 bootstrap the cluster" 1>&2
     echo "  apps <wc|sc> [--sync] [--skip-template-validate]  deploy the applications" 1>&2
     echo "  apply <wc|sc> [--sync] [--skip-template-validate] bootstrap and apps" 1>&2
@@ -58,7 +58,9 @@ done
 
 case "${1}" in
     init)
+        [[ "${2}" =~ ^(wc|sc|both)$ ]] || usage
         check_tools
+        export CK8S_CLUSTER="${2}"
         "${here}/init.bash" "${GEN_NEW_SECRETS}"
         ;;
     bootstrap)
@@ -92,7 +94,8 @@ case "${1}" in
         [[ "${3}" =~ ^(v[0-9]+\.[0-9]+)$ ]] || usage
         [[ "${4}" =~ ^(prepare|apply)$ ]] || usage
         check_tools
-        "${here}/upgrade.bash" "${2}" "${3}" "${4}"
+        export CK8S_CLUSTER="${2}"
+        "${here}/upgrade.bash" "${3}" "${4}"
         ;;
     team)
         case "${2}" in
@@ -159,7 +162,10 @@ case "${1}" in
         "${here}/update-ips.bash" "${2}" "${3}"
         ;;
     fix-psp-violations)
-        "${here}/fix-psp-violations.bash" "${2}"
+        [[ "${2}" =~ ^(wc|sc)$ ]] || usage
+        check_tools
+        export CK8S_CLUSTER="${2}"
+        "${here}/fix-psp-violations.bash"
         ;;
     clean)
         "${here}/clean.bash" "${2}"
--- a/bin/fix-psp-violations.bash
+++ b/bin/fix-psp-violations.bash
@@ -1,9 +1,6 @@
 #!/usr/bin/env bash
 
-usage() {
-    echo "Usage: fix-psp-violoations <wc|sc>" >&2
-    exit 1
-}
+: "${CK8S_CLUSTER:?Missing CK8S_CLUSTER}"
 
 ROOT="$(readlink -f "$(dirname "${0}")/../")"
 
@@ -17,10 +14,9 @@ exempt_namepsaces=("rook-ceph")
 function set_violating_resources() {
   results_mult_uniq=""
   results=()
-  cluster="$1"
 
   # Get violations for PSPs
-  violations=$(kubectl_do "$cluster" get constraints -o yaml | yq4 '[.items[] | select(.kind == "K8sPSP*") | .status.violations[]]')
+  violations=$(kubectl_do "${CK8S_CLUSTER}" get constraints -o yaml | yq4 '[.items[] | select(.kind == "K8sPSP*") | .status.violations[]]')
 
   # Build array of maps with relevant information
   resources=$(echo "$violations" | yq4 '.[] |[{"name": .name, "namespace": .namespace}]' | yq4 'unique_by(.name,.namespace)')
@@ -32,7 +28,7 @@ function set_violating_resources() {
     namespace=$(echo "$resource" | yq4 e '.namespace')
     pod_name=$(echo "$resource" | yq4 e '.name')
 
-    owner_reference=$(kubectl_do "$cluster" -n "$namespace" get pod "$pod_name" --ignore-not-found=true -oyaml | yq4 '.metadata.ownerReferences.[0]')
+    owner_reference=$(kubectl_do "${CK8S_CLUSTER}" -n "$namespace" get pod "$pod_name" --ignore-not-found=true -oyaml | yq4 '.metadata.ownerReferences.[0]')
 
     # Skip standalone Pods and stale references
     if [ "$owner_reference" = "null" ] || [ -z "$owner_reference" ]; then continue; fi
@@ -45,7 +41,7 @@ function set_violating_resources() {
 
     # Get owner of ReplicaSets
     if [ "$owner_kind" == "ReplicaSet" ]; then
-      owner_reference=$(kubectl_do "$cluster" -n "$namespace" get rs "$owner_name" --ignore-not-found=true -oyaml | yq4 '.metadata.ownerReferences.[0]')
+      owner_reference=$(kubectl_do "${CK8S_CLUSTER}" -n "$namespace" get rs "$owner_name" --ignore-not-found=true -oyaml | yq4 '.metadata.ownerReferences.[0]')
 
       # Skip standalone ReplicaSets and stale references
       if [ "$owner_reference" = "null" ] || [ -z "$owner_reference" ]; then continue; fi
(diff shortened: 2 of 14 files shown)


pipeline/README.md
```diff
@@ -3,8 +3,11 @@
 If you want to run Bats tests locally (instead of in the CI pipeline), execute the following from the project root directory:
 
 ```bash
+git submodule update --recursive --init # init and update bats helpers
+
 docker build -t ck8s-apps ./pipeline
-docker run --rm -it -v "$(pwd):/ck8s-apps:ro" -w /ck8s-apps ck8s-apps \
+docker run --rm -it -e CK8S_PGP_FP=529D964DE0BBD900C4A395DA09986C297F8B7757 \
```
simonklb, line 9: Who's fingerprint is this?

scripts/migration/lib.sh
```diff
@@ -18,6 +18,7 @@ SC_CONFIG_FILES=(
   "defaults/sc-config.yaml"
   "common-config.yaml"
   "sc-config.yaml"
+  "secrets.yaml"
```
simonklb, line 21: What changed to require this change? Looks unrelated to this PR but I could be wrong.

bin/fix-psp-violations.bash
```diff
@@ -7,6 +7,11 @@ usage() {
 
 ROOT="$(readlink -f "$(dirname "${0}")/../")"
 
+export CK8S_CLUSTER="${1}"
```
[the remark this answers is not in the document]
anders-elastisys: With the changes of [this PR here](https://github.com/elastisys/compliantkubernetes-apps/pull/1778/files#diff-d157d0f7a32fe5a99591a42778e449309300204efe637caef726eefdceb9f0deR181-R186), since the `fix-psp-violations.bash` script sources the `lib.sh` script which runs `check_config`, `CK8S_CLUSTER` needs to be set, otherwise you get this:
```console
✗ ./bin/ck8s fix-psp-violations wc  
[ck8s] fix-psp-violations.bash: error: "CK8S_CLUSTER" is unset
```

But I agree moving it one level up for all commands might be good.

bin/init.bash
```diff
@@ -8,6 +8,9 @@
 set -eu -o pipefail
 
 here="$(dirname "$(readlink -f "$0")")"
+
+export CK8S_CLUSTER="${1}"
```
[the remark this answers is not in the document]
anders-elastisys: It does not need to be exported, but if we move the variable one level up (to `ck8s` base script) we can remove declaring this variable from this file.

migration/template/prepare/50-init.sh
```diff
@@ -6,4 +6,17 @@ ROOT="$(readlink -f "${HERE}/../../../")"
...
+  sc)
+    "${ROOT}/bin/ck8s" init sc
+    ;;
+  wc)
+    "${ROOT}/bin/ck8s" init wc
+    ;;
+  *)
+    log_fatal "usage: 50-init.sh <wc|sc|both>"
+    ;;
+esac
```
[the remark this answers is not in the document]
anders-elastisys: Yes, that is nicer

pipeline/README.md
anders-elastisys: It is the same as in the workflow file [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/.github/workflows/pipeline.yml#L73), it was previously found [here](https://github.com/elastisys/compliantkubernetes-apps/commit/7935222afbb6695a4a5406bd15d1e3e75431a725#diff-ac4a92945bb0aa4c42cfd49d29a88c433c8a52b85bded34032cc66184271ea2dL16)

bin/init.bash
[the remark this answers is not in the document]
simonklb: I think that makes sense. Ideally I would like it to be environment variables all the way but since we are used to passing arguments for the other commands this isn't the time for that big of a change I guess. :smile:

pipeline/README.md
simonklb: Ah. Do you know why it needs to be set there when it's being set in the `init.bats` file as well?

scripts/migration/lib.sh
simonklb: Gotcha! In the future I would suggest to send such a separate fix as a separate PR to make it clear since GitHub doesn't have a good "commit-by-commit" review system but instead makes you review all the changes at once. Not a blocker here though. Just don't squash it with the rest of the commits. :)

pipeline/README.md
anders-elastisys: Sorry if I wasn't clear, it is no longer set in `init.bats`, it got moved  to the workflow file, so when you try to run it locally one of the tests fails if you do not provide it as an environment variable.

bin/init.bash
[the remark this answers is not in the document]
anders-elastisys: @simonklb I looked into moving the `CK8S_CLUSTER` global variable to the `ck8s` script, please have another look. If this looks better, I can perhaps add the same logic to the other commands that take a _cluster_ name as an argument, or, if that perhaps should be left for another issue.

bin/ck8s
```diff
@@ -56,8 +56,10 @@ for arg in "$@"; do
   esac
 done
 
+export CK8S_CLUSTER="${2}"
```
simonklb, line 59: This won't be correct for commands that takes something other than a cluster as the second argument. For example `ck8s ops kubectl ...`.

simonklb (changes requested):

bin/ck8s
anders-elastisys: Yes, it would in that case be set something like this to override the variable:
```
    ops)
        CK8S_CLUSTER="${3}"
        case "${2}" in
```
But like I said, none of the other commands support using this global environment variable atm, so this would need more work either in this issue or in another issue

bin/init.bash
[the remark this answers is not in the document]
simonklb: I wouldn't mind if you made them rest of the commands use `CK8S_CLUSTER` instead of a positional argument. Up to you!

bin/ck8s
anders-elastisys: Otherwise you could export it for each command that support it like I initially did in this commit: https://github.com/elastisys/compliantkubernetes-apps/pull/1845/commits/593ef62aaa21f7a21ec0c0989cf4e37e13451a7c

simonklb: Again, I would have loved that it was simple a variable you set in your shell which defaulted to `both` instead of an argument but that's for the future I guess... :smile:

robinAwallace: Why are we using a global variable instead of passing it to functions/scripts?
To me it feels a bit dangerous expecting global variables to be set.

simonklb: > Why are we using a global variable instead of passing it to functions/scripts? To me it feels a bit dangerous expecting global variables to be set.

How so? The risk of accidentally writing the wrong letters when you set the variable vs writing the argument seems equally as dangerous. Similarly forgetting to set the variable to the right cluster or to rewrite your copy-pasted line seems equally as dangerous as well.

In this case neither is a problem though because the command-line isn't changed, only the variable passing.

simonklb: If I had rewritten this from scratch I would have made it configurable both using command-line flags and environment variables. Not positional arguments.

anders-elastisys: > Why are we using a global variable instead of passing it to functions/scripts? To me it feels a bit dangerous expecting global variables to be set.

Scripts that sources the `scripts/migration/lib.sh` now needs a global environment variable `CK8S_CLUSTER` to be able to validate configuration files etc per cluster. I changed it this way in a [previous PR](https://github.com/elastisys/compliantkubernetes-apps/commit/6cd8a4ee0692703df251dbc23bce804e00126618#diff-d157d0f7a32fe5a99591a42778e449309300204efe637caef726eefdceb9f0de) as a means to allow _per cluster_ commands, as I did not see any other good approach for solving this.

anders-elastisys: @simonklb I updated the PR, can I mark this as solved?

bin/init.bash
```diff
@@ -8,6 +8,7 @@
 set -eu -o pipefail
```
simonklb, line 8: Make sure that `CK8S_CLUSTER` is set.

```suggestion
: "${CK8S_CLUSTER:?Missing CK8S_CLUSTER}"

set -eu -o pipefail
```

bin/fix-psp-violations.bash
```diff
@@ -1,10 +1,5 @@
 #!/usr/bin/env bash
```
simonklb, line 1:
```suggestion
#!/usr/bin/env bash

: "${CK8S_CLUSTER:?Missing CK8S_CLUSTER}"
```

bin/upgrade.bash
```diff
@@ -6,7 +6,6 @@ ROOT="$(readlink -f "${here}/../")"
```
simonklb, line 6:
```suggestion

: "${CK8S_CLUSTER:?Missing CK8S_CLUSTER}"

```

simonklb (changes requested):

simonklb (approved):
Nice work!

bin/init.bash
```diff
@@ -235,18 +238,18 @@ set_nginx_config() {
             ;;
     esac
 
-    replace_set_me "$1" '.ingressNginx.controller.config.useProxyProtocol' "${use_proxy_protocol}"
-    replace_set_me "$1" '.ingressNginx.controller.useHostPort' "${use_host_port}"
-    replace_set_me "$1" '.ingressNginx.controller.service.enabled' "${service_enabled}"
-    replace_set_me "$1" '.networkPolicies.global.externalLoadBalancer' "${external_load_balancer}"
-    replace_set_me "$1" '.networkPolicies.global.ingressUsingHostNetwork' "${ingress_using_host_network}"
+    replace_set_me "$file" '.ingressNginx.controller.config.useProxyProtocol' "${use_proxy_protocol}"
```
aarnq, line 241: Nit
```suggestion
    replace_set_me "${file}" '.ingressNginx.controller.config.useProxyProtocol' "${use_proxy_protocol}"
```

aarnq (approved):
LGTM.

bin/init.bash
anders-elastisys: @aarnq please have a look at the [latest commit](https://github.com/elastisys/compliantkubernetes-apps/pull/1845/commits/5e5bc96d618870595343c895603b9be0055f1afe), resolve this if it looks fine and I can close this PR

aarnq: Looks good! I even got more than I asked for :smile:

PR #1847, opened 2023-10-30
tests: Migrate pipeline unit tests

aarnq:
> [!warning]

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

Migrates the init tests to run under the new test harness.
Additionally it uses transient GPG keys as an alternative to previous static GPG key.

- Fixes #

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:
- Bug checks:


aarnq:
Question to reviewers: Do we think we should keep parts of the matrix in the old workflow to split it against different cloud providers?

This could both speed up the checks and make it more obvious in which type of configuration the checks fail.

(The image building time will improve as I will add caching via the cache-image workflow.)

tests/Dockerfile
```diff
@@ -47,16 +47,19 @@ RUN wget --progress=dot:giga "https://github.com/mikefarah/yq/releases/download/
     install -Tm 755 yq_linux_amd64 /usr/local/bin/yq4 && \
     rm yq_linux_amd64
 
+# RUN mkdir -p /apps/tests
+# RUN chown -R ubuntu:ubuntu /apps
+
```
simonklb, line 52: Remove?

tests/Makefile
```diff
@@ -2,32 +2,43 @@
...
 # Template tests
-template_suites := $(foreach dir, $(dirs), $(wildcard $(dir)/*.bats.yaml))
+template_suites := $(foreach dir, $(dirs), $(shell find $(dir) -type f -name *.bats.yaml))
 template_gen := $(template_suites:.bats.yaml=.gen.bats)
 
-config_path := ${CK8S_CONFIG_PATH}
+endif
 
-container_image := elastisys-ck8s-apps-tests:main
+container_image := compliantkubernetes-apps/tests:main
```
simonklb, line 21: I think this only works because the "repository" will be `compliantkubernetes-apps`? I guess it's fine but it's a bit of a misuse, no?

tests/Makefile
```diff
@@ -36,7 +47,7 @@ found_apt := $(shell command -v apt 2>/dev/null)
 found_bats := $(shell command -v bats 2>/dev/null)
 found_npm := $(shell command -v npm 2>/dev/null)
 
-found_podman := $(shell command -v podman 2>/dev/null)
+found_podman := $(shell docker --version 2>/dev/null | grep podman &>/dev/null)
```
simonklb, line 50: Does the podman binary always shadow `docker`?

tests/Makefile
```diff
@@ -141,9 +177,9 @@ clean-gen:
 
 build:
 ifdef found_podman
-	podman build -t $(container_image) .
+	podman build --env CK8S_TESTS_BUILD=true --file tests/Dockerfile --tag $(container_image) ../
```
simonklb, line 180: Here you seem to be using `podman` but when checking if it exists you changed it to `docker --version`. Wouldn't `command -v podman` have better accuracy?

tests/common/lib/bin.bash
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+
+# Helpers to run templated bin tests
```
simonklb, line 3: Something rubs me the wrong way calling these "helpers" since they are very specific tests. It's not like they can be shared across different tests, right?

simonklb (changes requested):

tests/Makefile
aarnq: If I recall correctly. I'll double check.

tests/Makefile
aarnq: I'll change it to a dash instead of a slash, I keep forgetting that containerd and docker defaults to dockerhub for most things.

tests/Makefile
aarnq: I'll rework that check because I think I just wanted to shoehorn in to run docker first then podman but I didn't really think about different cases.

tests/common/lib/bin.bash
simonklb: Aren't these templates since they are used to render the templated tests? :smile:

tests/Makefile
aarnq: Reworked the checks entirely now so that should improve the remark below as well.

tests/common/lib/bin.bash
aarnq: Not really, these are plain test functions that the templated tests are using.
I'll see if I can rework the generator to add functions.

simonklb: I see, they look "complete", as in what else does the generator add to them? I should probably generate the tests and just see for myself. :smile:

simonklb: Let's not waste more time bikeshedding over this though haha.

tests/common/gen-bats.bash
```diff
@@ -59,30 +60,34 @@ render_test() {
...
-        echo "@test \"${name} - ${function//_/ } - ${cluster//\"/} / ${namespace//\"/} / ${target} - $(jq -r .[0] <<< "${args}")\" {"
+        echo "@test \"${identifier} - ${cluster//\"/} / ${namespace//\"/} / ${target} - $(jq -r .[0] <<< "${args}")\" {"
       else
-        echo "@test \"${name} - ${function//_/ } - ${cluster//\"/} / ${namespace//\"/} / ${target}\" {"
+        echo "@test \"${identifier} - ${cluster//\"/} / ${namespace//\"/} / ${target}\" {"
       fi
       render_conditions "${cluster}" "${conditions}"
-      echo "  with_kubeconfig ${cluster}"
-      echo "  with_namespace ${namespace}"
+      if [[ -n "${clusters#[]}" ]]; then
```
simonklb, line 85: Not sure I understand the reason for this. The way I interpret it is that you apply a "match nothing" pattern on the first element of the array which won't delete anything from it. What's different from just `-n "${clusters}"`?

Or in this case, couldn't we even assume that it never will be empty since we are looping over the array already?

simonklb (changes requested):

tests/common/gen-bats.bash
aarnq: I didn't even realise how strange I did this. It works because an empty one will be `[]`, and if we have that then we don't render the `with`.

simonklb: Are you sure you're not thing of `${foo:-[]}`? That would result in `[]` if it's empty. `${foo#[]}` deletes with pattern matching AFAIK.

aarnq: I am bad at explaining today :sweat_smile: 

If `clusters` or `namespaces` is unset in input file then the variable will be set to `[]` (an empty array) and that is the condition I mean. `${var#[]}` does indeed remove the leading `[]` when it is expanded, and thus if `clusters` or `namespaces` with this expand to nothing then I consider it empty.

I should really just check `cluster` and `namespace` if they are empty. :grin:

simonklb: Now it all makes sense! :)

tests/unit/bin/update-ips/resources/maximal-run-full-diff.out
```diff
@@ -1,63 +1,61 @@
...
-   baseDomain: dev-ck8s.com
-[32m+networkPolicies:[0m
-[32m+  global:[0m
-[32m+    objectStorage:[0m
-[32m+      ips:[0m
+[36m@@ -22,7 +22,7 @@[0m
+   global:
+     objectStorage:
+       ips:
+[31m-        - set-me[0m
```
simonklb, line 7: Why has the result changed?

tests/unit/bin/init.bats
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env bats
```
simonklb, line 1: Sorry if I've already asked this before. Why do we need this test when we have the same tests as template tests in `init.bats.yaml`?

simonklb (approved):
Nice work!

Mostly curious about the change in the test output for update-ips. I would have expected a migration not changing the results in any way.

I'm fine with shipping this though! :)

tests/unit/bin/init.bats
aarnq: It is one one of the tests that are overlapping, and that will be changed when the `none` cloud provider is added as I will change this one to use it instead.

But the reason why it still was included was because it makes more sense to have the test first that all files are created, then in a second test check that all files are backed up.

tests/unit/bin/update-ips/resources/maximal-run-full-diff.out
aarnq: Because the basis for the test now uses a config generated by init instead of a completely empty one. And that sets `set-me`s for networkpolices prompting for the configuration.

Also the base config changed as I worked to make it as much reusable as possible, and change `"true"`s into boolean `true`s.

tests/unit/bin/init.bats
simonklb: Makes sense!

OlleLarsson (approved):

lucianvlad (approved):

PR #1858, opened 2023-11-07
Update default upcloud storage class to "standard"

viktor-f:
> [!warning]

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

The default config for storage class name on upcloud has changed from "upcloud-block-storage" to "standard". This will impact prometheus and log-manager PVCs if they are enabled.

The default storage class elastisys generally uses on Upcloud is called "standard", not "upcloud-block-storage", so the config here has been updated to match.

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:
- Bug checks:

### DIFF
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -142,7 +142,7 @@ set_storage_class() {
             ;;
 
         upcloud)
-            storage_class=upcloud-block-storage
+            storage_class=standard
 
             yq4 -i '.networkPolicies.kubeSystem.upcloud.enabled = true' "${file}"
             yq4 -i '.networkPolicies.kubeSystem.upcloud.ips = ["94.237.0.0/23"]' "${file}" # api.upcloud.com
--- a/migration/template/README.md
+++ b/migration/template/README.md
@@ -36,6 +36,7 @@ As with all scripts in this repository `CK8S_CONFIG_PATH` is expected to be set.
 
 ## Prerequisites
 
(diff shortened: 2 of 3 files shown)


bin/init.bash
```diff
@@ -142,7 +142,7 @@ set_storage_class() {
             ;;
 
         upcloud)
-            storage_class=upcloud-block-storage
+            storage_class=standard
```
aarnq, line 145: So this will change in the defaults if it hasn't been overridden?
Then we should add an admin notice that the default storage class has changed so it doesn't come as a surprise for anyone that relies on the current default.
Or make a snippet that sets the previous default as an override if it previously wasn't set.

robinAwallace: Good point André :+1: Could we create a snippet that looks at the current storage and if it is not "standard" add a override?

viktor-f: I'll write an admin notice about this.
I will also update the migration template to include reading the changelog/release notes.

aarnq (approved):

Xartos (approved):

Issue #1750, opened 2023-09-01
Improve the logic of update-ips script


aarnq:
**Is your feature request related to a problem? Please describe.**
With the work of #1582 done we're improved the functions and general handling of IPs a lot, so now we should focus on improving the logic behind it.

**Describe the solution you'd like**
Better grouping of checks so that as an example all specific for SC may be under the same if statement.
Reusable function for looking up keystone+swift endpoints.
Reusable functions for extracting domain name and ports from endpoints.
Simplify the logic for rclone sync.
Moving things to functions to simplify structure and make it more digestible.

**Describe alternatives you've considered**
Keep current which is harder to maintain.

**Definition of done:**
The update-ips script is further improved.

PR #1873, opened 2023-11-20
Lookup Swift endpoint for RClone sync in update-ips script

simonklb:
> [!warning]

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

(OpenStack only) You should see a new Network Policy for `.networkPolicies.rcloneSync.destinationObjectStorageSwift` after running `update-ips` if Swift and Keystone have separate IPs and/or ports.

Before this only the auth/Keystone endpoint would be allow-listed when updating the network policy configuration for RClone sync using Swift.

This might not have been an issue as long as the main object storage also was Swift since .networkPolicies.global.objectStorageSwift would have also been set (which performed the Swift endpoint lookup).

- Part of https://github.com/elastisys/compliantkubernetes-apps/issues/1750

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:
- Bug checks:

### DIFF
--- a/bin/update-ips.bash
+++ b/bin/update-ips.bash
@@ -161,17 +161,21 @@ get_dns_ips() {
 }
 
 # Fetch the Swift URL.
+#
+# Usage: get_swift_url <swift_config_option>
 get_swift_url() {
+  local swift_config_option="${1}"
+
   local auth_url
   local os_token
   local swift_url
   local swift_region
   local response
 
-  auth_url="$(yq_read "sc" '.objectStorage.swift.authUrl' "")"
-  swift_region="$(yq_read "sc" '.objectStorage.swift.region' "")"
+  auth_url="$(yq_read "sc" "${swift_config_option}.authUrl" "")"
+  swift_region="$(yq_read "sc" "${swift_config_option}.region" "")"
 
-  if [ -n "$(yq_read_secret '.objectStorage.swift.username' "")" ]; then
+  if [ -n "$(yq_read_secret "${swift_config_option}.username" "")" ]; then
     response=$(curl -i -s -H "Content-Type: application/json" -d '
         {
           "auth": {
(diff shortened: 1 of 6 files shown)


bin/update-ips.bash
```diff
@@ -522,7 +534,9 @@ if [[ "${check_cluster}" =~ ^(sc|both)$ ]]; then
   allow_nodes "sc" '.networkPolicies.global.scApiserver.ips' "node-role.kubernetes.io/control-plane="
   allow_nodes "sc" '.networkPolicies.global.scNodes.ips' ""
 
-  allow_swift
+  if swift_enabled; then
+    sync_swift '.objectStorage.swift.authUrl' '.networkPolicies.global.objectStorageSwift'
```
aarnq, line 538:
```suggestion
    sync_swift '.objectStorage.swift' '.networkPolicies.global.objectStorageSwift'
```
I think you need to do something like this so you can pass it on to `get_swift_url`, else you will not run with the correct credentials for `.objectStorage.sync.swift` below.

pipeline/Dockerfile
```diff
@@ -24,10 +24,10 @@ RUN wget --progress=dot:giga "https://github.com/yannh/kubeconform/releases/down
 
 # sops
 # NOTE: This needs to be installed before the helm-secrets plugin.
-ENV SOPS_VERSION="3.7.3"
-RUN wget --progress=dot:giga "https://github.com/getsops/sops/releases/download/v${SOPS_VERSION}/sops-v${SOPS_VERSION}.linux" && \
-    install -Tm 755 "sops-v${SOPS_VERSION}.linux" /usr/local/bin/sops && \
-    rm "sops-v${SOPS_VERSION}.linux"
+ENV SOPS_VERSION="3.8.1"
```
Xartos, line 27: Should we update https://github.com/elastisys/compliantkubernetes-apps/blob/main/roles/get-requirements.yaml#L13 as well?

Xartos (dismissed):

pipeline/Dockerfile
simonklb: I don't know the impact of it for the rest of the repository so that should probably be a different task?

Xartos: I just feel that we should try to keep the pipeline as close to a dev environment as possible and in that case we'd like to keep these two versions the same. But I guess if we get a task for looking into that now then it'll soon become synced again

simonklb: Yea that is a good point. We probably should install the dependencies the same way then, to prevent this.

simonklb: I think these tests are going to be migrated to the new test framework @aarnq is working on though, so I won't create any issues until things have settled.

Xartos (approved):

aarnq (approved):

PR #1883, opened 2023-11-23
bin: set helmfile concurrency to avoid out of memory issues

### LINKED ISSUE #1880: Add ability to pass arguments to apps apply to avoid helmfile OOM
The self-managed team have had issues with helmfile diff crashing due to OOM when running `ck8s apply`.

This can be easily fixed by adding the flag `--concurrency <num>` to the helmfile apply command.

Either always set this or allow one to pass it as an argument to `ck8s apply`.

**Additional context**

**Definition of done:**
Should be possible in one way or another to run helmfile apply with the concurrency flag.

Elias-elastisys:
> [!warning]

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

The commands `apps` and `apply` now sets helmfile concurrency to 8 by default to avoid memory issues

Sets helmfile concurrency to avoid oom issues with diff.

- Fixes #1880 

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:
- Bug checks:

### DIFF
--- a/bin/apps.bash
+++ b/bin/apps.bash
@@ -24,7 +24,7 @@ apps_run_sc() {
     log_info "Applying applications in service cluster"
     (
         with_kubeconfig "${config[kube_config_sc]}" \
-            "${scripts_path}/deploy-sc.sh" "${1:-""}"
+            "${scripts_path}/deploy-sc.sh" "${@}"
     )
 }
 
@@ -32,7 +32,7 @@ apps_run_wc() {
     log_info "Applying applications in workload cluster"
     (
         with_kubeconfig "${config[kube_config_wc]}" \
-            "${scripts_path}/deploy-wc.sh" "${1:-""}"
+            "${scripts_path}/deploy-wc.sh" "${@}"
     )
 }
 
@@ -71,8 +71,7 @@ apps_sc() {
     # but feels good enough until we figure out something smarter.
     #
     #[ "$1" != "--skip-template-validate" ] && template_validate_sc
-
-    apps_run_sc "${2:-""}"
+    apps_run_sc "${2:-""}" "$3"
 
     log_info "Applications applied successfully!"
 }
@@ -82,7 +81,7 @@ apps_wc() {
     # See rationale above
     #[ "$1" != "--skip-template-validate" ] && template_validate_wc
 
-    apps_run_wc "${2:-""}"
+    apps_run_wc "${2:-""}" "$3"
 
     log_info "Applications applied successfully!"
 }
@@ -103,7 +102,7 @@ if [[ $1 == "wc" ]]; then
         fi
     fi
     config_load "$1"
-    apps_wc "$2" "$3"
+    apps_wc "$2" "$3" "$4"
 elif [[ $1 == "sc" ]]; then
     if ! "${here}/update-ips.bash" "sc" "dry-run"; then
         log_warning "Diff in sc config for network policy IPs, run 'ck8s update-ips sc apply' to fix this issue."
@@ -117,7 +116,7 @@ elif [[ $1 == "sc" ]]; then
         fi
     fi
     config_load "$1"
-    apps_sc "$2" "$3"
+    apps_sc "$2" "$3" "$4"
 else
     echo "ERROR:  [$1] is an invalid argument:"
     echo "usage:   ck8s apps <wc|sc>. "
--- a/bin/common.bash
+++ b/bin/common.bash
@@ -568,7 +568,7 @@ with_kubeconfig() {
     else
         log_info "Using unencrypted kubeconfig ${kubeconfig}"
         # shellcheck disable=SC2048
-        KUBECONFIG=${kubeconfig} ${*}
+        KUBECONFIG=${kubeconfig} "$@"
     fi
 }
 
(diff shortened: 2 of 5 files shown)


scripts/deploy-sc.sh
```diff
@@ -25,9 +25,9 @@ declare -a helmfile_opt_flags
 [[ -n "$INTERACTIVE" ]] && helmfile_opt_flags+=("$INTERACTIVE")
 
 if [ ${#} -eq 1 ] && [ "$1" = "sync" ]; then
-    helmfile -f . -e service_cluster sync
+    helmfile -f . -e service_cluster sync --concurrency 8
```
aarnq, line 28: Could you rework this into a flag so one can change this?

Elias-elastisys: Sure

scripts/deploy-wc.sh
```diff
@@ -50,9 +50,9 @@ declare -a helmfile_opt_flags
 [[ -n "$INTERACTIVE" ]] && helmfile_opt_flags+=("$INTERACTIVE")
```
OlleLarsson, line 50: We don't use this anymore it seems.
Can you remove it (and everything that goes with it)?
```suggestion
```

simonklb: `"${*}"` is used when you want to pass all arguments as a single string. However, `${*}` (without quotes) is generally a bad idea as that would both word-split and glob expand etc. I believe it's the same for `${@}`.

On the other hand `"${@}"` will pass the arguments as an array, so that means that `$1` and `$2` stays the same between functions even if `$1` is `a*` and $2 is `a b c`. This is probably what you want in most cases.

bin/apps.bash
```diff
@@ -24,15 +24,15 @@ apps_run_sc() {
     log_info "Applying applications in service cluster"
     (
         with_kubeconfig "${config[kube_config_sc]}" \
-            "${scripts_path}/deploy-sc.sh" "${1:-""}"
+            "${scripts_path}/deploy-sc.sh" "${1:-""}" "$2"
```
simonklb, line 27: This made me a bit suspicious. Since `$1` seems to be optional (because it has a fallback on `""`) I thought we shouldn't be able to assume that `$2` is set either. But it looks like the only uses of these functions are making sure to always set '$1`:
```console
$ fgrep -r 'apps_run_sc'
bin/apps.bash:apps_run_sc() {
bin/apps.bash:    apps_run_sc "${2:-""}" "$3"
$ fgrep -r 'apps_run_wc'
bin/apps.bash:apps_run_wc() {
bin/apps.bash:    apps_run_wc "${2:-""}" "$3"
```

Maybe just make it easy for ourselves and pass on all arguments?
```suggestion
            "${scripts_path}/deploy-sc.sh" "${@}"
```

bin/apps.bash
```diff
@@ -24,15 +24,15 @@ apps_run_sc() {
...
     )
 }
 
 apps_run_wc() {
     log_info "Applying applications in workload cluster"
     (
         with_kubeconfig "${config[kube_config_wc]}" \
-            "${scripts_path}/deploy-wc.sh" "${1:-""}"
+            "${scripts_path}/deploy-wc.sh" "${1:-""}" "$2"
```
simonklb, line 35: Same as above.

```suggestion
            "${scripts_path}/deploy-wc.sh" "${@}"
```

simonklb (approved):
LGTM!

We really should make an effort to get rid of the deploy scripts so that we can just use `helmfile` on its own.

simonklb (changes requested):
Hang on, before I approve. Is the default concurrency 8 or is this changing the existing behavior?

OlleLarsson:
> We really should make an effort to get rid of the deploy scripts so that we can just use `helmfile` on its own.

For sure.

> Hang on, before I approve. Is the default concurrency 8 or is this changing the existing behavior?

The default is no limit, so the concurrency is however many releases the apply/sync targets.

simonklb:
> > We really should make an effort to get rid of the deploy scripts so that we can just use `helmfile` on its own.
> 
> For sure.
> 
> > Hang on, before I approve. Is the default concurrency 8 or is this changing the existing behavior?
> 
> The default is no limit, so the concurrency is however many releases the apply/sync targets.

If `--concurrency=8` is a reasonable default I think this should be a `kind/admin-change` and it should be mentioned as a platform admin notice so that it's highlighted in the changelog when someone finds that their 32 cores aren't being fully utilized (probably not the case with unlimited either but still). :smile:

simonklb (approved):

bin/common.bash
```diff
@@ -568,7 +568,7 @@ with_kubeconfig() {
     else
         log_info "Using unencrypted kubeconfig ${kubeconfig}"
         # shellcheck disable=SC2048
-        KUBECONFIG=${kubeconfig} ${*}
+        KUBECONFIG=${kubeconfig} "$@"
```
aarnq, line 571: I feel a bit troubled that this has evaded our eyes for so long, good that it gets fixed. :+1:

aarnq (approved):

bin/common.bash
Zash: Isn't that what that [shellcheck directive](https://www.shellcheck.net/wiki/SC2048) disables?

PR #1886, opened 2023-11-27
apps: Added support for self-managed Kafka

AlbinB97:
> [!warning]

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

Adds support for self-managed Kafka using the [Strimzi Kafka Operator](https://github.com/strimzi/strimzi-kafka-operator).
...

- Part of elastisys/internal-issue-tracker#41

### DIFF
--- a/config/config/wc-config.yaml
+++ b/config/config/wc-config.yaml
@@ -67,6 +67,11 @@ user:
   # Requires that gatekeeper.allowUserCRDs.enabled is enabled.
   fluxv2:
     enabled: false
+  # Installs required cluster resources needed to install kafka-operator
+  # Requires that gatekeeper.allowUserCRDs.enabled is enabled.
+  kafka:
+    enabled: false
+
 falco:
   ## Falco alerting configuration.
   alerts:
--- a/helmfile/50-applications.yaml
+++ b/helmfile/50-applications.yaml
@@ -704,6 +704,7 @@ releases:
   - values/userCRDs/bitnami/sealedsecrets.yaml.gotmpl
   - values/userCRDs/mongodbcommunity/mongodb.yaml.gotmpl
   - values/userCRDs/flux/fluxv2.yaml.gotmpl
+  - values/userCRDs/strimzi/kafka.yaml.gotmpl
 
 # hierarchical namespace controller
 - name: hnc-controller
--- a/helmfile/values/user-rbac.yaml.gotmpl
+++ b/helmfile/values/user-rbac.yaml.gotmpl
@@ -42,6 +42,19 @@ userCRDs:
     - helmrepositories.source.toolkit.fluxcd.io
     - ocirepositories.source.toolkit.fluxcd.io
   {{- end }}
+  {{- if .Values.user.kafka.enabled }}
+    - kafkanodepools.kafka.strimzi.io
+    - kafkas.kafka.strimzi.io
+    - kafkaconnects.kafka.strimzi.io
+    - strimzipodsets.core.strimzi.io
+    - kafkatopics.kafka.strimzi.io
+    - kafkausers.kafka.strimzi.io
+    - kafkamirrormakers.kafka.strimzi.io
+    - kafkabridges.kafka.strimzi.io
+    - kafkaconnectors.kafka.strimzi.io
+    - kafkamirrormaker2s.kafka.strimzi.io
+    - kafkarebalances.kafka.strimzi.io
+  {{- end }}
   {{- range .Values.gatekeeper.allowUserCRDs.extraCRDs }}
   {{- toYaml .names | nindent 4 }}
   {{- end }}
--- a/helmfile/values/userCRDs/strimzi/kafka.yaml.gotmpl
+++ b/helmfile/values/userCRDs/strimzi/kafka.yaml.gotmpl
@@ -1,0 +1,182 @@
+{{- if .Values.user.kafka.enabled }}
+clusterRoles:
+  strimzi-cluster-operator-global:
+    rules:
+    - apiGroups:
+        - "storage.k8s.io"
+      resources:
+        - storageclasses
+      verbs:
+        - get
+    - apiGroups:
+        - ""
+      resources:
+        - nodes
+      verbs:
+        - list
+        - get
+    - apiGroups:
+        - ""
+      resources:
+        - endpoints
+      verbs:
+        - get
+        - list
+        - watch
+        - create
+        - delete
+        - patch
+        - update
+    - apiGroups:
+        - "apps"
+      resources:
+        - deployments/status
+      verbs:
+        - get
+        - list
+        - watch
+        - create
+        - delete
+        - patch
+        - update
+    - apiGroups:
+        - "build.openshift.io"
+      resources:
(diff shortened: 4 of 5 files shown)


AlbinB97:
> What has been tested here? Do we know if e.g. backups work? Does this come with some graphical administration endpoint? Help me understand what has been within scope and what has, perhaps, been left out. 😄

I have tested the basic functions of the operator with message queuing, as well as tested the different CRD functions that come with the operator, like bridge, connector, mirrormaker etc. I have also tested using both a single and multiple namespaces running both different and same kafka clusters. Also tested both producing and consuming messages in a different namespace. All this PR does is adding support so user can install the Strimzi Kafka Operator using the CRDs and provided roles that has been adjusted. Instructions for this will be available on elastisys.io later.

As for backups, there are a couple of different ways of doing it. Easiest way is just retaining the PV if the kafka-cluster would get deleted, and doing backups of the PV.

It does not come with any graphical administration endpoint.

Hope that helps clearing up a bit. If you want more information on how it works etc. we can book a meeting where I explain a bit what I've done and how it works with CK8S as a self-managed service.

@llarsson

cristiklein (approved):
I love this! Looking forward to the associated documentation page.

anders-elastisys (approved):

AlbinB97:
> This is more a general question regarding self-managed services, but should image repositories used for these services be added as allowed images in [gatekeeper](https://github.com/elastisys/compliantkubernetes-apps/blob/ed5a7edf901af4d6edb7ad495a9546dab75e8823/helmfile/values/gatekeeper/constraints.yaml.gotmpl#L19) already in the templating so that Platform Administrators don't have to manually add these, and to avoid Application Developers from getting warnings when deploying?

I think it could be a good idea. Maybe should be a separate issue because it would involve Flux as as well. It would be easy to delete the requirement for the image repository to be allowed in public docs later on.
I think the best solution would be making the image repositories be automatically allowed if either Kafka or Flux is enabled, but I am not sure how this solution would look at the moment.

cristiklein:
> This is more a general question regarding self-managed services, but should image repositories used for these services be added as allowed images in [gatekeeper](https://github.com/elastisys/compliantkubernetes-apps/blob/ed5a7edf901af4d6edb7ad495a9546dab75e8823/helmfile/values/gatekeeper/constraints.yaml.gotmpl#L19) already in the templating so that Platform Administrators don't have to manually add these, and to avoid Application Developers from getting warnings when deploying?

Good question! Indeed, we're not very consistent with this.

In some cases, we document how to push the right images into Harbor. (See [FerretDB](https://elastisys.io/compliantkubernetes/user-guide/self-managed-services/ferretdb/#pushing-ferretdb-image-to-harbor) and [JupyterHub](https://elastisys.io/compliantkubernetes/user-guide/self-managed-services/jupyterhub/#pushing-the-jupyeterhub-images-to-harbor).) In other cases, we document how to use the image locations advertised by the chart.

I see the following options:

1. Always include instructions to copy container images to Harbor.
  - Good, because consistent.
  - Good, because we preach more secure habits.
  - Bad, because it makes usage of self-managed services more cumbersome.
2. Always point to the default container images.
  - Good, because consistent.
  - Good, because it's quicker to get started for application developers.
  - Bad, because we preach less secure habits.
  - Bad, because the platform administrator needs to allowlist certain images.
3. Always point to the default container images *and* allowlist those images.
  - Good, because consistent.
  - Good, because it's quicker to get started for application developers.
  - Good, because we preach rather secure habits.
  - Bad, because it makes writing documentation a bit slower, especially for operators.
4. Choose 1-3 case-by-case.
  - Good, because it gives us flexibility.
  - Bad, because it makes things less consistent.

I'm leaning towards 4.

What do others think?

AlbinB97:
> > This is more a general question regarding self-managed services, but should image repositories used for these services be added as allowed images in [gatekeeper](https://github.com/elastisys/compliantkubernetes-apps/blob/ed5a7edf901af4d6edb7ad495a9546dab75e8823/helmfile/values/gatekeeper/constraints.yaml.gotmpl#L19) already in the templating so that Platform Administrators don't have to manually add these, and to avoid Application Developers from getting warnings when deploying?
> 
> Good question! Indeed, we're not very consistent with this.
> 
> In some cases, we document how to push the right images into Harbor. (See [FerretDB](https://elastisys.io/compliantkubernetes/user-guide/self-managed-services/ferretdb/#pushing-ferretdb-image-to-harbor) and [JupyterHub](https://elastisys.io/compliantkubernetes/user-guide/self-managed-services/jupyterhub/#pushing-the-jupyeterhub-images-to-harbor).) In other cases, we document how to use the image locations advertised by the chart.
> 
> I see the following options:
> 
>     1. Always include instructions to copy container images to Harbor.
> 
> 
>     * Good, because consistent.
> 
>     * Good, because we preach more secure habits.
> 
>     * Bad, because it makes usage of self-managed services more cumbersome.
> 
> 
>     2. Always point to the default container images.
> 
> 
>     * Good, because consistent.
> 
>     * Good, because it's quicker to get started for application developers.
> 
>     * Bad, because we preach less secure habits.
> 
>     * Bad, because the platform administrator needs to allowlist certain images.
> 
> 
>     3. Always point to the default container images _and_ allowlist those images.
> 
> 
>     * Good, because consistent.
> 
>     * Good, because it's quicker to get started for application developers.
> 
>     * Good, because we preach rather secure habits.
> 
>     * Bad, because it makes writing documentation a bit slower, especially for operators.
> 
> 
>     4. Choose 1-3 case-by-case.
> 
> 
>     * Good, because it gives us flexibility.
> 
>     * Bad, because it makes things less consistent.
> 
> 
> I'm leaning towards 4.
> 
> What do others think?

I am also leaning towards 4.
Since sometimes it's not necessary to allowlist an image when it is just one image that could easily be copied to harbor, and that should be the preferred method in those cases. In other cases when it involves operators where there are a lot of different images, it would be way easier all around to allowlist the images. But like I mentioned, it could be nice to have some automatic way to allowlist the images by just having the feature enabled, instead of having a platform administrator manually do it.

llarsson:
I think that since we anyways will get a ticket that says "please open up stuff so I can install and self-manage Kafka", because AFAIUI, we don't enable the custom CRD stuff by default, we could just have a script that then also allowlists the appropriate registries. Am I correct in my understanding here?

cristiklein:
> I think that since we anyways will get a ticket that says "please open up stuff so I can install and self-manage Kafka", because AFAIUI, we don't enable the custom CRD stuff by default, we could just have a script that then also allowlists the appropriate registries. Am I correct in my understanding here?

I was operating under the same assumption.

AlbinB97:
> I think that since we anyways will get a ticket that says "please open up stuff so I can install and self-manage Kafka", because AFAIUI, we don't enable the custom CRD stuff by default, we could just have a script that then also allowlists the appropriate registries. Am I correct in my understanding here?

Yes that is correct.
Having a script to do it would be the easiest solution.

PR #1951, opened 2024-01-11
Added alert for failing webhooks

### LINKED ISSUE #1870: [2] Alert for failing webhook
In Compliant Kubernetes we have both validating- and mutating webhooks with a failurePolicy of type fail.

This means that for most state altering operations, having a webhook be unavailable for whatever reason, can be detrimental for cluster performance and function.

We should figure out a way to monitor webhook calls and create some alert that can notify a platform administrator.

The Kubernetes API server does expose some metrics with regards to webhooks, see [here](https://github.com/kubernetes/apiserver/blob/master/pkg/admission/metrics/metrics.go#L116-L126).

We should be able to use the above generated metrics to notice any degradation or complete failure in calling important webhooks.

**Definition of done:**
There are sufficient prometheus AlertRules added to the platform that will notify platform administrators that webhooks calls are failing.

lunkan93:
> [!warning]

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

Added an alert for failing webhooks

- Fixes #1870 

#### Additional information to reviewers
Not sure if we would want to have it enabled by default, as of now I have disabled it by default.

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:
- Bug checks:

### DIFF
--- a/helmfile.d/charts/prometheus-alerts/files/webhooks.yaml
+++ b/helmfile.d/charts/prometheus-alerts/files/webhooks.yaml
@@ -1,0 +1,10 @@
+- name: webhook-failures
+  rules:
+  - alert: WebhookFailing
+    annotations:
+      description: Webhooks have been failing during the last 10m
+      summary: Webhooks requests are failing
+    expr: rate(apiserver_admission_webhook_rejection_count{error_type!="no_error"}[5m]) > 0
+    for: 10m
+    labels:
+      severity: warning
--- a/config/config/common-config.yaml
+++ b/config/config/common-config.yaml
@@ -378,6 +378,8 @@ prometheus:
   tolerations: []
   affinity: {}
   nodeSelector: {}
+  webhookAlerts:
+    enabled: false
   ## Predictive alert if the resource usage will hit the set percentage in 3 days
   capacityManagementAlerts:
     enabled: true
--- a/helmfile.d/charts/prometheus-alerts/templates/alerts/webhooks.yaml
+++ b/helmfile.d/charts/prometheus-alerts/templates/alerts/webhooks.yaml
@@ -1,0 +1,19 @@
+{{- if and .Values.defaultRules.create .Values.defaultRules.rules.webhooks }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ printf "%s-%s" (include "prometheus-alerts.fullname" .) "webhooks" | trunc 63 | trimSuffix "-" }}
+  labels:
+    app: {{ template "prometheus-alerts.name" . }}
+    {{- include "prometheus-alerts.labels" . | nindent 4 }}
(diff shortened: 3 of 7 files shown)


raviranjanelastisys (approved):
LGTM !!

aarnq (approved):

OlleLarsson (approved):
Could you provide a simple test to show that it works as expected 🤗 ?

helmfile.d/charts/prometheus-alerts/files/webhooks.yaml
```diff
@@ -0,0 +1,10 @@
+- name: webhook-failures
+  rules:
+  - alert: WebhookFailing
+    annotations:
+      description: Webhooks have been failing during the last 10m
+      summary: Webhooks requests are failing
+    expr: rate(apiserver_admission_webhook_rejection_count{error_type!="no_error"}[5m]) > 0
```
OlleLarsson, line 7: Does the count for `apiserver_admission_webhook_rejection_count` increase if, let's say, gatekeper restarts and one of the pods happens to handle a request (not sure if that's a valid scenario or not)?

lunkan93: I believe this should increase the counter (simply scaling down to 0 replicas of the gatekeeper controller seems to always increase the counter). This alone is not enough to trigger the alert from my testing, as it shouldn't keep the `rate` of the counter above 0 for a long enough period alone.

The following screenshot is from my testing after scaling down to 0 replicas and just trying to create a deployment continuously over a 10 minute period, which fails because gatekeeper is unavailable. This triggers the alert.

The second spike is me just trying to create a deployment a single time, then scaling up to 3 replicas again, which does not trigger the alert.

From my testing, the `error_type!="no_error"` (could also be `error_type="calling_webhook_error"`) should filter out every request that gets rejected due to violating a policy, i.e. trying to create a deployment using the latest tag, and only take into account request that actually fail.

Screenshot from 2024-01-12 07-33-21

lunkan93: @OlleLarsson Any thoughts on this? We could lower the interval for the alert to trigger but my fear is that it might get too spammy

OlleLarsson: We can adjust interval if needed further down the road.

OlleLarsson (approved):

Issue #1669, opened 2023-07-07
Investigate enabling falcoctl for new Falco Helm chart


anders-elastisys:
**What should be investigated.**
The Falco Helm chart uses [falcoctl](https://github.com/falcosecurity/charts/tree/falco-3.3.0/falco#about-falco-artifacts) starting from v3.0.0. `falcoctl` is  a tool used for continuously updating Falco plugins and rules, whereas before you had to wait on new releases of the Helm chart to get these updates (or manually add custom rules). [This PR](https://github.com/elastisys/compliantkubernetes-apps/pull/1662) looks to upgrades the Falco Helm chart to v3.3.0, but it is configured to always have `falcoctl` disabled as to not break the old values configurations according to the instructions given [here](https://github.com/falcosecurity/charts/blob/falco-3.3.0/falco/BREAKING-CHANGES.md). 

Investigate whether or not `falcoctl` should be enabled in CK8s and what permissions are required to be able to run it.

**What artifacts should this produce.**
Pros and cons for enabling `falcoctl` in CK8s.

anders-elastisys:
**Follow up:** The `artifacts` part was enabled in ck8s-apps in [this PR](https://github.com/elastisys/compliantkubernetes-apps/pull/1843), however, the [`follow` sidecar container](https://github.com/elastisys/compliantkubernetes-apps/blob/v0.34.0/helmfile/upstream/falcosecurity/falco/values.yaml#L371-L377) for automatically checking for updates of `artifacts` should still be investigated.

PR #1981, opened 2024-01-29
Add templating for air-gapped environments

anders-elastisys:
> [!warning]
> - [ ] personal data beyond what is necessary for interacting with this pull request
> - [ ] business confidential information, such as customer names

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

There is now an additional `CK8S_FLAVOR` called `air-gapped` with similar settings to `prod` but with some additional configuration to avoid any outgoing traffic to the Internet.

This PR introduces additional templating to be able to configure CK8s in an air-gapped environment. Most of the config options provides ways of overriding services that normally downloads plugins from the Internet to instead pull from a URL of a server or registry available inside the air-gapped environment.

Also added so that `http_2xx` blackbox monitors work even when `global.verifyTls=false`.

- Fixes #

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:
- Bug checks:

### DIFF
--- a/config/config/common-config.yaml
+++ b/config/config/common-config.yaml
@@ -118,6 +118,11 @@ falco:
       enabled: false
       version: 2.0.0
 
+  artifact:
+    install:
+      # set to false in an air-gapped environment, unless artifacts are self-hosted
+      enabled: true
+
   # Setting tty to "true" will immediately display Falco logs by flushing them as they are emitted
   tty: true
 
@@ -133,6 +138,14 @@ falco:
       # -- Needed to enable eBPF JIT at runtime for performance reasons.
       # Can be skipped if eBPF JIT is enabled from outside the container
       hostNetwork: false
+
+    module:
+      # override the URL used for downloading driver modules, e.g. to use a self hosted file server in an air-gapped environment
+      repoURL: ""
+
+  # use custom indexes for falcoctl downloads
+  # ref: https://github.com/falcosecurity/charts/blob/falco-3.8.3/charts/falco/values.yaml#L390-L395
+  customIndexes: []
 
   ## additional falco rules
   ## ref: https://falco.org/docs/rules/
@@ -853,6 +866,29 @@ trivy:
     interval: 5m
   tolerations: []
   affinity: {}
+
+  # configurations for an offline / air-gapped environment
+  # ref: https://github.com/aquasecurity/trivy-operator/tree/main/deploy/helm#values
+  scanner:
+    offlineScanEnabled: false
+    dbRegistry: ""
+    dbRepository: ""
+    dbRepositoryInsecure: false
+    javaDbRegistry: ""
+    javaDbRepository: ""
+    # if authorization is required for pulling from registry, create a pull
+    # secret in the monitoring namespace and configure the secret name
+    imagePullSecret:
+      name: ""
+    registry:
+      mirror: {}
+        # "docker.io": registry.example.org:5000
+        # "gcr.io": registry.example.org:5000
+        # "ghcr.io": registry.example.org:5000
+        # "index.docker.io": registry.example.org:5000
+        # "quay.io": registry.example.org:5000
+        # "registry.k8s.io": registry.example.org:5000
+
 kured:
   enabled: false
   # See options at https://github.com/weaveworks/kured/blob/1.9.1/charts/kured/values.yaml#L24
--- a/config/config/sc-config.yaml
+++ b/config/config/sc-config.yaml
@@ -718,6 +718,15 @@ opensearch:
   defaultTemplates: true
   additionalTemplates: {}
 
+  plugins:
+    # in an air-gapped environment where the nodes are not connected to the Internet, set
+    # following variable to false to prevent downloading external object storage plugin
+    installExternalObjectStoragePlugin: true
+
+    # in an air-gapped environment this can be used to install plugins from known sources
+    additionalPlugins: []
+    # - server.local:8080/repository-s3-2.8.0.zip
+
   # Index state management
   ism:
     rolloverSizeGB: 1
@@ -1034,6 +1043,8 @@ alerts:
     #   {{ end }}
   opsGenie:
     apiUrl: https://api.eu.opsgenie.com
+  # Configure custom alert receivers
+  customReceivers: []
 
 externalTrafficPolicy:
   # Whitelisting requires externalTrafficPolicy.local to be true
--- a/helmfile.d/values/prometheus-blackbox-exporter-sc.yaml.gotmpl
+++ b/helmfile.d/values/prometheus-blackbox-exporter-sc.yaml.gotmpl
@@ -4,6 +4,10 @@ pspEnabled: false
 
 config:
   modules:
+    http_2xx:
+      http:
+        tls_config:
+          insecure_skip_verify: {{ not .Values.global.verifyTls }}
     http_400:
       prober: http
       timeout: 5s
--- a/helmfile.d/values/prometheus-blackbox-exporter-wc.yaml.gotmpl
+++ b/helmfile.d/values/prometheus-blackbox-exporter-wc.yaml.gotmpl
@@ -4,6 +4,10 @@ pspEnabled: false
 
 config:
   modules:
+    http_2xx:
+      http:
+        tls_config:
+          insecure_skip_verify: {{ not .Values.global.verifyTls }}
     http_400:
       prober: http
       timeout: 5s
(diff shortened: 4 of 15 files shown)


helmfile.d/values/prometheus-blackbox-exporter-sc.yaml.gotmpl
```diff
@@ -4,6 +4,10 @@ pspEnabled: false
 
 config:
   modules:
+    http_2xx:
+      http:
+        tls_config:
+          insecure_skip_verify: {{ not .Values.global.verifyTls }}
```
salehsedghpour, line 10: Should we (is it possible) add a comment here to discuss that it's not recommended, but might be required in some Air-gapped environments.

helmfile.d/values/prometheus-blackbox-exporter-wc.yaml.gotmpl
```diff
@@ -4,6 +4,10 @@ pspEnabled: false
 
 config:
   modules:
+    http_2xx:
+      http:
+        tls_config:
+          insecure_skip_verify: {{ not .Values.global.verifyTls }}
```
salehsedghpour, line 10: Should we (is it possible) add a comment here to discuss that it's not recommended, but might be required in some Air-gapped environments.

salehsedghpour (changes requested):
Thanks @anders-elastisys! I was wondering if we should also discuss this feature in Readme files in this repository.

cristiklein (changes requested):
I love how you give detailed explanation on why each configuration is needed for airgapped.

However, I think this might be better to have as a flavor, alongside alongside `dev` and `prod`. I'm thinking the default values would be unchanged from what we have today, whereas the airgapped flavor would set all the correct values for an airgapped environment.

What do you think?

helmfile.d/values/prometheus-blackbox-exporter-sc.yaml.gotmpl
anders-elastisys: As this is a global variable that is used to change configurations of other services as well I don't see a need to have a comment for it (not here at least), as this should always be set to true in production environments. I added this so that you don't get alerts when `verifyTls=false`, as in this case you have made a conscious choice to not care about verifying TLS.

anders-elastisys:
@cristiklein yes I was considering something like that as well, I could look into adding a `offline` or `airgapped` flavor perhaps which atleast sets the override variables to `set-me` so that it is more clear what you need to configure to make it work in an air-gapped environment.

cristiklein:
> @cristiklein yes I was considering something like that as well, I could look into adding a `offline` or `airgapped` flavor perhaps which atleast sets the override variables to `set-me` so that it is more clear what you need to configure to make it work in an air-gapped environment.

Thanks @anders-elastisys . Please go with `air-gapped`, since that is what people look for.

anders-elastisys:
@salehsedghpour regarding adding a readme in this repo, I am working on public documentation that will be put in the compliantkubernetes repo, I am not sure what the consensus is on having documentation put in this repository. If I add a new flavor as suggested by @cristiklein I can update the main README of this repo to cover using this flavor by setting `CK8S_FLAVOR`.

anders-elastisys:
@cristiklein PTAL, I added an `air-gapped` flavor and I also added templating to add custom receivers for alertmanager. Right now the `air-gapped` falvor is the same as `prod` flavor but with additional config options. Maybe there is a better way to do it, such as merging the air-gapped config with prod instead of having the same definitions replicated in two places.

cristiklein (approved):
Love it, thanks!

cristiklein:
@anders-elastisys Just to make sure, have you tried something like `helmfile template` to check that this PR is essentially a no-op for the `dev` and `prod` flavours?

config/config/common-config.yaml
```diff
@@ -134,6 +139,13 @@ falco:
       # Can be skipped if eBPF JIT is enabled from outside the container
       hostNetwork: false
 
+    # override the URL used for downloading driver modules, e.g. to use a self hosted file server in an air-gapped environment
+    repoURL: ""
```
aarnq, line 143: Is this only used for the module driver? In that case I'd prefer if it was under the key `module`.

config/config/common-config.yaml
```diff
@@ -118,6 +118,11 @@ falco:
       enabled: false
       version: 2.0.0
 
+  artifact:
+    install:
+      # set to false in an air-gapped environment, unless artifacts are self-hosted
+      enabled: true
```
aarnq, line 124: Do we really want this on by default even for regular flavours?
What benefit does it bring compared to the rules bundled with the current image, is there a current version skew?

config/config/sc-config.yaml
```diff
@@ -718,6 +718,13 @@ opensearch:
   defaultTemplates: true
   additionalTemplates: {}
 
+  plugins:
+    # in an air-gapped environment where the nodes are not connected to the Internet , the following
+    # variable can be used to override where the objectStorage plugin should be downloaded from.
+    objectStorageUrlOverride: # e.g. server.local:8080/repository-s3-2.8.0.zip
+
+    additionalPlugins: []
+
```
aarnq, line 727:
```suggestion
  plugins:
    # in an air-gapped environment where the nodes are not connected to the Internet, 
    # this will prevent external/default object storage plugins from being installed
    installExternalObjectStoragePlugin: false
    # or installDefaultObjectStoragePlugin: false
    # or installObjectStoragePlugin: false

    # in an air-gapped environment this can be used to install plugins from known sources
    additionalPlugins: []
      # - server.local:8080/repository-s3-2.8.0.zip
```

Can we rework this a bit and then just reuse the `additionalPlugins`?

config/config/common-config.yaml
```diff
@@ -853,6 +865,29 @@ trivy:
     interval: 5m
   tolerations: []
   affinity: {}
+
+  # configurations for an offline / air-gapped environment
+  # ref: https://github.com/aquasecurity/trivy/tree/main/helm/trivy#parameters
```
aarnq, line 870:
```suggestion
  # ref: https://github.com/aquasecurity/trivy-operator/tree/main/deploy/helm#values
```

config/config/common-config.yaml
```diff
@@ -853,6 +865,29 @@ trivy:
...
+  offline:
+    offlineScan: false
```
aarnq, line 872:
```suggestion
  offlineScan:
    enabled: false
```
Or is there other conditions when we would like to edit configuration under `offline` that do not require `offlineScan: true`? Else we should rename the section, and rename the enable flag to `offlineScanEnabled` or similar.

aarnq:
Also, do add the new flavour under `tests/` where applicable. :slightly_smiling_face:

anders-elastisys:
@cristiklein  Templating seems to work fine when initializing a new environment using the new `air-gapped` flavor, however, I noticed that the validation done by the scripts in apps did not work as I expected as it only warns if there are `set-me`'s in the override config if they are also set to `set-me` in the default/template config as well. Thus, with how I have configured the variables now you won't get a warning that for example `global.issuer` is not set, which is what I expected. Do we want to keep it this way or do we want to warn for everything that is set to `set-me` for the current ck8s environment config?

config/config/common-config.yaml
anders-elastisys: I am not sure if there is any real benefit for having it enabled, I just kept it enabled from previous config. I can maybe try and look into if there is any benefit in keeping it enabled

salehsedghpour (approved):

config/config/common-config.yaml
aarnq: Maybe we should call the `offline` key `scanner` instead then, as this mainly applies to the scanner jobs, and the have it as `offlineScanEnabled` to match other config keys.

config/config/common-config.yaml
anders-elastisys: Currently we seem to get falco rules for a specific version [since this PR](https://github.com/elastisys/compliantkubernetes-apps/pull/1843/files) when artifact `install` is enabled (I noticed [v3.0.0](https://github.com/falcosecurity/rules/releases/tag/falco-rules-3.0.0) of falco rules dropped recently).  However, I feel like we can keep it like it is now and maybe look deeper into this in a separate issue, I created [this issue](https://github.com/elastisys/compliantkubernetes-apps/issues/1669) long ago before `artifact.install` was enabled in apps, so maybe it can be investigated as part of that issue instead.

aarnq (approved):
LGTM

PR #1996, opened 2024-02-14
add azure storage resources management script

vomba:
> [!warning]

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

Adds script to create storage resources on Azure (resource group, storage account and storage containers) for ck8s-apps.

- Fixes https://github.com/elastisys/mse-internal-docs/issues/189

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Network Policy checks:
- Pod Security Policy checks:
- Falco checks:
- Audit checks:
- Bug checks:


scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,111 @@
...
+esac
+shift
+
+function create_resource_group() {
+
+    echo "checking if resource group exists" >&2
+    GROUP_EXISTS=$(az group list --query '[].name' | awk "/${CK8S_ENVIRONMENT_NAME}-storage-resource-group/")
+    if [ "$GROUP_EXISTS" ]; then
+        echo "resource group [${CK8S_ENVIRONMENT_NAME}-storage-resource-group] already exists" >&2
+        echo "contnue using this group ? (y/n)" >&2
```
Xartos, line 30:
```suggestion
        echo "continue using this group ? (y/n)" >&2
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,111 @@
...
+        # shellcheck disable=SC2162
+        read -n 1 cmdinput
+        case "$cmdinput" in
+        y) return ;;
+        n) exit 0 ;;
```
Xartos, line 35: Maybe default to exit
```suggestion
        y|Y) return ;;
        *) exit 0 ;;
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,111 @@
...
+
+function create_resource_group() {
+
+    echo "checking if resource group exists" >&2
+    GROUP_EXISTS=$(az group list --query '[].name' | awk "/${CK8S_ENVIRONMENT_NAME}-storage-resource-group/")
+    if [ "$GROUP_EXISTS" ]; then
+        echo "resource group [${CK8S_ENVIRONMENT_NAME}-storage-resource-group] already exists" >&2
+        echo "contnue using this group ? (y/n)" >&2
+        # shellcheck disable=SC2162
+        read -n 1 cmdinput
```
Xartos, line 32: Why not adhere to the check?

```suggestion
        read -r -n 1 cmdinput
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,111 @@
...
+}
+
+if [[ "$ACTION" == "$CREATE_ACTION" ]]; then
+    echo "Creating Resource Group" >&2
+    create_resource_group
+
+    echo "Creating Storage Account" >&2
+    create_storage_account
+
+    CONTAINERS_LIST=$(az storage container list --account-name "$CK8S_ENVIRONMENT_NAME"storageaccount --query '[].name' --only-show-errors)
```
Xartos, line 102: Should this be moved into the function so it's not as much of a magic variable

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,111 @@
...
+        read -n 1 cmdinput
+        case "$cmdinput" in
+        y) return ;;
+        n) exit 0 ;;
+        esac
+    else
+        echo "resource group [${CK8S_ENVIRONMENT_NAME}-storage-resource-group] does not exist, creating it now" >&2
+        az group create \
+            --name "$CK8S_ENVIRONMENT_NAME"-storage-resource-group \
+            --location swedencentral --only-show-errors
```
Xartos, line 41: Do we think that we'll use this in some other location than sweden for the forseable future? If so, then maybe make this configurable

```suggestion
            --location "${AZURE_LOCATION}" --only-show-errors
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,111 @@
...
+    fi
+
+}
+
+function create_storage_account() {
+
+    echo "checking if storage account exists" >&2
+    ACCOUNT_EXISTS=$(az storage account list --query '[].name' | awk "/${CK8S_ENVIRONMENT_NAME}storageaccount/")
+    if [ "$ACCOUNT_EXISTS" ]; then
+        echo "storage account [${CK8S_ENVIRONMENT_NAME}storageaccount] already exists" >&2
```
aarnq, line 51: Is there a reason as to why the storage account isn't `${CK8S_ENVIRONMENT_NAME}-storage-account`?
Also do we need the `-<resource-type>` suffixes, isn't it clear from their name what it relates to?
Or do we in the case of the `resource-group` have multiple for a typical Compliant Kubernetes setup?

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+
+set -e
```
aarnq, line 3:
```suggestion
#!/usr/bin/env bash

set -euo pipefail
```

aarnq (changes requested):
I think we should consult the config for any and all values that we need instead of assuming that they adhere to a pattern.

scripts/azure/storage-manager.sh
vomba: Storage accounts names can only contain lowercase letter and numbers. So even `${CK8S_ENVIRONMENT_NAME}` needs to be in that format.
I just added the suffixes for extra "visibility" it can be removed.

scripts/azure/storage-manager.sh
vomba: I believe we will, I'll change that

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,138 @@
...
+}
+
+if [[ "$ACTION" == "$CREATE_ACTION" ]]; then
+    echo "Creating Resource Group" >&2
+    create_resource_group
+
+    echo "Creating Storage Account" >&2
+    create_storage_account
+
+    echo "Creating Storage Containers"
```
Xartos, line 130: I assume this should have 
```suggestion
    echo "Creating Storage Containers" >&2
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,138 @@
...
+        read -r -n 1 cmdinput
+        case "$cmdinput" in
+        y|Y) return ;;
+        *) exit 0 ;;
+        esac
+    else
+        echo "resource group [${CK8S_ENVIRONMENT_NAME}-storage-resource-group] does not exist, creating it now" >&2
+        az group create \
+            --name "$CK8S_ENVIRONMENT_NAME"-storage-resource-group \
+            --location "${AZURE_LOCATION}" --only-show-errors
```
simonklb, line 69: I don't think we want to create a new resource group for this right? Shouldn't the object storage resources live in the same resource group as the rest of the environment?

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,138 @@
...
+    fi
+
+}
+
+function create_storage_account() {
+
+    echo "checking if storage account exists" >&2
+    ACCOUNT_EXISTS=$(az storage account list --query '[].name' | awk "/${CK8S_ENVIRONMENT_NAME}storageaccount/")
```
simonklb, line 77: There is a command specifically for this which you probably should use instead:
```shell
$ out=$(az storage account check-name --name not-valid)
$ echo $out | jq -r .nameAvailable
false
$ echo $out | jq -r .reason
AccountNameInvalid

$ out=$(az storage account check-name --name hanistorageaccount)
$ echo $out | jq -r .nameAvailable
false
$ echo $out | jq -r .reason
AlreadyExists

$ out=$(az storage account check-name --name simonstorageaccount)
$ echo $out | jq -r .nameAvailable
true
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,138 @@
...
+function create_containers() {
+
+    CONTAINERS_LIST=$(az storage container list --account-name "$CK8S_ENVIRONMENT_NAME"storageaccount --query '[].name' --only-show-errors)
+
+    # shellcheck disable=SC2068
+    for container in ${CONTAINERS[@]}; do
+
+        echo "checking status of container ${container}]" >&2
+
+        CONTAINER_EXISTS=$(echo "$CONTAINERS_LIST" | awk "/${container}/")
```
simonklb, line 106: While this probably won't become a problem this could give a false positive, for example a container named `foobar` would make this script think a container named only `foo` exists.

For example:
```shell
$ CONTAINERS_LIST=$(az storage container list --account-name hanistorageaccount --query '[].name' --only-show-errors)
$ echo "$CONTAINERS_LIST" | awk "/hani-audit/"
  "hani-audit",
$ echo "$CONTAINERS_LIST" | awk "/hani-a/"
  "hani-audit",
```

Since the command outputs json, what you could do is to use `jq`, something like:
```shell
$ echo $CONTAINERS_LIST | jq '. | index("hani-sclogs")'
3
$ echo $CONTAINERS_LIST | jq '. | index("doesnotexist")'
null
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,138 @@
...
+    create_resource_group
+
+    echo "Creating Storage Account" >&2
+    create_storage_account
+
+    echo "Creating Storage Containers"
+    create_containers "${CONTAINERS}"
+elif [[ "$ACTION" == "$DELETE_ACTION" ]]; then
+    echo "deleting..." >&2
+    delete_all
```
simonklb, line 134: Since this script is in charge of creating the storage account, should it not also delete it after deleting all the containers?

simonklb (changes requested):

scripts/azure/storage-manager.sh
vomba: Yeah, that makes sense.

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,138 @@
...
+
+[ "$objectstorage_type_sc" = "azure" ] && buckets_sc=$(echo "${sc_config}" | yq4 '.objectStorage.buckets.*' -)
+[ "$objectstorage_type_wc" = "azure" ] && buckets_wc=$(echo "${wc_config}" | yq4 '.objectStorage.buckets.*' -)
+
+CONTAINERS=$( { echo "$buckets_sc"; echo "$buckets_wc"; } | sort | uniq | tr '\n' ' ' | sed s'/.$//')
+
+function create_resource_group() {
+
+    echo "checking if resource group exists" >&2
+    GROUP_EXISTS=$(az group list --query '[].name' | awk "/${CK8S_ENVIRONMENT_NAME}-storage-resource-group/")
```
simonklb, line 56: If there isn't any better command for checking if a resource group exists, at least use `jq` instead of `awk`.

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+readonly CREATE_ACTION="create"
+readonly DELETE_ACTION="delete"
+
+log_info() {
```
robinAwallace, line 8: Did you consider using the log functions in `bin/common.sh`? Thar would make further alternations to logs be more uniform :slightly_smiling_face:

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,157 @@
...
+fi
+
+[ "$objectstorage_type_sc" = "azure" ] && buckets_sc=$(echo "${sc_config}" | yq4 '.objectStorage.buckets.*' -)
+[ "$objectstorage_type_wc" = "azure" ] && buckets_wc=$(echo "${wc_config}" | yq4 '.objectStorage.buckets.*' -)
+
+CONTAINERS=$( { echo "$buckets_sc"; echo "$buckets_wc"; } | sort | uniq | tr '\n' ' ' | sed s'/.$//')
+
+log_info "Operating on containers: ${CONTAINERS// /', '}"
+
+function usage() {
```
robinAwallace, line 46: I can not see where usage is referenced. It is in a higher up script?

vomba: It's in the last 3 lines of the script.

scripts/azure/storage-manager.sh
vomba: I will check it out, i basically used the s3 logging scripts.

raviranjanelastisys (approved):
LGTM !!

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,157 @@
...
+        log_info "continue using this group ? (y/n)" >&2
+        read -r -n 1 cmdinput
+        case "$cmdinput" in
+        y|Y) return ;;
+        *) exit 0 ;;
+        esac
+    else
+        log_info "resource group [${CK8S_ENVIRONMENT_NAME}-storage-resource-group] does not exist, creating it now" >&2
+        az group create \
+            --name "$CK8S_ENVIRONMENT_NAME"-storage-resource-group \
```
simonklb, line 77: Could skip the `-resource-group` suffix in the resource group name.

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,157 @@
...
+            --location "${AZURE_LOCATION}" --only-show-errors
+    fi
+}
+
+function create_storage_account() {
+
+    log_info "checking storage account availability" >&2
+    out=$(az storage account check-name --only-show-errors --name "${CK8S_ENVIRONMENT_NAME}"storageaccount)
```
simonklb, line 85: Could skip the `storageaccount` suffix on the storage account name.

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,157 @@
...
+            log_info "container ${container} does not exist, creating it now" >&2
+            az storage container create \
+                -n "$container" \
+                --account-name "$CK8S_ENVIRONMENT_NAME"storageaccount --only-show-errors
+        fi
+    done
+}
+
+function delete_all() {
+    az group delete --name "$CK8S_ENVIRONMENT_NAME"-storage-resource-group
```
simonklb, line 139: Consider making the resource group name a global variable since it's used throughout the script and if you update it at one place it could be easy to updated it in all places if you're not careful!

simonklb (approved):
Nice!

Just added some non-blockers, you do with them as you please! :)

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,157 @@
...
+
+log_info "Operating on containers: ${CONTAINERS// /', '}"
+
+function usage() {
+    echo "Usage:" 1>&2
+    echo " $0 create" 1>&2
+    echo " $0 delete" 1>&2
+}
+
+case "$1" in
```
simonklb, line 52:
```suggestion
if [ "${#}" -lt 1 ]; then
  usage
  exit 1
fi

case "$1" in
```

simonklb (changes requested):
Found this while testing it.

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
```
simonklb, line 4:
```suggestion
set -euo pipefail

: "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
: "${CK8S_ENVIRONMENT_NAME:?Missing CK8S_ENVIRONMENT_NAME}"
: "${AZURE_LOCATION:?Missing AZURE_LOCATION}"
```

simonklb:
We probably need to either have some documentation on how to get the storage account key or this script could patch the secrets configuration

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,157 @@
...
+delete)
+    ACTION=$DELETE_ACTION
+    ;;
+esac
+shift
+
+function create_resource_group() {
+
+    log_info "checking if resource group exists" >&2
+    GROUP_EXISTS=$(az group list --query '[].name' | jq --arg group "${CK8S_ENVIRONMENT_NAME}"-storage-resource-group '. | index($group)')
```
aarnq, line 65: The resource group should be read from the config or secrets, and set during init.

scripts/azure/storage-manager.sh
aarnq, line 85: The storage account should be read from the config or secrets, and set during init.

aarnq (changes requested):

scripts/azure/storage-manager.sh
simonklb: That would solve this as well: https://github.com/elastisys/compliantkubernetes-apps/pull/1999#issuecomment-2009931660

vomba:
> We probably need to either have some documentation on how to get the storage account key or this script could patch the secrets configuration

Will look into it.

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,168 @@
...
+
+if [ "$objectstorage_type_sc" != "azure" ] && [ "$objectstorage_type_wc" != "azure" ]; then
+    log_error "Azure Storage is not enabled in either cluster, aborting!"
+    exit 1
+fi
+
+[ "$objectstorage_type_sc" = "azure" ] && buckets_sc=$(echo "${sc_config}" | yq4 '.objectStorage.buckets.*' -)
+[ "$objectstorage_type_wc" = "azure" ] && buckets_wc=$(echo "${wc_config}" | yq4 '.objectStorage.buckets.*' -)
+
+RESOURCE_GROUP=$(echo "${common_config}" | yq4 '.azure.resourceGroup')
```
simonklb, line 47: I think this should be scoped under `objectStorage`, something like `objectStorage.azure.resourceGroup` to align with the secrets `objectStorage.azure.storageAccount(Name|Key)`?

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,168 @@
...
+        az group create \
+            --name "${RESOURCE_GROUP}" \
+            --location "${AZURE_LOCATION}" --only-show-errors
+    fi
+}
+
+function create_storage_account() {
+
+    log_info "checking storage account availability" >&2
+    out=$(az storage account check-name --only-show-errors --name "${CK8S_ENVIRONMENT_NAME}")
```
aarnq, line 96: Isn't the storage account name supposed to be what we set under `objectStorage.azure.storageAccountName`?

scripts/azure/storage-manager.sh
vomba: Yeah will change that.

scripts/azure/storage-manager.sh
vomba: I don't remember discussing that, but can be changed to that.

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+: "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
+: "${CK8S_ENVIRONMENT_NAME:?Missing CK8S_ENVIRONMENT_NAME}"
```
aarnq, line 6: Move this down and set:
```suggestion
CK8S_ENVIRONMENT_NAME="$(yq4 .global.ck8sEnvironmentName <<< "${common_config}")"
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,168 @@
...
+if [ "$objectstorage_type_sc" != "azure" ] && [ "$objectstorage_type_wc" != "azure" ]; then
+    log_error "Azure Storage is not enabled in either cluster, aborting!"
+    exit 1
+fi
+
+[ "$objectstorage_type_sc" = "azure" ] && buckets_sc=$(echo "${sc_config}" | yq4 '.objectStorage.buckets.*' -)
+[ "$objectstorage_type_wc" = "azure" ] && buckets_wc=$(echo "${wc_config}" | yq4 '.objectStorage.buckets.*' -)
+
+RESOURCE_GROUP=$(echo "${common_config}" | yq4 '.objectStorage.azure.resourceGroup')
+CONTAINERS=$( { echo "$buckets_sc"; echo "$buckets_wc"; } | sort | uniq | tr '\n' ' ' | sed s'/.$//')
```
aarnq, line 48: This will fail if Azure isn't enabled on one of the clusters as either variable will be unset. You can use `${varname:-}` to expand them to an empty string if they are unset.

scripts/azure/storage-manager.sh
aarnq: I mean that is what will become `storageAccountName` after it is created? Then we should use that as the variable for this.

We can add to init to set this to env name and similar for resource group.

scripts/azure/storage-manager.sh
```diff
@@ -3,9 +3,10 @@
 set -euo pipefail
 
 : "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
-: "${CK8S_ENVIRONMENT_NAME:?Missing CK8S_ENVIRONMENT_NAME}"
 : "${AZURE_LOCATION:?Missing AZURE_LOCATION}"
 
+CK8S_ENVIRONMENT_NAME="$(yq4 .global.ck8sEnvironmentName <<< "${common_config}")"
```
aarnq, line 8: Move this to where you read the `RESOURCE_GROUP` and it should be fine.

aarnq (approved):

bin/init.bash
```diff
@@ -212,6 +218,10 @@ set_object_storage() {
     esac
 
     replace_set_me "${file}" ".objectStorage.type" "\"${object_storage_type}\""
+    if [ "$object_storage_type" == "azure" ]; then
+        replace_set_me "${file}" ".objectStorage.azure.resourceGroup" "${CK8S_ENVIRONMENT_NAME}"
```
simonklb, line 222: I think we should distinguish this resource group from the other resource groups by suffixing it.

```suggestion
        replace_set_me "${file}" ".objectStorage.azure.resourceGroup" "${CK8S_ENVIRONMENT_NAME}-storage"
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,169 @@
...
+        case "$cmdinput" in
+        y|Y) return ;;
+        *) exit 0 ;;
+        esac
+    else
+        log_info "resource group [${RESOURCE_GROUP}] does not exist, creating it now" >&2
+        az group create \
+            --name "${RESOURCE_GROUP}" \
+            --location "${AZURE_LOCATION}" --only-show-errors
+    fi
```
simonklb, line 91: `az group create` is idempotent, no need to check if it already exists.

```suggestion
    log_info "creating resource group [${RESOURCE_GROUP}]"
    az group create \
        --name "${RESOURCE_GROUP}" \
        --location "${AZURE_LOCATION}" --only-show-errors
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,169 @@
...
+            log_info "creating storage account ${STORAGE_ACCOUNT}"
+            az storage account create \
+            --name "$STORAGE_ACCOUNT" \
+            --resource-group "${RESOURCE_GROUP}" \
+            --location swedencentral \
+            --sku Standard_RAGRS \
+            --kind StorageV2 \
+            --allow-blob-public-access false --only-show-errors
+            ;;
+    esac
```
simonklb, line 125: `az storage account create` is idempotent, no need to check if it already exists.

```suggestion
        log_info "creating storage account ${STORAGE_ACCOUNT}"
        az storage account create \
            --name "${STORAGE_ACCOUNT}" \
            --resource-group "${RESOURCE_GROUP}" \
            --location ${AZURE_LOCATION} \
            --sku Standard_RAGRS \
            --kind StorageV2 \
            --allow-blob-public-access false --only-show-errors
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,169 @@
...
+}
+
+function create_containers() {
+
```
simonklb, line 129:
```suggestion
```

scripts/azure/storage-manager.sh
```diff
@@ -0,0 +1,169 @@
...
+    CONTAINERS_LIST=$(az storage container list --account-name "$STORAGE_ACCOUNT" --query '[].name' --only-show-errors)
+
+    # shellcheck disable=SC2068
+    for container in ${CONTAINERS[@]}; do
+
```
simonklb, line 134:
```suggestion
```

simonklb (changes requested):

scripts/azure/storage-manager.sh
vomba: The check is for telling the user that it already exists and give them the option to change it.

scripts/azure/storage-manager.sh
vomba: same as resource group.

scripts/azure/storage-manager.sh
simonklb: Funny you removed all of the whitespaces except the one I pointed out. It's just a minor styling thing though so not a blocker!

simonklb (changes requested):
I think the only thing I'm missing now is that Azure CLI is added as a dependency in https://github.com/elastisys/compliantkubernetes-apps/blob/main/roles/get-requirements.yaml, after that this is ready to be shipped! :)

bin/common.bash
```diff
@@ -105,7 +105,7 @@ check_tools() {
   warn=0
   err=0
 
-  for executable in jq yq4 s3cmd sops kubectl helm helmfile dig pwgen htpasswd; do
+  for executable in jq yq4 s3cmd sops kubectl helm helmfile dig pwgen htpasswd az; do
```
simonklb, line 108: I don't think `az` is a requirement here, only the script, right?

roles/get-requirements.yaml
```diff
@@ -172,6 +172,23 @@
...
+
+        - name: Check azure-cli
+          shell: az version --query '"azure-cli"'
+          register: current_azure_cli_version
+          when: azure_cli_exists.stat.exists
+
+        - name: Get azure-cli
+          command: "curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash"
+          when: (not azure_cli_exists.stat.exists) or
+            (current_azure_cli_version.stdout == "")
```
simonklb, line 190: Is there a more predictable installation method? I'd like to be able to determine which version is getting installed.

I think you can install it via pip?

simonklb (changes requested):

bin/common.bash
vomba: The script needs az, so i thought it should be set a requirement there.

roles/get-requirements.yaml
vomba: Won't that make pip a requirement then?

bin/common.bash
simonklb: I _think_ this is only a check for tools that are required for running things under `bin/`. Please correct me if I'm wrong though!

roles/get-requirements.yaml
simonklb: Yea that is unfortunate. Is it available in apt repositories so that you can add it to `apt_list`?

bin/common.bash
vomba: So, My instinct directly led me to check s3cmd, which is technically invoked through `bin/` :laughing: , but i believe that we will be do the same with the azure creation, or maybe remove the s3cm command and make it run the proper thing depending on the storage type.

But, i think we can remove it for now until we decide on the implementation there.

roles/get-requirements.yaml
vomba: Yes, but it needs to add the Microsoft repo for it. I'll look into that.

bin/common.bash
aarnq: I'd say that we should not introduce this by the same rationale as why #2060 dropped the introduction of `CK8S_CLUSTER_PROVISIONER`.

The integration to make Azure work is targeting beta, and by introducing this we introduce an Azure CLI dependency for _everyone_ using Apps, and we should resolve the part from #2000 to make such dependencies optional depending on the infrastructure provider.

simonklb (approved):
Tried running it and had to fix some minor things: https://github.com/elastisys/compliantkubernetes-apps/pull/1996/commits/1534a96b96042fb8840e8ceb31d0f64ebcaa1fdc

Please check it before merging but from my side it's good to go! :tada:

bin/common.bash
vomba: Will remove the azure-cli parts before merging then.

raviranjanelastisys (approved):
LGTM too !!

PR #1862, opened 2023-11-10
Config Structural Validation using a declarative JSON Schema Part 0: Scripts, Global

Zash:
> [!warning]
> **This is public repository, ensure not to disclose:**
> - [ ] personal data beyond what is necessary for interacting with this pull request
> - [ ] business confidential information, such as customer names

### What kind of PR is this?

**Required**: Mark one of the following that is applicable:

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

*Optional*: Mark one or more of the following that are applicable:

> [!important]
> Breaking changes should be marked `kind/admin-change` or `kind/dev-change` depending on type
> Critical security fixes should be marked with `kind/security`

- [x] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

### Release notes

The configuration files are now validated against a JSON Schema.

### Platform Administrator notice

When validating config against the schema you may encounter cases where they disagree on what type a field should be and it may be necessary to tweak either the config or the schema to match the intent.

### What does this PR do / why do we need this PR?

...

- Fixes #1427

#### Additional information to reviewers

This adds a huge JSON Schema file (in YAML format) that describes the structure of the config. Feel free to ignore it when reviewing, it does not need to be perfect immediately but can be incrementally perfected over time.

A new tool, [yajsv](https://github.com/neilpa/yajsv), written in Go, is added as requirement which needs to be installed (`bin/install-requirements.bash`). Since JSON Schema is an an open specification with multiple implementations, it could be replaced in the future. We could even write our own using one of the available libraries.

To try this, run e.g.:

``` bash
./bin/ck8s validate sc
```

Information about JSON schema can be found at https://json-schema.org/

The specifications themselves are found at
- https://json-schema.org/draft/2020-12/json-schema-core (basics)
- https://json-schema.org/draft/2020-12/json-schema-validation (definitions of many kinds of constraints)

Notes about certain common schema fields:

- `default` can be used to describe a default value, which could be used in future tooling to actually fill in default values, but for now serves as documentation.
- `example` and `examples` can contain one or many example values. This could be shown in future tooling to improve UX.
- `format` allows specifying what kind of string to expect. Common usable values include `hostname`, `uri` (for URLs etc), `email`. Unknown values are ignored by the validator, so common formats could be described and potentially implemented in future tooling. 
[...]

PR #2063, opened 2024-03-27
Config Structural Validation Schema Part 1: Commons, HNC, CAPI, Harbor, Storage

Zash:
> [!warning]
> **This is a public repository, ensure not to disclose:**
>
> - [ ] personal data beyond what is necessary for interacting with this pull request, nor
> - [ ] business confidential information, such as customer names.

### What kind of PR is this?

**Required**: Mark one of the following that is applicable:

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

*Optional*: Mark one or more of the following that are applicable:

> [!important]
> Breaking changes should be marked `kind/admin-change` or `kind/dev-change` depending on type
> Critical security fixes should be marked with `kind/security`

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

### What does this PR do / why do we need this PR?

This adds structural validation for the following top-level config sections:

- user
- hnc
- clusterApi
- harbor
- clusterApi
- storageClasses
- objectStorage
- rookCeph

For more context including the required scripts and see #1862 
Part of issue #1427 

#### Information to reviewers

`ck8s validate` should now perform validation of the above sections.

While the CI is failing:
Please help determine whether it is the schema that is lacking or whether it detected a config mistake.

Suggestions and improvements very welcome!

[Also see the **README** for brief information about the schema format](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/schemas/README.md)

#### Checklist

- [ ] Proper commit message prefix on all commits
  
- Change checks:
  - [x] The change is transparent
  - [ ] The change is disruptive
  - [x] The change requires no migration steps
  - [ ] The change requires migration steps
  - [ ] The change upgrades CRDs
- Metrics checks:
  - [ ] The metrics are still exposed and present in Grafana after the change
  - [ ] The metrics names didn't change (Grafana dashboards and Prometheus alerts are not affected)
  - [ ] The metrics names did change (Grafana dashboards and Prometheus alerts were fixed)
- Logs checks:
  - [ ] The logs do not show any errors after the change
- Pod Security Policy checks:
  - [ ] Any changed pod is covered by Pod Security Admission
  - [ ] Any changed pod is covered by Gatekeeper Pod Security Policies
  - [ ] The change does not cause any pods to be blocked by Pod Security Admission or Policies
[...]

PR #2065, opened 2024-03-27
Config Schema Part 3: Networking, Ingress, DNS

Zash:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

This adds structural validation for the following top-level config sections:

- calicoAccountant
- calicoFelixMetrics
- certmanager
- ingressNginx
- issuers
- networkPolicies
- nodeLocalDns

For more context including the required scripts and **README** see #1862 

#### Information to reviewers

`ck8s validate` should now perform validation of the above sections.

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:


config/schemas/config.yaml
```diff
@@ -126,6 +411,654 @@ properties:
...
+                    - IPv4
+                    - IPv6
+                default:
+                  - IPv4
+              nodePorts:
+                additionalProperties: false
+                properties:
+                  http:
+                    default: 30080
+                    type: integer
```
anders-elastisys, line 550: I tested this out and I got some errors due to invalid type for `type: integer`, as we define e.g. ports as strings in our config. I think it is preferable if we use type string over type integer [as mentioned in the Helm docs as to avoid casting issues](https://helm.sh/docs/chart_best_practices/values/#make-types-clear), unless required by an upstream helm chart.
```suggestion
                    type: string
```

config/schemas/config.yaml
```diff
@@ -126,6 +411,654 @@ properties:
...
+              e.g. lack of documentation and disruptive bugs.
+            default: false
+            type: boolean
+          extraArgs: true
+          nodeSelector:
+            $ref: '#/$defs/nodeSelector'
+          resources:
+            $ref: '#/$defs/resources'
+          service:
+            additionalProperties: false
```
anders-elastisys, line 503: Is there some way to avoid validating part of a schema based  on the value of another variable? Because when I tested I had `service.enabled=false` and  thus the `service.type` is not set (or rather it is set to `set-me`) and it fails the schema validation:

```console
[ck8s] Validating wc config
[ck8s] Failed schema validation:
wc-config.yaml: ingressNginx.controller.service.type: ingressNginx.controller.service.type must be one of the following: "ClusterIP", "NodePort", "LoadBalancer", "ExternalName"
```

config/schemas/config.yaml
```diff
@@ -14,6 +14,291 @@ $defs:
...
+    $comment: Real k8s resources might have existing schemas that could be reused as-is
+    additionalProperties: false
+    default:
+      nodeAffinity:
+        requiredDuringSchedulingIgnoredDuringExecution:
+          nodeSelectorTerms:
+            - matchExpressions:
+                - key: node-role.kubernetes.io/control-plane
+                  operator: Exists
+    properties:
```
anders-elastisys, line 27: I think `nodeAffinity` and `podAffinity` might need to be added here, as atleast for `nodeAffinity` the following schema validation failed:

```console
[ck8s] Validating wc config
[ck8s] Failed schema validation:
...
wc-config.yaml: ingressNginx.defaultBackend.affinity: Additional property nodeAffinity is not allowed
```

config/schemas/config.yaml
Zash: Yes, it's possible but more complex, using either `oneOf` or `if-then` constructs.

Do we want to keep the unused `set-me` entries? Deleting them from the configs when not used would make it less cluttered, and easier to validate if unused settings aren't present at all.

Example `oneOf` method:


```yaml
oneOf:
- properties:
    enabled:
      const: false
- properties:
    enabled:
      const: true
  # other properties
  required: [ type, ... ] # list of mandatory properties if enabled
```

config/schemas/config.yaml
Zash: I would expect unsigned 16 bit integers to survive most casting. Isn't it better to handle that coercion in the templating layer, keeping port numbers consistently as integers in the config?

Won't numbers as strings get cast back into numbers when passed trough the yaml templating and parsing unless they're explicitly quoted?

Based on grepping the READMEs in `helmfile.d/upstream` I gather that many port settings can be set to the empty string. Also about half port settings seem to be `int` (of those easily greppable, probably missed a ton). 

```
$ find helmfile.d/ -name README.md -exec grep -w port {} +|grep -w string|wc -l
31
$ find helmfile.d/ -name README.md -exec grep -w port {} +|grep -w int|wc -l
28
```

Both could be supported with
```suggestion
                    type:
                      - string
                      - integer
```

anders-elastisys: I would be fine with supporting both string and integer in this case.

config/schemas/config.yaml
anders-elastisys: I am not sure what would be best in this case, removing unused `set-me` entries would require some reworking of our current config and scripts. Maybe improving both with e.g. adding `oneOf` for templating which we have `enabled` for and rework `set-me`s a bit in a follow up task. Maybe someone else has an opinion on this?

config/schemas/config.yaml
```diff
@@ -126,6 +365,654 @@ properties:
...
+              ipFamilyPolicy:
+                type: string
+                description: |-
+                  Represents the dual-stack-ness requested or required by this Service. When
+                  utilizing an internal loadbalancer service (ie MetalLB), set this field to
+                  "RequireDualStack" if you want both IPv4 and IPv6 connectivity. The
+                  ipFamilies and clusterIPs fields depend on the value of this field.
+
+                  See [reference](https://kubernetes.io/docs/concepts/services-networking/dual-stack/)
+                enum:
```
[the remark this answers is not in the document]
aarnq: Yes, that sounds good.

config/schemas/config.yaml
Zash: Do we use symbolic names anywhere? I.e. how in some contexts `"https"` is a valid port number. If not then 0-5 digits could be enforced by regex.

config/schemas/config.yaml
```diff
@@ -126,6 +363,658 @@ properties:
...
+          nodeSelector:
+            $ref: '#/$defs/nodeSelector'
+          resources:
+            $ref: '#/$defs/resources'
+          service:
+            additionalProperties: false
+            properties:
+              annotations:
+                additionalProperties:
+                  type: string
```
anders-elastisys, line 459: I tested it on a different cloud provider today and got this error:
```console
wc-config.yaml: ingressNginx.controller.service.annotations.loadbalancer.openstack.org/proxy-protocol: Invalid type. Expected: string, given: boolean
```
This comes from config generated by an internal repository which did not have quotes around the annotation value `true`, and I am not sure if we want to leave it as string to better match the [Kubernetes documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects) which clearly states that keys and values of annotation mappings must be **strings**. Maybe we should fix so we always put quotes around the value of an annotation in our configs, so create a follow up issue(s)? But I wanted to hear others opinions on this @Zash @aarnq

config/schemas/config.yaml
```diff
@@ -126,6 +363,658 @@ properties:
...
+        examples:
+          - |-
+            example.com:53 {
+              errors
+              cache 30
+              reload
+              loop
+              forward . 127.0.0.1:9005
+            }
+        type: string
```
anders-elastisys, line 1014: [Our config for `nodeLocalDns.customConfig`](https://github.com/elastisys/compliantkubernetes-apps/blob/v0.37.0/config/config/common-config.yaml#L1090) is not set to anything (null), so I get this:
```
wc-config.yaml: nodeLocalDns.customConfig: Invalid type. Expected: string, given: null
```
This should fix it:
```suggestion
        type: 
          - string
          - "null"
```
Or if there is a better way to ignore it, since it is a optional configuration.

config/schemas/config.yaml
```diff
@@ -126,6 +363,658 @@ properties:
...
+            type: boolean
+        type: object
+      gatekeeper:
+        additionalProperties: false
+        properties:
+          enabled:
+            default: true
+            type: boolean
+        type: object
+      global:
```
anders-elastisys, line 693: Seem to be missing some properties here:
```console
wc-config.yaml: networkPolicies.kubeSystem: Additional property openstack is not allowed
wc-config.yaml: networkPolicies.alertmanager: Additional property enabled is not allowed
wc-config.yaml: networkPolicies.global: Additional property wcApiserver is not allowed
wc-config.yaml: networkPolicies.global: Additional property wcNodes is not allowed
```

config/schemas/config.yaml
aarnq: I think I'd prefer if, then, else for boolean flags given that it becomes very explicit.

And the false case could be allow anything for now, and I think we need to take a decision on how we want to handle optional configuration with this moving forward. But that will become more important once we have a way to render the default configuration from the schema.

config/schemas/config.yaml
aarnq: Yes, annotations here should follow Kubernetes annotations.

But I don't know if it would be more friendly to enforce that through the template process instead and allow any scalars to be more forgiving. 

(Though this is something that would need to be noted in the description of this parameter.)

config/schemas/config.yaml
aarnq: I'm generally against having things be nullable, as that is a different type according to most tools. So I see it kinda more as a bug that we haven't set the config to `""` by default to keep it empty.

config/schemas/config.yaml
Zash: How's https://github.com/elastisys/compliantkubernetes-apps/pull/2065/commits/82441f88d61b3296924e9a42a1548b161669cc37 ?

anders-elastisys: Now I get this, as some of our netpols have an `enabled` property (not all of them it seems, as we are using different netpol charts for some apps):
```
wc-config.yaml: networkPolicies.kubeSystem.upcloud: Additional property enabled is not allowed
```

I also noticed some missing properties for certManager:
```bash
sc-config.yaml: certmanager.tolerations: Invalid type. Expected: array, given: object
sc-config.yaml: certmanager.webhook.tolerations: Invalid type. Expected: array, given: object
sc-config.yaml: certmanager.cainjector: Additional property extraArgs is not allowed
sc-config.yaml: certmanager.cainjector.tolerations: Invalid type. Expected: array, given: object
```

The `tolerations` ones should be updated to the correct type (in this case changed to array) in the [config](https://github.com/elastisys/compliantkubernetes-apps/blob/v0.37.0/config/config/common-config.yaml#L711). Since we noticed other cases where the config should be updated, I was wondering if you are planning to update such issues in these PRs or would that be a follow up task?

Zash: Probably worth prioritizing CI blockers.

The end goal is that schema vs config disagreements can be resolved by a PR from anyone. It doesn't have to be perfect at first.

config/schemas/config.yaml
```diff
@@ -127,6 +366,712 @@ properties:
...
+            example.com:53 {
+              errors
+              cache 30
+              reload
+              loop
+              forward . 127.0.0.1:9005
+            }
+        type:
+          - string
+          - "null"
```
anders-elastisys, line 1070: Did you also try to update the [config](https://github.com/elastisys/compliantkubernetes-apps/blob/v0.37.0/config/config/common-config.yaml#L1090) to an empty string as well instead of adding `null` here as suggested by @aarnq ?

anders-elastisys (approved):
Nice, I tested running `validate` on a couple of different configs for different cloud providers without issues.

Zash:
I will rebase this once #2063 has been merged.

config/schemas/config.yaml
Zash: Turning it into an `if-then` then 😄

config/schemas/config.yaml
Zash: I observe that there are many annotations throughout that should probably be refactored into a common definition under `$defs`. Future task perhaps?

config/schemas/config.yaml
Zash: Yes, in fe957df9224ea7381bf2be96f9255ece26bef59b

config/schemas/config.yaml
```diff
@@ -1363,6 +1372,727 @@ properties:
...
+      dnsAutoscaler:
+        additionalProperties: false
+        properties:
+          enabled:
+            default: true
+            type: boolean
+        type: object
+      enableAlerting:
+        default: true
+        type: boolean
```
robinAwallace, line 1680: We should probably consider refactoring this. So that `enableAlerting` becomes `alerting.enabled`. But that is different task then this one.

config/schemas/config.yaml
```diff
@@ -1363,6 +1372,727 @@ properties:
...
           type: string
     type: object
+  calicoAccountant:
+    additionalProperties: false
+    properties:
+      backend:
+        default: iptables
+        enum:
+          - iptables
+          - nftables
```
robinAwallace, line 1382: When rendering to markdown we want to add this `meta:enum`. Could you add the them? Do not spend to much time writing a description for them :slightly_smiling_face:  
```suggestion
        enum:
          - iptables
          - nftables
        meta:enum:
          iptables: "Description of iptables"
          nftables: "Description of nftables"
```

robinAwallace (commented):
We noticed that we should add at least a title for each type when we were generating markdown text from the schemas. You can take some inspiration from my [PR](https://github.com/elastisys/compliantkubernetes-apps/pull/2137) :slightly_smiling_face: 

Overall LGTM :slightly_smiling_face:

config/schemas/config.yaml
Zash: Is this something from the new docs work? Don't think I have seen this way of documenting enum values. I would have expected a `description: "iptables dolor sit amet\n\nnftables lorem ipsum"` or similar

robinAwallace: Yes, it is from the new docs work :slightly_smiling_face: 
It will look like this in the end :slightly_smiling_face: 
image

Zash: Looks amazing! Added some words paraphrased from Wikipedia and from comments in `common-config.yaml` for now.

Zash:
> we should add at least a title for each type when we were generating markdown text from the schemas

I added a few titles. I have earlier copied some of the comments from `{common,sc,wc}-config.yaml` into titles and descriptions where it seemed reasonable, but in general I have focused on the structural validation part here.

config/schemas/config.yaml
```diff
@@ -144,13 +148,18 @@ $defs:
     minimum: 0
     title: Percentage
     type: number
+  port:
+    maximum: 65535
+    minimum: 1
+    type:
+      - integer
+      - string
+    regex: ^[0-9]{0,5}$
```
aarnq, line 157: What is the reason for allowing this to be a string?

aarnq (approved):

config/schemas/config.yaml
Zash: Leniency towards existing configs, see https://github.com/elastisys/compliantkubernetes-apps/pull/2065#discussion_r1570122860 but we should probably push everyone towards integers, but now or later?

aarnq: I would prefer now, but we could do it later.

Elias-elastisys: I've ran into issues when not storing numbers as strings, e.g. due to needing leading zero "0777" where the zero was discarded when it was interpreted as a number. I know this is not the case for ports as it has a well defined range but I'm wondering if we want to be consistent in always having numbers as strings or as many numbers as possible as integers and hoping we don't run into issues?

aarnq: We should have number as numbers in the furthest extent possible. Because that is the most intuitive, and JSON schema and YAML should have no problem handling octals, but if it does become a problem we set it as a string for the key that needs it.

I do not see why we should enforce a blanket "numbers as strings" for this since it make the configuration much less intuitive and unnecessary complex. Also we'd loose the ability to validate the numbers.

Elias-elastisys: Yeah I guess we will realize when there are issues and just put those numbers as strings. The reason why numbers as strings was suggested is because that's what Helm recommends. But I agree that we'd loose ability to validate numbers that way.

Zash: So it would be sensible to keep things as integers/numbers in the config and coercing / formatting them into the expected formats at the helmfile → helm boundary then?

I note that `0777` is treated as the integer 777 by [YAML](https://yaml.org/spec/1.2.2/#tag-resolution), it would have to be written `0o777` and then formatted to produce the string `"0777"` if that is the needed format.

aarnq: I see, then some tools do unstandard things with those.

Zash: So, now or later?

```diff
@@ -151,13 +151,7 @@ $defs:
   port:
     maximum: 65535
     minimum: 1
-    type:
-      - integer
-      - string
-    regex: ^[0-9]{0,5}$
-    $comment: |-
-      The string form should be phased out and replaced by integers which are then
-      coerced into strings during templating.
+    type: integer
   portlist:
     title: Port number
     description: A 16 bit unsigned integer
```

aarnq: I'd rather do it now.

Zash: Let's do it now then!

PR #2082, opened 2024-04-09
tests: Add releases unit tests

aarnq:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

This adds test coverage to the needs set on releases, and this should catch errors such as [these](https://github.com/elastisys/compliantkubernetes-apps/pull/2074)

#### Information to reviewers

This takes _**ages**_ to run since it has to render every release that it can, though I've tried to optimise it so that it only ever have to render every release once and then pull it from cache.

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:
- Config checks:
  - [ ] The schema was updated

### DIFF
--- a/helmfile.d/stacks/ingress-nginx.yaml
+++ b/helmfile.d/stacks/ingress-nginx.yaml
@@ -68,5 +68,7 @@ templates:
     chart: charts/ingress-nginx-probe-ingress
     installed: {{ .Values | get "wcProbeIngress.enabled" false }}
     version: 0.1.0
+    needs:
+      - ingress-nginx/networkpolicy
     values:
       - values/ingress-nginx-probe-ingress.yaml.gotmpl
--- a/tests/common/lib/env.bash
+++ b/tests/common/lib/env.bash
@@ -106,19 +106,16 @@ env.teardown() {
   rm -rf "${CK8S_CONFIG_PATH}"
 }
 
-env.cache_create() {
+# Create a private copy of the current config path
+env.private() {
   if [[ -z "${CK8S_CONFIG_PATH:-}" ]]; then
     fail "CK8S_CONFIG_PATH is unset!"
   fi
 
-  cp -r "${CK8S_CONFIG_PATH}" "${CK8S_CONFIG_PATH}-cache"
-}
+  local target
+  target="$(mktemp --directory)"
 
-env.cache_delete() {
-  rm -rf "${CK8S_CONFIG_PATH}" "${CK8S_CONFIG_PATH}-cache"
-}
+  cp -Tr "${CK8S_CONFIG_PATH}" "${target}"
 
-env.cache_restore() {
-  rm -rf "${CK8S_CONFIG_PATH}"
-  cp -r "${CK8S_CONFIG_PATH}-cache" "${CK8S_CONFIG_PATH}"
+  export CK8S_CONFIG_PATH="${target}"
 }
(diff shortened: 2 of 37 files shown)


Elias-elastisys (approved):

lunkan93 (approved):

simonklb (approved):
This is impressive work!

As far as I can tell it is doing what it is supposed to but some parts of the code was pretty hard to fully comprehend. I wonder if another language could make these kinds of tests be more readable, compared to long lines of `yq`. We don't want to trap ourselves in code that only one person will be able to fix in the future! :smile: 

That being said, the parts I tried out myself seemed to be doing what it's supposed to. Testing the dependencies in Helmfile over the rendered Kubernetes manifests is no easy task, so huge props to you!

aarnq:
> This is impressive work!

Thank you!

> As far as I can tell it is doing what it is supposed to but some parts of the code was pretty hard to fully comprehend. I wonder if another language could make these kinds of tests be more readable, compared to long lines of `yq`. We don't want to trap ourselves in code that only one person will be able to fix in the future! 😄

Probably, I think we can reevaluate what language we have our tests in when we've migrated to Golang for the bin scripts. I think it will be more natural to make that switch then as well.

I'll try to restructure and comment it a bit so it _might_ become easier to understand. :sweat_smile:

simonklb:
> Question to reviewers, do we think it would be better to test this per infrastructure provider or not?

In an ideal world we would run it for every "config variation", right? Doing it per infrastructure provider is at least one step closer to that so I vote yes.

> Also what do think of this method of making templated tests? It is _much_ faster, and it would allow one to inline the templating where one would need it making it easier to understand, and crucially in this PR split it into multiple files.

I like it! But I would also like that the generated files were commited and have a pre-commit hook that makes sure that there is no diff uncommited after generating them.

aarnq:
> > Question to reviewers, do we think it would be better to test this per infrastructure provider or not?
> 
> In an ideal world we would run it for every "config variation", right? Doing it per infrastructure provider is at least one step closer to that so I vote yes.
> 
> > Also what do think of this method of making templated tests? It is _much_ faster, and it would allow one to inline the templating where one would need it making it easier to understand, and crucially in this PR split it into multiple files.
> 
> I like it! But I would also like that the generated files were commited and have a pre-commit hook that makes sure that there is no diff uncommited after generating them.

Right, yeah I don't think we should do this per config variation right now given how slow they are. :sweat_smile: 

But I can look into the changes to templated tests. I think the main thing is that we need to reduce the amount of lines needed to be generated, so it might need to be split up differently. 
Like the way it is now with this new releases tests it would generate ~270 lines per infra provider, when it is just two lines that are different between those files.

aarnq:
Migrated to the new test generation method, and integrated these tests into the validate suite so it can take advantage of its cache as best as possible.

simonklb:
> Migrated to the new test generation method, and integrated these tests into the validate suite so it can take advantage of its cache as best as possible.

Caching in tests makes me worry. :smile: But I get why it's being done here.

PR #2088, opened 2024-04-15
Fix dropped ingress packets in azure

vomba:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

This PR fixes the problem with dropped ingress packets to the ingress nginx on azure, it also enables preserving the source IP of incoming traffic.

- Fixes https://github.com/elastisys/ck8s-issue-tracker/issues/250

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:
- Config checks:
  - [ ] The schema was updated

### DIFF
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -246,6 +246,7 @@ set_nginx_config() {
             service_enabled=false
             external_load_balancer=true
             ingress_using_host_network=true
+            allocate_loadbalancer_node_ports=true
             ;;
 
         citycloud | elastx)
@@ -256,6 +257,7 @@ set_nginx_config() {
             service_annotations='{}'
             external_load_balancer=false
             ingress_using_host_network=false
+            allocate_loadbalancer_node_ports=true
             ;;
 
         aws)
@@ -266,6 +268,7 @@ set_nginx_config() {
             service_annotations='{ "service.beta.kubernetes.io/aws-load-balancer-type": "nlb" }'
             external_load_balancer=false
             ingress_using_host_network=false
+            allocate_loadbalancer_node_ports=true
             ;;
 
         azure)
@@ -276,6 +279,7 @@ set_nginx_config() {
             service_annotations='{ "service.beta.kubernetes.io/azure-load-balancer-health-probe-request-path": "/healthz" }'
             external_load_balancer=false
             ingress_using_host_network=false
+            allocate_loadbalancer_node_ports=false
             ;;
 
         baremetal)
@@ -284,6 +288,7 @@ set_nginx_config() {
             service_enabled=false
             external_load_balancer=true
             ingress_using_host_network=true
+            allocate_loadbalancer_node_ports=true
             ;;
         none)
             return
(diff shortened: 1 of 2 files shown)


vomba:
- For `whitelist-source-range` it was there in the `common-config` by default, i don't have a clear idea if it is needed or not.
- The network policies that does look very weird, did that happen after you did the init? i don't remember touching anything there to make it behave like that. running a `diff` doesn't show that.
- `allocateLoadBalancerNodePorts` to use direct routing from load balancer and fix the NetworkPolicies issue, which gave the opportunity to run `externalTrafficPolicy` and preserve the source IP

bin/init.bash
```diff
@@ -295,8 +300,16 @@ set_nginx_config() {
     replace_set_me "${file}" '.ingressNginx.controller.service.enabled' "${service_enabled}"
     replace_set_me "${file}" '.networkPolicies.global.externalLoadBalancer' "${external_load_balancer}"
     replace_set_me "${file}" '.networkPolicies.global.ingressUsingHostNetwork' "${ingress_using_host_network}"
-    replace_set_me "${file}" '.ingressNginx.controller.service.allocateLoadBalancerNodePorts' 'true'
+    replace_set_me "${file}" '.ingressNginx.controller.service.allocateLoadBalancerNodePorts' "${allocate_loadbalancer_node_ports}"
 
+    if [ "${CK8S_CLOUD_PROVIDER}" = "azure" ]; then
+        replace_set_me "${file}" '.externalTrafficPolicy.local' 'true'
+        replace_set_me "${file}" '.networkPolicies.ingressNginx.ingressOverride.enabled' 'true'
+        replace_set_me "${file}" '.networkPolicies.ingressNginx.ingressOverride.ips' "[ 0.0.0.0/0 ]"
```
simonklb, line 308: The network policy bug is that this is currently being replaced in common but it should be replaced in the sc and wc configs

simonklb:
> * `allocateLoadBalancerNodePorts` to use direct routing from load balancer and fix the NetworkPolicies issue

Are you sure this is supported by Azure? I think it uses node ports for the health probes.

simonklb:
Note that disabling node port allocation does not also automatically remove the node ports: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation

This means that even though you set it to false and applies it on an existing cluster it will not take effect.

vomba:
> > * `allocateLoadBalancerNodePorts` to use direct routing from load balancer and fix the NetworkPolicies issue
> 
> Are you sure this is supported by Azure? I think it uses node ports for the health probes.

It creates a specific port for health checks.

vomba:
> Note that disabling node port allocation does not also automatically remove the node ports: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation
> 
> This means that even though you set it to false and applies it on an existing cluster it will not take effect.

Yeah, i completely removed and then reapplied ingress-nginx so that the new load balancers get created without them.

Xartos (approved):

bin/init.bash
simonklb: Yes, but it's not a `set-me` value in that config file.

vomba: Weirdly enough, when i add `ips: set-me` i get bad expression error, but without it, the init works fine.

config/config/common-config.yaml
```diff
@@ -1041,7 +1041,8 @@ networkPolicies:
   ingressNginx:
     enabled: true
     ingressOverride:
-      enabled : false
+      enabled : set-me
```
aarnq, line 1044:
```suggestion
      enabled: set-me
```
Nit

aarnq (approved):

bin/init.bash
simonklb: Sent a fix for this here: https://github.com/elastisys/compliantkubernetes-apps/pull/2095

PR #2104, opened 2024-04-26
tests: Rework tests generators

aarnq:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

This reworks the generated bats tests to use go-templates to make the process of writing them be more natural.

This also makes generation part of the pre-commit hook so that one can see the impact of changes.

#### Information to reviewers

Do let me know what you think about this, I see it the template process as an improvement, but it will add a lot of noise when one makes changes as it like here adds ~70 loc for ~20 targets which yields 1400 loc in total. 

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:
- Config checks:
  - [ ] The schema was updated


simonklb:
I like it! The fact that you can see the results from your code in the PR is good IMO.

Could you maybe make the generating a Makefile target so that you can do it manually instead of having to commit to do it? If I was writing tests I would like to generate it and diff it in a short loop while developing.

aarnq:
> I like it! The fact that you can see the results from your code in the PR is good IMO.
> 
> Could you maybe make the generating a Makefile target so that you can do it manually instead of having to commit to do it? If I was writing tests I would like to generate it and diff it in a short loop while developing.

Yes, I'd want to make it so that it works as the previous model did so you only have to regenerate them if they changed and I also want to better package everything into a container so the generation can happen through it instead.

This was more of a stop gap to ensure that the formatting works against other hooks.

aarnq:
I think this will be it for this PR, I'll leave the `pods-ready.bats.yaml` for now since it will need more rethinking, and it is not used in the same capacity since we've yet to use it via `./bin/ck8s test`.

I'll also investigate the possibility to use the container more effectively when running tests, but that will be introduced in a different PR as well.

tests/common/gen.bash
```diff
@@ -0,0 +1,121 @@
...
+
+  file="$(readlink -f "$1")"
+
+  local -a input
+
+  readarray -t input < "${file}"
+
+  local describe test
+
+  local -a its
```
lunkan93, line 20: Might be a dumb question but what is the acronym here? :sweat_smile:

aarnq: It collects the `it`s declared in the `.cy.js` files, so it is not really an acronym :sweat_smile:

lunkan93: Aaah now it all makes sense, thanks :+1:

lunkan93 (approved):
I think this looks good :+1: I personally don't mind the noise aspect too much

robinAwallace:
It looks good to me, just one observation :slightly_smiling_face:  I ran the pre-commit hooks just to check how fast the templating was and I got a un-trackted file `tests/unit/bin/init.gen.bats`. Should that be part of this PR? :eyes:

aarnq:
> It looks good to me, just one observation 🙂 I ran the pre-commit hooks just to check how fast the templating was and I got a un-trackted file `tests/unit/bin/init.gen.bats`. Should that be part of this PR? 👀

I think that is left over from the old method, because no file should template into `init.gen.bats`, that one has been replaced by the tests in `tests/unit/bin/init/`.

You can run `make clean-gen` from `tests/` and then `make gen` and it should be up to date without untracked files.

robinAwallace (approved):

Issue #2031, opened 2024-03-01
[2] Upgrade Velero chart


AlbinB97:
### Chart to upgrade

vmware-tanzu/velero

### Reason for upgrade

- [X] Security patching
- [X] Upgrade to allow new features
- [X] Upgrade to prevent future issues
- [ ] Other - _Write the reason below_

### Additional context

Velero has been updated to fix CVEs. There's also some new features that could be useful, take a look at the changes here: https://github.com/vmware-tanzu/velero/releases

### Acceptance criteria

- I checked the migration of the new chart version:
  - [ ] The upgrade does not require migration steps
  - [ ] The upgrade requires migration steps
- I tested the functionality of the new chart version:
  - [ ] The upgrade does not change current functionality
  - [ ] The upgrade changes current functionality

anders-elastisys:
@Elias-elastisys as there was no chart released using Velero version v1.13.1, it currently uses v1.13.0. So if this is still needed I suggest that we either add templating for overriding the image version, or we update the chart once [this version is released](https://github.com/vmware-tanzu/helm-charts/pull/557).

PR #2141, opened 2024-05-16
Add `ck8s ops velero` command and clean up new velero finalizers

anders-elastisys:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

It is now possible to run `velero` commands with `ck8s ops velero <wc|sc>`

Running clean script got stuck when trying to delete Velero CRDs after having upgraded to the new Velero chart. This was caused by newly added finalizers to Velero resources (see breaking changes in [this Velero release](https://github.com/vmware-tanzu/velero/releases/tag/v1.12.0)). This PR adds a step for running `velero uninstall` before deleting the `velero` namespace or any of its CRDs, as it is the recommended way according to [their docs](https://velero.io/docs/v1.13/uninstalling/). To achieve this in the clean scripts, a new `ck8s ops` command was added for `velero` to be able to run the `velero` CLI against the correct cluster context.

- Follow up fix from #2031 

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/REQUIREMENTS
+++ b/REQUIREMENTS
@@ -11,5 +11,6 @@ pkg:golang/github.com/helmfile/helmfile@v0.162.0
 pkg:golang/github.com/mikefarah/yq/v3@3.4.1
 pkg:golang/github.com/mikefarah/yq/v4@v4.42.1
 pkg:golang/github.com/neilpa/yajsv@v1.4.1
+pkg:golang/github.com/vmware-tanzu/velero@v1.13.0
 pkg:golang/getsops/sops/v3@v3.8.1
 pkg:golang/helm.sh/helm/v3@v3.13.3
--- a/bin/ck8s
+++ b/bin/ck8s
@@ -28,6 +28,7 @@ usage() {
     #       confident that the apply command and migrations are good enough
     #       that direct Helmfile access is not necessary.
     echo "  ops helmfile <wc|sc>                           run helmfile as cluster admin" 1>&2
+    echo "  ops velero <wc|sc>                             run velero as cluster admin" 1>&2
     echo "  s3cmd [cmd]                                    run s3cmd" 1>&2
     echo "  kubeconfig <user|dev|admin>                    generate user kubeconfig, stored at CK8S_CONFIG_PATH/user"
     echo "  completion bash                                output shell completion code for bash" 1>&2
@@ -116,6 +117,11 @@ case "${1}" in
                 [[ "${3}" =~ ^(wc|sc)$ ]] || usage
                 shift 2
                 "${here}/ops.bash" helmfile "${@}"
+            ;;
+            velero)
+                [[ "${3}" =~ ^(wc|sc)$ ]] || usage
+                shift 2
+                "${here}/ops.bash" velero "${@}"
             ;;
             *) usage ;;
         esac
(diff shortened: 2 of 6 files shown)


simonklb (commented):
This feels like it's working around something that was added with a purpose. Is there something preventing us from fulfilling the finalizers instead of just patching them?

robinAwallace:
> This feels like it's working around something that was added with a purpose. Is there something preventing us from fulfilling the finalizers instead of just patching them?

I had the same thought. Should we try to use helmfile destroy more? Will that satisfy finalizers? Also I saw that Velero recommends doing `velero uninstall` to completely removing velero from a cluster.  

```
Prior to v1.12, deleting the Velero namespace would easily remove all the resources within it.
However, with the introduction of finalizers attached to the Velero CR including restore, dataupload, and datadownload in this version, directly deleting Velero namespace may get stuck indefinitely because the pods responsible for handling the finalizers might be deleted before the resources attached to the finalizers.
To avoid this issue, please use the command velero uninstall to delete all the Velero resources or ensure that you handle the finalizer appropriately before deleting the Velero namespace.
```

anders-elastisys:
> This feels like it's working around something that was added with a purpose. Is there something preventing us from fulfilling the finalizers instead of just patching them?

@simonklb ye, my reasoning would be that there should not matter in this case as we just want to be able to delete everything without interruption, but I will look into using `velero uninstall` instead as it is the recommended way.

> I had the same thought. Should we try to use helmfile destroy more? Will that satisfy finalizers?

@robinAwallace Helmfile destroy would not delete all Velero resources that have finalizers in this case since some are created after deployment such as `restores`.

> Also I saw that Velero recommends doing velero uninstall to completely removing velero from a cluster.

I saw that too and was considering using it, but wasn't sure about adding it as another dependency to apps, but it might be good to do, as then we can also ensure the Velero CLI is installed with the a version matching the cluster as well. I can look into adding that and uninstall Velero that way, as that is the recommended way.

simonklb:
> > This feels like it's working around something that was added with a purpose. Is there something preventing us from fulfilling the finalizers instead of just patching them?
> 
> @simonklb ye, my reasoning would be that there should not matter in this case as we just want to be able to delete everything without interruption

:+1: The only thing I'm worried about is that if we bypass the finalizers and delete we might end up leaving things behind. I assume the finalizers are there to make sure that everything gets cleaned up and terminated gracefully.

On the other hand, it's not such a big deal if backups are left in the object storage since we delete it eventually as well.

> but I will look into using `velero uninstall` instead as it is the recommended way.

Yea, if it's not a huge PITA it might be worth doing it "the right way". :smile:

anders-elastisys:
Added so Velero is installed with `install-requirements`, also added it as an `ops` command, let me know what you think of this, I think it can be a nice feature to run Velero for the correct cluster context, it makes it easier to run `velero uninstall` in this case from the clean scripts as well.

simonklb (approved):
Really nice! :)

Would love to see the fix to the challenges cleanup in a separate commit, but that is completely optional.

robinAwallace (approved):

PR #2162, opened 2024-06-05
Rework provider config templates

### LINKED ISSUE #2018: [5] Rework configuration templates
### Proposed feature

Currently we write infrastructure provider defaults during init, we should move this into plain configuration files that we merge during init instead.

We should also consider merging the sc and wc secrets as that is whats done anyway.

### Proposed alternatives

So the `config/` dir instead contains:

- `flavours/<flavours...>/`
- `providers/<infrastructure-providers...>/`
- `common-config.yaml`
- `sc-config.yaml`
- `wc-config.yaml`
- `secrets.yaml`

And then the init scripts will for flavours lookup if such a flavour exists under `config/flavours/` and similar for infrastructure providers in `config/providers/`.

### Additional context

If we would want to, for example to add commonality for OpenStack providers, we could add a `meta.yaml` which we could declare other flavours or providers to drag into the merge.

### Definition of done

- [ ] Reworked configuration templates

davidumea:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [x] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Reworked config templates so provider specific config is handled through static template manifests instead of the init script.

- Fixes #2018

#### Information to reviewers

The linked issue also suggests:
> If we would want to, for example to add commonality for OpenStack providers, we could add a meta.yaml which we could declare other flavours or providers to drag into the merge.

However I'm not sure that we want to do this, it could be confusing to keep track of where you should add your config when you want to add provider-specific config instead of just adding it for the provider you wish.

#### Checklist

- Change checks:
- Metrics checks:
  - [x] The metrics names didn't change (Grafana dashboards and Prometheus alerts are not affected)
- Logs checks:
- Pod Security Policy checks:
  - [x] Any changed pod is covered by Pod Security Admission
  - [x] Any changed pod is covered by Gatekeeper Pod Security Policies
  - [x] The change does not cause any pods to be blocked by Pod Security Admission or Policies
- Network Policy checks:
  - [x] Any changed pod is covered by Network Policies
  - [x] The change does not cause any dropped packets in the `NetworkPolicy Dashboard`
- Audit checks:
- Falco checks:
- Bug checks:


bin/init.bash
```diff
@@ -124,391 +124,38 @@ generate_default_config() {
...
         azure)
-            object_storage_type="azure"
-            yq4 --inplace '.objectStorage.azure.resourceGroup = "set-me"' "${file}"
-            yq4 --inplace '.objectStorage.azure.storageAccountName = "set-me"' "${file}"
-            ;;
-
-        baremetal)
-            object_storage_type="none"
+            replace_set_me "${file}" ".objectStorage.azure.resourceGroup" "\"${CK8S_ENVIRONMENT_NAME}-storage\""
+            replace_set_me "${file}" ".objectStorage.azure.storageAccountName" "\"${CK8S_ENVIRONMENT_NAME}\""
```
Elias-elastisys, line 137: Can this not also be envsubst or something? Would be nice to get rid of every single special case if possible.

davidumea: That would add another special case, also I am not sure but I get the feeling we want to avoid using envsubst :thinking:

What do other people think about this?

Elias-elastisys: Yeah true, I was just thinking that the whole point of this restructure was to avoid setting values in init. But I'm not sure what the best approach is for variable values. I know envsubst is not ideal.

aarnq: Since this ends up in the default config, and the environment name is readily available form the environment, then I think it should be fine to do envsubst here as we'd be able to remove the `set_object_storage` entirely.

We could generalise the envsubst we currently do to make it after everything is merged though, instead of just for the base config. 
Then it wouldn't be a special case, rather we'd have the ability to refer to certain dynamic values.

bin/init.bash
```diff
@@ -124,391 +124,38 @@ generate_default_config() {
...
-        replace_set_me "${file}" ".objectStorage.azure.resourceGroup" "\"${CK8S_ENVIRONMENT_NAME}-storage\""
-        replace_set_me "${file}" ".objectStorage.azure.storageAccountName" "\"${CK8S_ENVIRONMENT_NAME}\""
-    fi
 }
 
-# Usage: set_nginx_config <config-file>
-# baremetal support is experimental, keep as separate case until stable
-set_nginx_config() {
+# Usage: merge_provider_config <config-file>
+merge_provider_config() {
```
aarnq, line 146: Could we perhaps change this so we merge it in [together with the flavour](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/init.bash#L121)?

```diff
# in generate_default_config()

  new_config=$(mktemp)
  append_trap "rm ${new_config}; chmod 444 ${default_config}" EXIT

+ local -a files
+ files=("${config_template_path}/config/${config_name}" "${config_template_path}/config/flavors/${CK8S_FLAVOR}/${config_name}")
+
+ if [[ -f "${config_template_path}/config/providers/${CK8S_CLOUD_PROVIDER}/${config_name}" ]]; then
+   files+=("${config_template_path}/config/providers/${CK8S_CLOUD_PROVIDER}/${config_name}")
+ fi

- # Change this to use one flavor for cloud provider and one for "size" (e.g. dev, growth, enterprise)
- envsubst < "${config_template_path}/config/${config_name}" | yq_merge -  "${config_template_path}/config/flavors/${CK8S_FLAVOR}/${config_name}" > "${new_config}"
+ yq_merge "${files[@]}" | envsubst > "${new_config}"

  chmod 644 "${default_config}"
  cat "${new_config}" > "${default_config}"
```
And with the envsubst as general case for Azure Blob Storage.

aarnq (commented):
Would it be too much from me to ask to remove one directory level of `config`? :sweat_smile: 

So all in `config/config/` would rather be just in `config/`?

davidumea:
> Would it be too much from me to ask to remove one directory level of `config`? 😅
> 
> So all in `config/config/` would rather be just in `config/`?

No I think that would be for the best :+1:

Elias-elastisys (approved):
Lgtm!

aarnq (approved):
LGTM, really great to be able to remove all the specifics from the init scripts :+1:

PR #2193, opened 2024-06-20
config: add openstack as a provider

robinAwallace:
OAuth security improvement

> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [x] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Adds openstack as a generic provider.
For situations when apps is deployed on a openstack cloud but not safespring, elastx or citycloud. 

...

- Fixes #

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/providers/openstack/common-config.yaml
+++ b/config/providers/openstack/common-config.yaml
@@ -1,0 +1,42 @@
+openstackMonitoring:
+  enabled: true
+storageClasses:
+  default: cinder-csi
+networkPolicies:
+  kubeSystem:
+    openstack:
+      enabled: true
+      ips:
+        - set-me
+      ports:
+        - 5000
+        - 8774
+        - 8776
+  global:
+    externalLoadBalancer: false
+    ingressUsingHostNetwork: false
+  ingressNginx:
+    enabled: true
+    ingressOverride:
+      enabled: false
+objectStorage:
+  type: s3
+  s3:
+    region: set-me
+    regionEndpoint: set-me
+    forcePathStyle: true
+ingressNginx:
+  controller:
+    config:
+      useProxyProtocol: false
+    useHostPort: false
+    service:
+      enabled: true
+      type: LoadBalancer
+      annotations: {}
+      allocateLoadBalancerNodePorts: true
+externalTrafficPolicy:
+  local: false
+opa:
+  rejectLoadBalancerService:
+    enabled: false
--- a/README.md
+++ b/README.md
@@ -50,6 +50,9 @@ Currently we support four cloud providers:
 - Citycloud/Cleura
 - Elastx
 - UpCloud
+- Azure
+- Aws
+- Openstack
 - In addition to this we support running Compliant Kubernetes on bare metal (beta).
 
 ## Setup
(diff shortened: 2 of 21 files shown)


config/providers/openstack/common-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+      type: LoadBalancer
+      annotations: {}
+      allocateLoadBalancerNodePorts: true
+externalTrafficPolicy:
+  local: false
+opa:
+  rejectLoadBalancerService:
+    enabled: false
+calicoAccountant:
+  backend: iptables
```
anders-elastisys, line 44: Currently this is set by the `k8s-installer` config as well, maybe we should make a decision on where this should be set, since the `backend` is dependent on the OS version used :thinking:

config/providers/openstack/common-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+global:
+  clusterDns: 10.233.0.3
```
anders-elastisys, line 46: This would also be set in the `k8s-installer` config.

config/providers/openstack/common-config.yaml
Zash: And the OS version is decided by the installer? Makes sense to me!

Zash (approved):
LGTM
I especially like that the script got shorter!

config/providers/openstack/common-config.yaml
robinAwallace: 3d folder, with base image flavor? So one for ubuntu-2004 and one for 2204?

linus-astrom:
Can you also add Openstack to [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/README.md?plain=1#L130) so the apps README continues to have all providers.

Elias-elastisys (approved):

linus-astrom (approved):

simonklb (commented):
Schema needs to be updated to allow `openstack` as a valid value for `.global.ck8sCloudProvider`.

config/providers/openstack/common-config.yaml
anders-elastisys: I think we can keep it in the `installer` layer for now and make the assumption that for `capi` we use nftables (it will be added as a platform administrator notice in my PR), as we only have one variable dependent on the OS version. If there will be more in the future I think it could make sense to add an additional flavor.

anders-elastisys:
Could you generate tests for the provider as well by adding `openstack` [here](https://github.com/elastisys/compliantkubernetes-apps/blob/a3710eb2b76100d853b53b703294e3e68366f440/tests/unit/bin/init/template.bats.gotmpl#L50) and [here](https://github.com/elastisys/compliantkubernetes-apps/blob/a3710eb2b76100d853b53b703294e3e68366f440/tests/unit/validate/template.bats.gotmpl#L77)?

PR #2203, opened 2024-07-10
apps sc: Alert size over limit only on active indicies

aarnq:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

This changes the index over size limit alert to only consider indices with a positive document rate over the last hour, meaning it will eventually stop alerting for indices that are inactive such as after a successful rollover.

#### Information to reviewers

I haven't had the time to test it yet.

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml
+++ b/helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml
@@ -67,13 +67,13 @@ spec:
         summary: Opensearch Cluster Red (instance {{`{{ $labels.instance }}`}})
         description: Opensearch Cluster is in a Red status VALUE = {{`{{ $value }}`}}  LABELS = {{`{{ $labels }}`}}
     {{- range $prefixes := .Values.osIndexAlerts }}
-    - alert: {{ $prefixes.prefix | title }}SizeIncreasedOverLimit
-      expr: elasticsearch_indices_store_size_bytes_primary{namespace="opensearch-system",index=~"{{ $prefixes.prefix }}.+"} / (1024^2) > {{ $prefixes.alertSizeMB }}
+    - alert: OpenSearch{{ $prefixes.prefix | trimSuffix "-" | title }}IndexSizeOverLimit
+      expr: (elasticsearch_indices_store_size_bytes_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"} / (1024 ^ 2) > {{ $prefixes.alertSizeMB }}) * clamp(rate(elasticsearch_indices_docs_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"}[4h]) > 0, 1, 1)
       for: 15m
       labels:
         severity: warning
       annotations:
-        message: Primary shard size for index {{`{{ $labels.index }}`}} has increased over the limit of {{ $prefixes.alertSizeMB }}MB current size is {{`{{ $value | printf "%.0f" }}`}}MB
-        description: If the size keeps increasing, that might indicate a problem with ISM
+        message: Active primary shard size for index {{`{{ $labels.index }}`}} has increased over the limit of {{ $prefixes.alertSizeMB }}MB, current size is {{`{{ $value | printf "%.0f" }}`}}MB
+        description: This indicates a problem with index state management preventing the alias to roll over to a new index.
     {{- end }}
 {{- end }}


aarnq:
@elastisys/goto-logging-stack Would anyone of you like to take over this idea to make the alerts only report indices that are actively written to?

lunkan93:
Do you have time for this @salehsedghpour ? I might have time next week, but otherwise I wont have time until January, if not I can take it anyway.

salehsedghpour:
Yeah sure, I'll test it, I might have time for it next week.

salehsedghpour:
I've tested it and it works perfectly. I believe the use of `elasticsearch_indices_docs_primary`  instead of `elasticsearch_indices_docs_total` would be more precise, as the primary shard should get the and validate the values and then forward to other replicas. What do you think @lunkan93?

lunkan93:
> I've tested it and it works perfectly. I believe the use of `elasticsearch_indices_docs_primary` instead of `elasticsearch_indices_docs_total` would be more precise, as the primary shard should get the and validate the values and then forward to other replicas. What do you think @lunkan93?

Nice! :+1: 

I agree that it makes more sense to use `elasticsearch_indices_docs_primary` :smile:

helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml
```diff
@@ -67,13 +67,13 @@ spec:
         summary: Opensearch Cluster Red (instance {{`{{ $labels.instance }}`}})
         description: Opensearch Cluster is in a Red status VALUE = {{`{{ $value }}`}}  LABELS = {{`{{ $labels }}`}}
     {{- range $prefixes := .Values.osIndexAlerts }}
-    - alert: {{ $prefixes.prefix | title }}SizeIncreasedOverLimit
-      expr: elasticsearch_indices_store_size_bytes_primary{namespace="opensearch-system",index=~"{{ $prefixes.prefix }}.+"} / (1024^2) > {{ $prefixes.alertSizeMB }}
+    - alert: OpenSearch{{ $prefixes.prefix | trimSuffix "-" | title }}IndexSizeOverLimit
+      expr: (elasticsearch_indices_store_size_bytes_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"} / (1024 ^ 2) > {{ $prefixes.alertSizeMB }}) * clamp(rate(elasticsearch_indices_docs_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"}[1h]) > 0, 1, 1)
```
OlleLarsson, line 71: I'm not very proficcient in PromQL so could there be some explanation of the expression

The first part I get 
`(elasticsearch_indices_store_size_bytes_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"} / (1024 ^ 2)`
but it would still evaluate against all indices part of the regex and not just today's index right?

The second part is a bit more unclear to me

`{{ $prefixes.alertSizeMB }}) * clamp(rate(elasticsearch_indices_docs_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"}[1h]) > 0, 1, 1)`
```

maybe it is possible to mention how the magical `clamp` function works as to only have it evaluate the latest index?

Comment (salehsedghpour) [reply helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml:71]: So basically, the first part `(elasticsearch_indices_store_size_bytes_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"} / (1024 ^ 2) > {{ $prefixes.alertSizeMB }})` checks whether the indices size are larger than the predefined `alertSize` and it returns the value of it. 

Then the second part `clamp(rate(elasticsearch_indices_docs_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"}[1h]) > 0, 1, 1)` checks if there is any changes in the document count. If there is the `clamp` function will then clamps the sample values of all elements in the vector to have a lower limit of min and an upper limit of max which in this case it convert those values to 1. 

So basically, the whole expression returns the Active primary shard size for the given index if it is more than the `alertSizeMB`.

Comment (OlleLarsson) [reply helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml:71]: Thanks for the explanation. I will have to read it a couple of times :sweat_smile:

**Review by OlleLarsson — APPROVED** [review]

File: helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml [hunk 2]
```diff
@@ -67,13 +67,13 @@ spec:
         summary: Opensearch Cluster Red (instance {{`{{ $labels.instance }}`}})
         description: Opensearch Cluster is in a Red status VALUE = {{`{{ $value }}`}}  LABELS = {{`{{ $labels }}`}}
     {{- range $prefixes := .Values.osIndexAlerts }}
-    - alert: {{ $prefixes.prefix | title }}SizeIncreasedOverLimit
-      expr: elasticsearch_indices_store_size_bytes_primary{namespace="opensearch-system",index=~"{{ $prefixes.prefix }}.+"} / (1024^2) > {{ $prefixes.alertSizeMB }}
+    - alert: OpenSearch{{ $prefixes.prefix | trimSuffix "-" | title }}IndexSizeOverLimit
```
OlleLarsson, line 70: Q, should this alert only be installed when `indexPerNamespace: false` :thinking: ?

salehsedghpour: I would like to receive such alerts every time there is an issue with moving to the new index.

helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml
robinAwallace, line 71:
```suggestion
      expr: (elasticsearch_indices_store_size_bytes_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"} / (1024 ^ 2) > {{ $prefixes.alertSizeMB }}) * clamp(rate(elasticsearch_indices_docs_primary{namespace="opensearch-system", index=~"{{ $prefixes.prefix }}.+"}[2h]) > 0, 1, 1)
```
I took a look at long-running and found that increasing the sample window makes the metrics less choppy. But I still see that the active index can still be not considered as active with this function. So the alert can start flipping. 

But I do like this change and it will probably not happen that much. So I think it is fine to try it out like this and update it later on, if we notice that the alert is flipping between active and not.

OlleLarsson: > I took a look at long-running and found that increasing the sample window makes the metrics less choppy. But I still see that the active index can still be not considered as active with this function. So the alert can start flipping.

Sounds like we should spend some time to make sure that the alert don't start flipping.

OlleLarsson: I'm thinking, doesn't increasing the rate window make the alert fire for longer after it has been resolved, or capture the odd case where it is a long break in between new a new document being indexed? I have tried to look at the alert expression and I'm a bit puzzled on why it sometimes does not always consistently fire, I'm wondering if the exporter failed to scrape the metrics for some time.

salehsedghpour: @elastisys/goto-monitoring-stack do you have any input on this?

lunkan93: So I've had a look at this and here are my thoughts:

As @robinAwallace suggested 2h seems to be a good window, as there are frequent 1 hour windows where no documents are written to the `authlog` index, which would cause alerts to flip as the rate would go to 0 in this situation, and then back up once a document is written again.

And as @OlleLarsson pointed out, this would unfortunately have the side effect that an index would still considered active for 2 hours after a rollover. I think this is something that we can live with, and should be an improvement over the current state of these alerts.

aarnq: > ..., I'm wondering if the exporter failed to scrape the metrics for some time.

My guess is timed out due to resource limits either on the exporter or OpenSearch itself.

> ..2h seems to be a good window, as there are frequent 1 hour windows where no documents are written to the `authlog` index, ..., this would unfortunately have the side effect that an index would still considered active for 2 hours after a rollover.

I think this is reasonable as it still would close eventually rather than when the index is deleted. I would be fine for it to be even longer if that would ensure it wouldn't flap.

lunkan93: I don't mind setting it to something like 4h either to be extra safe, does anyone object to this?

OlleLarsson (approved):

robinAwallace (approved):

PR #2237, opened 2024-08-14
Added prometheus metrics to the diagnostics script

Elias-elastisys:
> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Adds the possibility to query prometheus metrics using the diagnostics script.

- Fixes https://github.com/elastisys/ck8s-issue-tracker/issues/285

#### Information to reviewers
Do we want more metrics?

<details>
<summary>Here is an example output</summary>

```
❯ ./bin/ck8s diagnostics sc --with-metrics-since 1day
[ck8s] Running diagnostics...
[ck8s] Validating sc config
[ck8s] Using unencrypted kubeconfig /home/elias/Documents/environments/coffee-brewer-dev/.state/kube_config_sc.yaml
------------------------------------------------------------------------------------------------------------------
Querying fluentd output error rate.
Fluentd output error rate over 0 on the dates:
ons 14 aug 2024 03:02:00 CEST
ons 14 aug 2024 08:44:00 CEST
ons 14 aug 2024 08:45:00 CEST

------------------------------------------------------------------------------------------------------------------
Querying dropped packages.
Found dropped packages going to pod: "fluentd-aggregator-0" on dates:
ons 14 aug 2024 08:44:00 CEST
ons 14 aug 2024 08:45:00 CEST

Found dropped packages going from pod: "fluentd-forwarder-j54f2" on dates:
ons 14 aug 2024 08:42:00 CEST
ons 14 aug 2024 08:43:00 CEST
ons 14 aug 2024 08:44:00 CEST
ons 14 aug 2024 08:45:00 CEST

Found dropped packages going from pod: "fluentd-forwarder-kh98r" on dates:
ons 14 aug 2024 08:42:00 CEST
ons 14 aug 2024 08:43:00 CEST
ons 14 aug 2024 08:44:00 CEST
ons 14 aug 2024 08:45:00 CEST

Found dropped packages going from pod: "fluentd-forwarder-qlbjq" on dates:
ons 14 aug 2024 08:42:00 CEST
ons 14 aug 2024 08:43:00 CEST
ons 14 aug 2024 08:44:00 CEST
ons 14 aug 2024 08:45:00 CEST

Found dropped packages going from pod: "fluentd-forwarder-svtsz" on dates:
ons 14 aug 2024 08:42:00 CEST
ons 14 aug 2024 08:43:00 CEST
ons 14 aug 2024 08:44:00 CEST
ons 14 aug 2024 08:45:00 CEST

Found dropped packages going from pod: "thanos-query-query-frontend-85f8d59458-fhthk" on dates:
tis 13 aug 2024 16:54:00 CEST
tis 13 aug 2024 16:55:00 CEST

Found dropped packages going to pod: "thanos-query-query-frontend-85f8d59458-fhthk" on dates:
tis 13 aug 2024 16:54:00 CEST
tis 13 aug 2024 16:55:00 CEST

Found dropped packages going from pod: "hnc-controller-webhook-57777c7778-8gxv7" on dates:
tis 13 aug 2024 16:42:00 CEST

------------------------------------------------------------------------------------------------------------------
Querying uptime.
The target "fluentd" was down on dates:
ons 14 aug 2024 08:44:00 CEST
ons 14 aug 2024 08:45:00 CEST

------------------------------------------------------------------------------------------------------------------
Querying Opensearch cluster status
Opensearch is in "yellow" state!
```

</details>

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:


simonklb:
I've not yet looked into the specifics, but this looks more and more like we are starting to export everything which we usually get access to by having access to the clusters and the dashboards themselves. It does not sounds feasible that we continue down this road as eventually we will make a complete data export of everything just to be able to debug customer clusters.

Could this be solved differently? VPN access? Customer OPS training?
Ping @cristiklein @llarsson @elastisys/sme

salehsedghpour:
> I've not yet looked into the specifics, but this looks more and more like we are starting to export everything which we usually get access to by having access to the clusters and the dashboards themselves. It does not sounds feasible that we continue down this road as eventually we will make a complete data export of everything just to be able to debug customer clusters.
> 
> Could this be solved differently? VPN access? Customer OPS training? Ping @cristiklein @llarsson @elastisys/sme

Thanks @simonklb. Actually, we discussed this in a meeting before summer, the meeting notes are linked in the issue (please let me know if you don't have access to it). Generally speaking, we aim to be able to support customers even those who are running air-gapped environments, and some of them might need to limit our access. That's why we are trying improve diagnostic script in such way.

salehsedghpour (commented):
Awesome work @Elias-elastisys! I believe the `diagnostic-script-sa-rbac` chart should be enabled first to be able to run this script. Right? 
If yes, can you make sure that we have proper documentation in public docs and/or in usage script?

Elias-elastisys:
> If yes, can you make sure that we have proper documentation in public docs and/or in usage script?

As it is now, it is always enabled if thanos is enabled. Thought it was such a small thing anyways. Should I change that?

salehsedghpour:
> > If yes, can you make sure that we have proper documentation in public docs and/or in usage script?
> 
> As it is now, it is always enabled if thanos is enabled. Thought it was such a small thing anyways. Should I change that?

I'm not sure about the impacts that it might have on our security promises. Do you have any input on this @cristiklein?

Elias-elastisys:
> I'm not sure about the impacts that it might have on our security promises. Do you have any input on this @cristiklein?

For some context, you need the ServiceAccount token in order to get access by the kubernetes API. Which means the person running the diagnostics scripts needs access to describe secrets in the thanos namespace in order to even get past kube api. So in my opinion I did not feel it opened up any new security issues.

salehsedghpour:
Right, I was mainly thinking about  creating SA (I believe it's a quick procedure), and then get the data and then delete the created resources when calling diagnostic script. The main idea is to close everything unless you need it. 

But I see your point, it might not have any drastic impact. But I would love to see what @OlleLarsson and @cristiklein think about it.

bin/diagnostics.bash
```diff
@@ -308,14 +324,86 @@ get_config_files() {
     done
 }
 
+run_diagnostics_with_metrics() {
+    TOKEN=$("${here}/ops.bash" kubectl "${cluster}" get secret diagnostics-script-access-token -n thanos -ojsonpath='{.data.token}' | base64 -d)
```
cristiklein, line 328: I'm unsure to understand why we need a SA token here. Why can't the platform administrator use their Kubernetes API access to proxy into Thanos directly?

I'd prefer the diagnostic script not to need a new token. Instead, it should "just" do whatever it can with the privileges of the user invoking it.

cristiklein (changes requested):

bin/diagnostics.bash
llarsson: I kind of like that an SA and a clear Role would be used, so that it is clear what one intends to do with the access that is granted. Makes it much easier to audit, and therefore to not "trust" but rather **verify**.

I would rather see "run this as an admin plz" like the trend of wget-ting some script and piping it into a sudo shell.

simonklb: I like the idea of using a service account token as well, but I think I'd like us to use the `TokenRequest` API via `kubectl create token` instead of creating a dedicated long-lived token.

simonklb (commented):
I have a feeling this is not the end of the expansion of this, errors are going to crop up that are not debugable only by looking at metrics. If I'm correct then I think we should consider turning this into it's own project and/or CLI.

It could even be worth investigating if running the diagnostic as a job inside of the cluster rather than proxying endpoints could be a way forward. The resulting artifacts could be uploaded to the object storage.

bin/diagnostics.bash
Elias-elastisys: I preferred using a token as to not require some awkward handling of a `kubectl proxy` background process in the script. Will have to keep the SA and RBAC but can switch to creating a token with a duration with kubectl instead.

bin/diagnostics.bash
```diff
@@ -7,32 +7,48 @@ here="$(dirname "$(readlink -f "$0")")"
...
     log_info "any GPG keys listed in CK8S_CONFIG_PATH/.sops.yaml or set by CK8S_PGP_FP"
     log_info ""
-    log_info "\t-h, --help\t\tdisplay help for this command and exit"
-    log_info "\t-n, --namespace\t\trun diagnostics for specified namespace only"
-    log_info "\t    --include-config\tinclude config yaml files found in CK8S_CONFIG_PATH"
+    log_info "\t-h, --help\t\t\tdisplay help for this command and exit"
+    log_info "\t-n, --namespace\t\t\ttrun diagnostics for specified namespace only"
+    log_info "\t    --include-config\t\tinclude config yaml files found in CK8S_CONFIG_PATH"
+    log_info "\t    --with-metrics-since\tquery a predefined set of metrics since the specified date"
+    log_info "\t    --query-metric\t\tquery any arbitrary metric"
```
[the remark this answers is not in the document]
Elias-elastisys: Sure makes sense, I can change

bin/diagnostics.bash
aarnq: Would it be possible to use kubectl oidc-login manually to retrieve an id token for the current user? I'm also not a fan of this setup either.

aarnq: I've tested this and it actually works, the only thing is that you have to ensure that oidc-login is used, and that you have to gather all arguments from the target kubeconfig.

```bash
# Would be better to actually resolve the current context and associated user.
readarray -t args <<< "$(yq '.users[0].user.exec.args[]' <kubeconfig-path>)"

# Must be oidc-login here
[[ "${args[0]" == "oidc-login" ]] || exit 1

token="$(kubectl "${args[@]}" | yq '.status.token')"
```

simonklb: What I like about using a dedicated service account token is that we call the API with least amount of permissions required as well make way for the diagnostics to be executed as a machine job. It's kind of similar to AWS IAM roles.

This is just an opinion not a blocker from my side though.

aarnq: > What I like about using a dedicated service account token is that we call the API with least amount of permissions required as well make way for the diagnostics to be executed as a machine job. It's kind of similar to AWS IAM roles.

Sure, I would agree if this script wasn't part of the bin scripts, because we are already platform administrator and if we do a request we should do it as ourself, not via indirection.

If this script was placed somewhere else, to be run in say a Kubernetes job or just standalone then I would want it to to have proper roles setup, and service accounts for each intended runner.
There shouldn't be a global sa that everyone runs with, as the sa is the identity.

Elias-elastisys:
Changed to separate commands.
Changed token to oidc-login and removed the SA/rbac chart.
PTAL

bin/diagnostics.bash
```diff
@@ -7,32 +7,38 @@ here="$(dirname "$(readlink -f "$0")")"
...
```
simonklb, line 20: Fix the usage

bin/diagnostics.bash
```diff
@@ -308,15 +315,102 @@ get_config_files() {
...
+case "${sub_command}" in
+    namespace)
+        # check that namespace exists
+        "${here}/ops.bash" kubectl "${cluster}" get namespace "${command_arg}" > /dev/null
+        run_diagnostics_namespaced "${command_arg}" >"${file}" 2>&1
+        ;;
+    with-metrics-since)
+        # Verify date argument
+        date -d -"${command_arg}" > /dev/null
+        #run_diagnostics_with_metrics "${with_metrics_since}" >"${file}" 2>&1
```
simonklb, line 402: Remove?

simonklb (changes requested):

bin/diagnostics.bash
```diff
@@ -7,32 +7,38 @@ here="$(dirname "$(readlink -f "$0")")"
 source "${here}/common.bash"
 
 usage() {
-    log_info "usage: ck8s diagnostics <sc|wc> [--namespace namespace] [--include-config]"
+    log_info "usage: ck8s diagnostics <sc|wc> [--namespace namespace] [--include-config] [--with-metrics-since date] [--query-metric metric]"
```
aarnq: Nit
```suggestion
    log_info "usage: ck8s diagnostics <sc|wc> [--namespace namespace] [--include-config] [--with-default-metrics-since date] [--with-metric metric]"
```
or
```suggestion
    log_info "usage: ck8s diagnostics <sc|wc> [--namespace namespace] [--include-config] [--query-default-metrics-since date] [--query-metric metric]"
```
Just from the name I initially thought one was simply time range and the other the metric

bin/diagnostics.bash
aarnq: Align it with spaces.

Elias-elastisys: :+1:

bin/diagnostics.bash
Elias-elastisys: Yikes, was a bit too quick with it.

bin/diagnostics.bash
```diff
@@ -7,32 +7,38 @@ here="$(dirname "$(readlink -f "$0")")"
...
     case "${1}" in
-        -n | --namespace)
-            namespace="${2:-}"
-            shift
-            ;;
         --include-config)
             include_config=true
             ;;
+        namespace|with-metrics-since|query-metric)
+            [[ ${#} -lt 2 || "${2}" = -* || -n "$sub_command" ]] && usage
```
aarnq, line 40: How is this check supposed to work?
Shouldn't it be `[[ check that all is alright ]] || usage`, where `usage` terminates the script?

bin/diagnostics.bash
```diff
@@ -308,15 +315,102 @@ get_config_files() {
     done
 }
 
+run_diagnostics_with_metrics() {
+    # shellcheck disable=SC2016
+    readarray -t args <<< "$(yq4 '. as $root | ($root.contexts[] | select(.name == $root.current-context) | .context) as $context | ($root.users[] | select(.name == $context.user) | .user) as $user | $user.exec.args[]' "${config["kube_config_sc"]}")"
+    [[ "${args[0]}" == "oidc-login" ]] || exit 1
```
aarnq, line 324:
```suggestion
    [[ "${args[0]}" == "oidc-login" ]] || exit 1
```
You might want to add a `log_fatal` explain why we exit here.

bin/diagnostics.bash
```diff
@@ -308,15 +315,102 @@ get_config_files() {
...
+    TOKEN="$(kubectl "${args[@]}" | yq4 '.status.token')"
```
aarnq, line 325: This should be its own function.

bin/diagnostics.bash
Elias-elastisys: Its just the opposite, works the same. Do you want me to flip it?
`[[ is anything wrong ]] && usage`

aarnq: Did you validate this?
Because with the bash options we usually set, if a command returns `1` then the script should fail.
And if the check `[[ is anything wrong fail ]]` fails (so when all is good) then it should short circuit.

aarnq: I had to dig into some documentation about this and I am wrong, it does work, but at the same time I don't understand why it manages to work.

> An AND list has the form `command1 && command2`, _command2_ is executed if, and only if, _command1_ returns an exit status of zero (success).
> ...
> The return status of AND and OR lists is the exit status of the last command executed in the list.

So that happens, however as I tested a similar construct bash does indeed not treat that as a failure that will cause the script to terminate.
Even though the status after the check is `1`.

I would prefer if we used constructs that behave according to what should be right, rather than this exception.

aarnq: I did some more digging, apparently `errexit` explicitly allows this:
> The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, part of any command executed in a && or || list except the command following the final && or ||, any command in a pipeline but the last, or if the command’s return status is being inverted with !.

But I keep my stance on it.

Elias-elastisys: I've tested this and it works, and I'm not quite sure I understand what you mean.
The check currently makes sense to me in my head. We have some checks that can't be true, if any is true it short circuits to usage, usage prints usage and exits. If all checks are fine (i.e false) it does not run usage and continues on.
That is how an if statement makes sense to me, e.g.
```
if (somethings wrong) then {
    print error
    exit
}

rest of function
```


But I see we use `|| usage` everywhere else so I'll try to flip it just for sake of consistency.

aarnq: Yes, but here you are leaving out the actual if statement, and are instead running two commands chained by `&&`, which makes it different.
If you'd rewrite it to use an if statement then that would be fine with me as an alternative.

With `errexit` which we set, if a command fails the script fails.
However adding `&&` inhibits this behaviour.
And this is what's not making sense to me.

So normally we expect commands to cause the script to exit if the commands of a line fails.
However the happy path with `[[ error check ]] && usage` is that the line _should_ fail and not exit the script.
And I want to avoid constructs that use these exceptions to make it clearer on how they work.
That is the difference with `[[ ok check ]] || usage`, as that line succeeds if everything is ok.

Not sure if that makes it clearer or just more confusing. :sweat_smile:

Elias-elastisys: Ah I see, I understand. But does `||` not also inhibit this behaviour but just reversed? So those line also does not really make sense to me then with `errexit` set. The [Bash reference](https://www.gnu.org/software/bash/manual/bash.html#index-trap) seems to indicate that at least.
We explicitly exit in usage in both cases so we are not really relying on `errexit` anyway, or am I misunderstanding?

I'll change no matter what just to be the same everywhere just trying to make things clear 😅.

Elias-elastisys: Yeah I guess I understand. In my head just looking at the lines I still think my original way makes more sense as I see it as the if case but I guess that ignores bash being bash.

simonklb: > Yes, but I think the overall logic makes more sense there as we then have `true || ?`. The overall expression is true, so it make sense to continue.
> 
> Compare it to `false && ?` where the overall logic says `false`.

`true || ?` does not evaluate `?` in this case. It's the other way around:
```bash
$ true && echo foo
foo
$ false && echo foo
$ true || echo foo
$ false || echo foo
foo
```

simonklb: To me it makes perfect sense to use `[ ... ] || usage` => "check if command line is correct, else show usage".

bin/diagnostics.bash
```diff
@@ -308,15 +325,93 @@ get_config_files() {
...
-else
-    # check that namespace exists
-    "${here}/ops.bash" kubectl "${cluster}" get namespace "${namespace}" > /dev/null
-    run_diagnostics_namespaced >"${file}" 2>&1
-fi
+
+case "${sub_command}" in
+    namespace)
+        # check that namespace exists
+        "${here}/ops.bash" kubectl "${cluster}" get namespace "${command_arg}" > /dev/null
```
aarnq, line 400: Nit, for consistency.
```suggestion
        "${here}/ops.bash" kubectl "${cluster}" get namespace "${command_arg}" >/dev/null
```
Same for L405

aarnq (approved):
LGTM

bin/diagnostics.bash
```diff
@@ -308,15 +325,93 @@ get_config_files() {
     done
 }
 
+run_diagnostics_default_metrics() {
+    TOKEN=$(fetch_oidc_token)
```
cristiklein, line 329: I love this solution!

cristiklein (approved):

bin/diagnostics.bash
```diff
@@ -308,15 +325,93 @@ get_config_files() {
...
+    DOMAIN="https://kube.$(yq4 '.global.opsDomain' "${config["config_file_sc"]}"):6443"
+    ENDPOINT="${DOMAIN}/api/v1/namespaces/thanos/services/thanos-query-query-frontend:9090/proxy/api/v1"
+    HEADER="Authorization: Bearer ${TOKEN}"
+    RANGE_ARG="--data-urlencode start=$(date -d -"${1}" -Iseconds) --data-urlencode end=$(date -Iseconds) --data-urlencode step=1m"
```
simonklb, line 333: nit: Make these lowercase like the rest of the variable names

bin/diagnostics.bash
```diff
@@ -308,15 +325,93 @@ get_config_files() {
...
+        if [[ $(echo "${res}" | jq 'select(.data.result != [])') ]]; then
+            for row in $(echo "${res}" | jq -c '.data.result[]'); do
+                "${print_func}" "${row}" "${@:3}"
+            done
+        fi
+    }
+
+    # shellcheck disable=SC2317
+    print_dates() {
+        eval "${2}"
```
simonklb, line 349: Don't use `eval`. Create concrete functions instead of passing them as strings.

bin/diagnostics.bash
```diff
@@ -308,15 +325,93 @@ get_config_files() {
...
+}
+
+# run_diagnostics_query_metric <metric>
+run_diagnostics_query_metric() {
+    TOKEN=$(fetch_oidc_token)
+    DOMAIN="https://kube.$(yq4 '.global.opsDomain' "${config["config_file_sc"]}"):6443"
+    ENDPOINT="${DOMAIN}/api/v1/namespaces/thanos/services/thanos-query-query-frontend:9090/proxy/api/v1"
+    HEADER="Authorization: Bearer ${TOKEN}"
+
+    curl "${ENDPOINT}/query" -k --header "${HEADER}" --data-urlencode query="${1}" | jq
```
simonklb, line 391: non-blocker: Consider creating a "query thanos using curl" wrapper to avoid repeating building it.

simonklb (changes requested):

bin/diagnostics.bash
simonklb: > > Yes, but I think the overall logic makes more sense there as we then have `true || ?`. The overall expression is true, so it make sense to continue.
> > Compare it to `false && ?` where the overall logic says `false`.
> 
> `true || ?` does not evaluate `?` in this case. It's the other way around:
> 
> ```shell
> $ true && echo foo
> foo
> $ false && echo foo
> $ true || echo foo
> $ false || echo foo
> foo
> ```

My bad, should have read your message with the entire thread in context. I get what you mean with `true || ?` now. :smile:

bin/diagnostics.bash
```diff
@@ -308,15 +325,102 @@ get_config_files() {
     done
 }
 
+run_diagnostics_default_metrics() {
+    token=$(fetch_oidc_token)
+    domain="https://kube.$(yq4 '.global.opsDomain' "${config["config_file_sc"]}"):6443"
+    endpoint="${domain}/api/v1/namespaces/thanos/services/thanos-query-query-frontend:9090/proxy/api/v1"
+    header="Authorization: Bearer ${token}"
+    range_arg="--data-urlencode start=$(date -d -"${1}" -Iseconds) --data-urlencode end=$(date -Iseconds) --data-urlencode step=1m"
```
aarnq, line 333:
```suggestion
    range_arg=("--data-urlencode" "start=$(date -d "-${1}" -Iseconds)" "--data-urlencode" "end=$(date -Iseconds)" "--data-urlencode" "step=1m")
```

bin/diagnostics.bash
```diff
@@ -308,15 +325,102 @@ get_config_files() {
...
+
+    query_and_parse() {
+        query="${1}"
+        print_func="${2}"
+        # shellcheck disable=SC2086
+        res=$(curl "${endpoint}/query_range" -k -s --header "${header}" --data-urlencode query="${query}" $range_arg)
```
aarnq, line 339: This isn't a check that you should disable.
```suggestion
        res="$(curl "${endpoint}/query_range" -k -s --header "${header}" --data-urlencode query="${query}" "${range_arg[@]}")"
```

bin/diagnostics.bash
```diff
@@ -308,15 +325,102 @@ get_config_files() {
...
+        if [[ $(echo "${res}" | jq 'select(.data.result != [])') ]]; then
```
aarnq, line 340: First avoid `echo "${variable} | command`, use `command <<< "${variable}"`,
second be explicit about what you check here,
third wrap `$(command)` in quotes everywhere except for simple `for var in $(list-generator); do`where you can be _absolutely sure that the data will never contain spaces_, for all other times use `readarray` first.

Elias-elastisys: What is the issue with `echo "${variable} | command`? (seeing as its everywhere in our bin already)

aarnq: It's not really an issue but a general recommendation. It is more optimised, as it doesn't create a process, and it makes debugging easier when tracing enabled.
And the fact that it is everywhere is because we essentially never refactor anything.

Elias-elastisys: > second be explicit about what you check here
What do you mean exactly? What I want to check is if the array is empty or not. How do I do that more explicitly here?

aarnq: > > second be explicit about what you check here
> 
> What do you mean exactly? What I want to check is if the array is empty or not. How do I do that more explicitly here?

You have a test `[[ ]]` where you only provide the output of a command, which means it defaults to a `-n` check.
If that is what you want to do add the `-n` to make it explicit.

Elias-elastisys: Ah okay I see what you mean now. I'll try to make it more explicit in that regard.

Elias-elastisys: Think I have adressed all of your comments now.

helmfile.d/values/networkpolicies/service/thanos.yaml.gotmpl
```diff
@@ -87,6 +87,9 @@ policies:
         - rule: egress-rule-dns
         - rule: egress-rule-query
       ingress:
+        - rule: ingress-rule-apiserver
```
simonklb, line 90: Please add a comment that explains why this network policy is needed, something like:
```suggestion
        # This is needed to allow the diagnostic script proxying Thanos.
        - rule: ingress-rule-apiserver
```

simonklb (approved):

bin/diagnostics.bash
```diff
@@ -308,15 +325,100 @@ get_config_files() {
     done
 }
 
+run_diagnostics_default_metrics() {
+    token="$(fetch_oidc_token)"
+    domain="https://kube.$(yq4 '.global.opsDomain' "${config["config_file_sc"]}"):6443"
+    endpoint="${domain}/api/v1/namespaces/thanos/services/thanos-query-query-frontend:9090/proxy/api/v1"
+    header="Authorization: Bearer ${token}"
+    range_arg=("--data-urlencode" "start=$(date -d -"${1}" -Iseconds)" "--data-urlencode" "end=$(date -Iseconds)" "--data-urlencode" "step=1m")
```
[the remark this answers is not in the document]
Elias-elastisys: It cant. Needs to be array to be able to pass to the command correctly for proper word splitting. 
Previous way, just as a string, required me to disable a spellcheck which [was not allowed](https://github.com/elastisys/compliantkubernetes-apps/pull/2237#discussion_r1740528716)

bin/diagnostics.bash
```diff
@@ -308,15 +325,100 @@ get_config_files() {
...
+
+    query_and_parse() {
+        query="${1}"
+        print_func="${2}"
+        res="$(curl "${endpoint}/query_range" -k -s --header "${header}" --data-urlencode query="${query}" "${range_arg[@]}")"
+        if [[ $(jq '.data.result | length' <<< "${res}") -gt 0  ]]; then
+            readarray metric_results_arr < <(jq -c '.data.result[]' <<< "${res}")
+            for row in "${metric_results_arr[@]}"; do
```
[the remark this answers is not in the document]
Elias-elastisys: Just changed from that because of [this comment](https://github.com/elastisys/compliantkubernetes-apps/pull/2237#discussion_r1740535723)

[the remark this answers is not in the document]
simonklb: Sorry should have checked the previous comments first. Makes sense!

bin/diagnostics.bash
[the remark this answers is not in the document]
simonklb: Right, would have to be inline then. :+1:

aarnq (approved):
LGTM

PR #2249, opened 2024-08-27
Add ClusterIssuer and Issuer solvers schema

simonklb:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

I wanted to validate my cluster issuer solver config but noticed that we did not support that. This fixes that.

#### Information to reviewers

The schema change is really large since we don't narrow down the solvers config. I'm not sure if this is acceptable or not. Let me know what you think!

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -832,6 +832,3013 @@ $defs:
         description: Value is the taint value the toleration matches to. If the operator is Exists, the value should be empty, otherwise just a regular string.
         type: string
     type: object
+  # Ref: https://github.com/cert-manager/cert-manager/blob/7d797a45d7040eaad420d04dc829e2ae9457bb35/deploy/crds/crd-clusterissuers.yaml#L219
+  io.cert-manager.v1.clusterissuers.acme.solvers:
+    title: Solvers
+    description: |-
+      Solvers is a list of challenge solvers that will be used to solve
+      ACME challenges for the matching domains.
+      Solver configurations must be provided in order to obtain certificates
+      from an ACME server.
+      For more information, see: https://cert-manager.io/docs/configuration/acme/
+    type: array
+    items:
+      description: |-
+        An ACMEChallengeSolver describes how to solve ACME challenges for the issuer it is part of.
+        A selector may be provided to use different solving strategies for different DNS names.
+        Only one of HTTP01 or DNS01 must be provided.
+      type: object
+      properties:
+        dns01:
+          description: |-
+            Configures cert-manager to attempt to complete authorizations by
+            performing the DNS01 challenge flow.
+          type: object
+          properties:
+            acmeDNS:
+              description: |-
+                Use the 'ACME DNS' (https://github.com/joohoi/acme-dns) API to manage
+                DNS01 challenge records.
+              type: object
+              required:
+                - accountSecretRef
+                - host
+              properties:
+                accountSecretRef:
+                  description: |-
+                    A reference to a specific 'key' within a Secret resource.
+                    In some instances, `key` is a required field.
+                  type: object
+                  required:
+                    - name
+                  properties:
+                    key:
+                      description: |-
+                        The key of the entry in the Secret resource's `data` field to be used.
+                        Some instances of this field may be defaulted, in others it may be
+                        required.
+                      type: string
+                    name:
+                      description: |-
+                        Name of the resource being referred to.
+                        More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                      type: string
+                host:
+                  type: string
+            akamai:
+              description: Use the Akamai DNS zone management API to manage DNS01 challenge records.
+              type: object
+              required:
+                - accessTokenSecretRef
+                - clientSecretSecretRef
+                - clientTokenSecretRef
+                - serviceConsumerDomain
+              properties:
+                accessTokenSecretRef:
+                  description: |-
+                    A reference to a specific 'key' within a Secret resource.
+                    In some instances, `key` is a required field.
+                  type: object
+                  required:
+                    - name
+                  properties:
+                    key:
+                      description: |-
+                        The key of the entry in the Secret resource's `data` field to be used.
+                        Some instances of this field may be defaulted, in others it may be
+                        required.
+                      type: string
+                    name:
+                      description: |-
+                        Name of the resource being referred to.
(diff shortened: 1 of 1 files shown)


aarnq:
My thinking is that this is _way_ to much to be able to include a resource in the schema and I will see if we can make this easier somehow.

simonklb:
> My thinking is that this is _way_ to much to be able to include a resource in the schema and I will see if we can make this easier somehow.

I'd love to see us reducing the different possible config options here to only what we know and support instead of _everything_ that cert-manager allows in `solvers`. Should hopefully be easier to know what to add tests for and realize when something breaks upstream.

aarnq:
> > My thinking is that this is _way_ to much to be able to include a resource in the schema and I will see if we can make this easier somehow.
> 
> I'd love to see us reducing the different possible config options here to only what we know and support instead of _everything_ that cert-manager allows in `solvers`. Should hopefully be easier to know what to add tests for and realize when something breaks upstream.

I am on the opposite side, I want to pull in the relevant parts of the CRD as verbatim as possible and automatically update it from the chart, so it gets tracked if there are changes. :sweat_smile:

simonklb:
> > > My thinking is that this is _way_ to much to be able to include a resource in the schema and I will see if we can make this easier somehow.
> > 
> > 
> > I'd love to see us reducing the different possible config options here to only what we know and support instead of _everything_ that cert-manager allows in `solvers`. Should hopefully be easier to know what to add tests for and realize when something breaks upstream.
> 
> I am on the opposite side, I want to pull in the relevant parts of the CRD as verbatim as possible and automatically update it from the chart, so it gets tracked if there are changes. 😅

Regarding solving the schema explosion being able to directly reference upstream schemas would be fantastic.

Regarding maintainability of ck8s, keeping the config as small as possible I think would be preferable since then we can more accurately predict what users have put in their config so that we don't break it in future releases.

aarnq:
> > > > My thinking is that this is _way_ to much to be able to include a resource in the schema and I will see if we can make this easier somehow.
> > > 
> > > 
> > > I'd love to see us reducing the different possible config options here to only what we know and support instead of _everything_ that cert-manager allows in `solvers`. Should hopefully be easier to know what to add tests for and realize when something breaks upstream.
> > 
> > 
> > I am on the opposite side, I want to pull in the relevant parts of the CRD as verbatim as possible and automatically update it from the chart, so it gets tracked if there are changes. 😅
> 
> Regarding solving the schema explosion being able to directly reference upstream schemas would be fantastic.
> 
> Regarding maintainability of ck8s, keeping the config as small as possible I think would be preferable since then we can more accurately predict what users have put in their config so that we don't break it in future releases.

Yeah, I agree, but given that we would track this schema more closely it would be easy to see the entirety of how it changes.
And it we truly don't want to support config migrations for upstream schema changes then we can always add a big caution message that says that this has to migrated manually.

simonklb:
> > > > > My thinking is that this is _way_ to much to be able to include a resource in the schema and I will see if we can make this easier somehow.
> > > > 
> > > > 
> > > > I'd love to see us reducing the different possible config options here to only what we know and support instead of _everything_ that cert-manager allows in `solvers`. Should hopefully be easier to know what to add tests for and realize when something breaks upstream.
> > > 
> > > 
> > > I am on the opposite side, I want to pull in the relevant parts of the CRD as verbatim as possible and automatically update it from the chart, so it gets tracked if there are changes. 😅
> > 
> > 
> > Regarding solving the schema explosion being able to directly reference upstream schemas would be fantastic.
> > Regarding maintainability of ck8s, keeping the config as small as possible I think would be preferable since then we can more accurately predict what users have put in their config so that we don't break it in future releases.
> 
> Yeah, I agree, but given that we would track this schema more closely it would be easy to see the entirety of how it changes. And it we truly don't want to support config migrations for upstream schema changes then we can always add a big caution message that says that this has to migrated manually.

I'm good with that! At least if we keep track of what parameters are available in the schema it will be noticable. I would love to see a big red CAUTION: THIS IS OUT OF OUR HAND in front of all "additional properties", "extra config" and what not where everything goes. :smile:

Zash:
> I'd love to see us reducing the different possible config options here to only what we know and support instead of _everything_ that cert-manager allows in `solvers`.

Simplifying config, reducing complexity is nice.

> I want to pull in the relevant parts of the CRD as verbatim as possible and automatically update it from the chart, so it gets tracked if there are changes. 😅

Makes sense where we do pass things through as-is.

Elias-elastisys:
> I'd love to see us reducing the different possible config options here to only what we know and support instead of _everything_ that cert-manager allows in `solvers`. Should hopefully be easier to know what to add tests for and realize when something breaks upstream.

I like the idea of this, trying to keep the possible config as small as possible. But I feel as if we would need to do the same then for every other part of config that accepts arbitrary yaml (extra args etc) which I feel would make the schema explode and hard to maintain but also might lock us in too much. Since we rely so much on upstream I worry it might make chart upgrades and other upstream changes more difficult to deal with, resulting in us avoiding or delaying upgrades and falling behind.

simonklb:
> > I'd love to see us reducing the different possible config options here to only what we know and support instead of _everything_ that cert-manager allows in `solvers`. Should hopefully be easier to know what to add tests for and realize when something breaks upstream.
> 
> I like the idea of this, trying to keep the possible config as small as possible. But I feel as if we would need to do the same then for every other part of config that accepts arbitrary yaml (extra args etc) which I feel would make the schema explode and hard to maintain but also might lock us in too much. Since we rely so much on upstream I worry it might make chart upgrades and other upstream changes more difficult to deal with, resulting in us avoiding or delaying upgrades and falling behind.

The counter-point to this is that for every part of the config that accepts arbitrary yaml we have no idea what is being set by the user. That means that when we upgrade something we risk breaking deployments without having a migration path.

kristiangronas:
> My thinking is that this is _way_ to much to be able to include a resource in the schema and I will see if we can make this easier somehow.

Hey, see https://github.com/elastisys/compliantkubernetes-apps/pull/2493 for how you could do this in a less painful way

PR #2043, opened 2024-03-12
Added new diagnostics script

Elias-elastisys:
> [!warning]
> **This is public repository, ensure not to disclose:**
> - [x] personal data beyond what is necessary for interacting with this pull request
> - [x] business confidential information, such as customer names

### What kind of PR is this?

**Required**: Mark one of the following that is applicable:

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

*Optional*: Mark one or more of the following that are applicable:

> [!important]
> Breaking changes should be marked `kind/admin-change` or `kind/dev-change` depending on type
> Critical security fixes should be marked with `kind/security`

- [x] kind/admin-change  
- [ ] kind/dev-change    
- [ ] kind/security      
- [ ] [kind/adr]()       

### Platform Administrator notice
Added a new diagnostics command with the primary goal of aiding in supporting self-managed customers. By default it encrypts the file with the sops config inside of `$CK8S_CONFIG_PATH` but setting `$CK8S_PGP_FP` will override this.

### What does this PR do / why do we need this PR?

Added a new diagnostics command which fetches the following information from a cluster into a file that is then sops encrypted:
* Getting and describing Nodes that are not ready
* Getting Deployments / DaemonSets / StatefulSets that have replicas that are not ready
* Getting and describing Pods that are not ready
* Getting the 20 newest lines of logs from Pods that are not ready if possible (also  --previous logs)
* Getting resource usage of nodes and pods
* Getting helm releases that are not deployed
* Getting various cert-manager resources
* Describing challenges that are not ready
* Getting and sorting all events

Fixes #2005 

#### Additional information to reviewers
Getting pods with -o yaml can give information not visible with describe. Do we want this also? My tests now generates a log with almost 8000 lines, and yaml would greatly increase that even further.

#### Checklist

- [x] Proper commit message prefix on all commits
  
- Change checks:
  - [x] The change is transparent
  - [ ] The change is disruptive
  - [ ] The change requires no migration steps
  - [ ] The change requires migration steps
- Metrics checks:
  - [ ] The metrics are still exposed and present in Grafana after the change
  - [ ] The metrics names didn't change (Grafana dashboards and Prometheus alerts are not affected)
  - [ ] The metrics names did change (Grafana dashboards and Prometheus alerts were fixed)
- Logs checks:
  - [ ] The logs do not show any errors after the change
[...]

PR #2250, opened 2024-08-28
Use file with GPG keys for diagnostics script and other improvements

anders-elastisys:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

The diagnostics script now assumes to retrieve GPG keys from a file `$CK8S_CONFIG_PATH/diagnostics_receiver.gpg` if the `CK8S_PGP_FP` environment variable is not set.

Want to simplify the usage of the diagnostics script, the script will now by default import GPG keys from a file `$CK8S_CONFIG_PATH/diagnostics_receiver.gpg` and use their fingerprints if `CK8S_PGP_FP` is not already set.

- Follow up of https://github.com/elastisys/ck8s-issue-tracker/issues/258

#### Information to reviewers

I am planning to update the public docs to reflect this change.

#### Checklist

- Change checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/bin/diagnostics.bash
+++ b/bin/diagnostics.bash
@@ -120,18 +120,18 @@ run_diagnostics() {
     printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
     if [ -d "${CK8S_CONFIG_PATH}/capi" ]; then
         # shellcheck disable=SC2002
-        capi_version=$(cat "${CK8S_CONFIG_PATH}"/capi/defaults/values.yaml | yq4 '.clusterApiVersion')
+        capi_version=$(cat "${CK8S_CONFIG_PATH}/capi/defaults/values.yaml" | yq4 '.clusterApiVersion')
         echo "CAPI version: ${capi_version}"
-    elif [ -d "${CK8S_CONFIG_PATH}/sc-config" ]; then
+    elif [ -d "${CK8S_CONFIG_PATH}/${cluster}-config" ]; then
         # shellcheck disable=SC2002
-        kubespray_version=$(cat "${CK8S_CONFIG_PATH}"/sc-config/group_vars/all/ck8s-kubespray-general.yaml | yq4 '.ck8sKubesprayVersion')
+        kubespray_version=$(cat "${CK8S_CONFIG_PATH}/${cluster}-config/group_vars/all/ck8s-kubespray-general.yaml" | yq4 '.ck8sKubesprayVersion')
         echo "Kubespray version: ${kubespray_version}"
     else
         echo "Can't find config directory"
     fi
     printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
     # shellcheck disable=SC2002
-    apps_version=$(cat "${CK8S_CONFIG_PATH}"/defaults/common-config.yaml | yq4 '.global.ck8sVersion')
+    apps_version=$(cat "${CK8S_CONFIG_PATH}/defaults/common-config.yaml" | yq4 '.global.ck8sVersion')
     echo "Apps version: ${apps_version}"
     printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
     # -- Nodes --
@@ -219,7 +219,7 @@ run_diagnostics() {
     # -- Helm --
     printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
     echo -e "\nFetching Helm releases that are not deployed (<helm>)"
-    helm=$("${here}"/ops.bash helm wc list -A --all -o yaml | yq4 '.[] | select(.status != "deployed")')
+    helm=$("${here}"/ops.bash helm "${cluster}" list -A --all -o yaml | yq4 '.[] | select(.status != "deployed")')
     if [ -z "${helm}" ]; then
         echo -e "All charts are deployed"
     else
--- a/bin/diagnostics.bash
+++ b/bin/diagnostics.bash
@@ -7,11 +7,12 @@ here="$(dirname "$(readlink -f "$0")")"
 source "${here}/common.bash"
 
 usage() {
-    log_info "usage: ck8s diagnostics <sc|wc> [namespace namespace] [query-default-metrics-since date] [query-metric metric] [--include-config]"
-    log_info
+    log_info "usage: ck8s diagnostics <sc|wc> [command] [options]"
+    log_info ""
     log_info "Collects diagnostics from the current environment set by CK8S_CONFIG_PATH and"
     log_info "store them in a file in the CK8S_CONFIG_PATH directory encrypted with SOPS using"
-    log_info "any GPG keys listed in CK8S_CONFIG_PATH/.sops.yaml or set by CK8S_PGP_FP"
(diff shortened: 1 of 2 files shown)


cristiklein:
@anders-elastisys I guess you wanted some early feedback on your proposal.

Overall, I think this is a good idea. It would be great if we could bring the script to the point where the on-call person only needs to type:

```
./bin/ck8s diagnostics sc
./bin/ck8s diagnostics wc
```

The expression [convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration) comes to my mind. The on-call person should be provided with "sensible defaults".

anders-elastisys:
> @anders-elastisys I guess you wanted some early feedback on your proposal.
> 
> Overall, I think this is a good idea. It would be great if we could bring the script to the point where the on-call person only needs to type:
> 
> ```
> ./bin/ck8s diagnostics sc
> ./bin/ck8s diagnostics wc
> ```
> 
> The expression [convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration) comes to my mind. The on-call person should be provided with "sensible defaults".

Maybe https://github.com/elastisys/compliantkubernetes-apps/pull/2250/commits/13414e179fe9f740586312bc50cf4648f1fdf4a1 is more in line with what we want? I removed the sops-config file flag and instead added so that by default the script will look for the file `${CK8S_CONFIG_PATH}/diagnostics_receiver.gpg` containing GPG keys that are then imported to the operators keys and used by SOPS later in the script. It is still possible to set `CK8S_PGP_FP` manually, but I removed the support for using `.sops.yaml` file as I deemed it not necessary for what we want this command to achieve. Let me know what you think.

cristiklein (approved):
I like the transparency of the error message and that we provide the platform administrator with sane defaults.

Elias-elastisys (approved):

salehsedghpour (approved):

bin/diagnostics.bash
```diff
@@ -42,37 +47,54 @@ while [ "${#}" -gt 0 ] ; do
...
 
-    log_warning "Notice for self-managed customers:"
-    echo -e "\tEncrypting using the fingerprints: $fingerprints." 1>&2
-    echo -e "\tIf you want to send diagnostic data to Elastisys, make sure to do:\n" 1>&2
+    echo -e "\tIf you want to send diagnostic data to Elastisys, make sure to store GPG keys" 1>&2
+    echo -e "\tretrieved during onboarding in a file named:\n" 1>&2
+    echo -e "\t\${CK8S_CONFIG_PATH}/diagnostics_receiver.gpg\n" 1>&2
 
-    echo -e "\tCK8S_PGP_FP=<fingerprint provided during onboarding> ./bin/ck8s diagnostics [sc|wc]\n" 1>&2
+    echo -e "\tIf in doubt, contact support@elastisys.com\n" 1>&2
```
salehsedghpour, line 65: Shouldn't our customer contact `sme-support@elastisys.com`?

salehsedghpour (approved):

bin/diagnostics.bash
```diff
@@ -42,37 +47,54 @@ while [ "${#}" -gt 0 ] ; do
...
-if [ -z "${CK8S_PGP_FP:-}" ]; then
-    fingerprints=$(yq4 '.creation_rules[].pgp' "${sops_config}")
+log_self_managed_notice() {
+    log_warning "WARNING: Notice for self-managed customers:"
 
-    log_warning "Notice for self-managed customers:"
-    echo -e "\tEncrypting using the fingerprints: $fingerprints." 1>&2
-    echo -e "\tIf you want to send diagnostic data to Elastisys, make sure to do:\n" 1>&2
+    echo -e "\tIf you want to send diagnostic data to Elastisys, make sure to store GPG keys" 1>&2
+    echo -e "\tretrieved during onboarding in a file named:\n" 1>&2
```
simonklb, line 62: This is not appropriate for this repository since it is open source. We should not assume that Elastisys is owning the platform.

bin/diagnostics.bash
simonklb, line 65: This is not appropriate for this repository since it is open source. We should not assume that Elastisys is owning the platform.

simonklb (changes requested):

bin/diagnostics.bash
anders-elastisys: The notice was added as when the script was first created, see the discussion [here](https://github.com/elastisys/compliantkubernetes-apps/pull/2043#discussion_r1523258042). Maybe @cristiklein can chip in regarding this?

cristiklein: I agree with @simonklb that this is one of the few instance when we talk with "company voice" as opposed to talking with "project voice". As far as I saw from other project, it is okay to advertise support for an open source project. Hence, I propose the following:

```
If you are an Elastisys self-managed customer, you can send diagnostic data to Elastisys.
Make sure to store GPG keys retrieved during onboarding in a file named:
```

bin/diagnostics.bash
anders-elastisys: Fixed it now in https://github.com/elastisys/compliantkubernetes-apps/pull/2250/commits/3ad844b283db5cade0dfd3b2e7fff84e2884a792

bin/diagnostics.bash
```diff
@@ -400,8 +420,18 @@ run_diagnostics_query_metric() {
...
+fi
+log_info "Using the following fingerprints:"
+log_info "${CK8S_PGP_FP}"
+
+file="${CK8S_CONFIG_PATH}/diagnostics-${cluster}-$(date +%y%m%d%H%M%S).log"
+touch "${file}"
+
 config_load "${cluster}"
+log_info "Running diagnostics..."
+export CK8S_AUTO_APPROVE=true
```
simonklb, line 434: Is there a reason why this is hardcoded instead of up to the user to set?

simonklb (approved):
LGTM! :)

bin/diagnostics.bash
anders-elastisys: Yes, otherwise the validation will trigger a prompt that is not visible when running the script since we are piping all output to a file in CK8S_CONFIG_PATH. The output from the validation will still be added to the diagnostics file, but with this it won't be seen as the script gets "stuck".

anders-elastisys: And since we already do a validation before running diagnostics, the user will already know if there is something deviating in their config, and hopefully fix it or at least be aware of it.

simonklb: Makes sense. It's important to note that `CK8S_AUTO_APPROVE` doesn't just automatically approve validation failure prompts but _all_ prompts. So be careful with what you run as part of the diagnostics collection. :smile:

anders-elastisys: Good point, I believe currently it is only the validation that has a prompt, but good to keep in mind. 
We should probably improve it somehow in the future, like better indication that the script is actually running and doing stuff, as it can take quite a while to run and sometimes it is unclear if it is running or stuck.

PR #2289, opened 2024-09-24
Fix update-ips fetch swift url

robinAwallace:
> [!warning]
> **This is a public repository, ensure not to disclose:**
>
> - [ ] personal data beyond what is necessary for interacting with this pull request, nor
> - [ ] business confidential information, such as customer names.

### What kind of PR is this?

**Required**: Mark one of the following that is applicable:

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

*Optional*: Mark one or more of the following that are applicable:

> [!important]
> Breaking changes should be marked `kind/admin-change` or `kind/dev-change` depending on type
> Critical security fixes should be marked with `kind/security`

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

### What does this PR do / why do we need this PR?

I noticed that the update-ips did not find the swift url. 
Im unsure if it is my version of curl that returns 1 less line. 
If that is the case, we probably need to find a more stable way extract the url from the response.

```
curl --version
curl 7.88.1 (x86_64-pc-linux-gnu) libcurl/7.88.1 OpenSSL/3.0.8 zlib/1.2.13 brotli/1.0.9 zstd/1.5.4 libidn2/2.3.3 libpsl/0.21.2 (+libidn2/2.3.3) libssh/0.10.4/openssl/zlib nghttp2/1.52.0 librtmp/2.3
Release-Date: 2023-02-20
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd

```

#### Checklist

- [ ] Proper commit message prefix on all commits
  
- Change checks:
  - [x] The change is transparent
  - [ ] The change is disruptive
  - [ ] The change requires no migration steps
  - [ ] The change requires migration steps
  - [ ] The change upgrades CRDs
  - [ ] The change updates the config *and* the schema
- Documentation checks:
  - [ ] The [public documentation](https://github.com/elastisys/compliantkubernetes) required no updates
[...]

PR #2302, opened 2024-10-08
bin: make update swift ip simpler

robinAwallace:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Continuation of this [PR](https://github.com/elastisys/compliantkubernetes-apps/pull/2289) that needed to be in a release.
Makes get swift ip simpler with extra test.

#### Checklist

- Change checks:
- Documentation checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/bin/update-ips.bash
+++ b/bin/update-ips.bash
@@ -182,11 +182,8 @@ get_swift_url() {
   auth_url="$(yq_read "sc" "${swift_config_option}.authUrl" "")"
   swift_region="$(yq_read "sc" "${swift_config_option}.region" "")"
 
-  headers=$(mktemp --suffix="_headers")
-  append_trap "rm ${headers}" EXIT
-
   if [ -n "$(yq_read_secret "${swift_config_option}.username" "")" ]; then
-    response=$(curl -s -D "${headers}" -H "Content-Type: application/json" -d '
+    response=$(curl -s -i -H "Content-Type: application/json" -d '
         {
           "auth": {
             "identity": {
@@ -208,7 +205,7 @@ get_swift_url() {
           }
         }' "${auth_url}/auth/tokens")
   elif [ -n "$(yq_read_secret "${swift_config_option}.applicationCredentialID" "")" ]; then
-    response=$(curl -s -D "${headers}" -H "Content-Type: application/json" -d '
+    response=$(curl -s -i -H "Content-Type: application/json" -d '
         {
           "auth": {
             "identity": {
@@ -225,8 +222,14 @@ get_swift_url() {
     exit 1
   fi
 
-  os_token=$(grep -oP "x-subject-token:\s+\K\S+" "${headers}")
-  swift_url=$(echo "${response}" | jq -r '.token.catalog[] | select( .type == "object-store" and .name == "swift") | .endpoints[] | select(.interface == "public" and .region == "'"${swift_region}"'") | .url')
+  {
+
+    while read -rd $'\r\n' header value; do
+      [[ -z "${header}" ]] && break
+      [[ "${header}" == "x-subject-token:" ]] && os_token="${value}"
+    done
+    swift_url=$(jq -r '.token.catalog[] | select( .type == "object-store" and .name == "swift") | .endpoints[] | select(.interface == "public" and .region == "'"${swift_region}"'") | .url')
+  } <<< "${response}"
 
   curl -i -s -X DELETE -H "X-Auth-Token: ${os_token}" -H "X-Subject-Token: ${os_token}" "${auth_url}/auth/tokens" >/dev/null
 
--- a/tests/common/bats/update-ips.bash
+++ b/tests/common/bats/update-ips.bash
@@ -26,20 +26,6 @@ update_ips.setup_mocks() {
   export -f kubectl
 }
 
-update_ips.setup_mktemp_mock(){
-  mock_mktemp="$(mock_create)"
-  export mock_mktemp
-  mktemp() {
-    # shellcheck disable=SC2317
-    touch /tmp/tmp.mock_headers
-    # shellcheck disable=SC2317
-    "${mock_mktemp}" "${@}"
-  }
-  export -f mktemp
-
-  mock_set_output "${mock_mktemp}" /tmp/tmp.mock_headers
-}
-
 update_ips.assert_mocks_none() {
   assert_equal "$(mock_get_call_num "${mock_curl}")" 0
   assert_equal "$(mock_get_call_num "${mock_dig}")" 0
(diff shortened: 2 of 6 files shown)


bin/update-ips.bash
```diff
@@ -224,9 +221,9 @@ get_swift_url() {
     log_error "Could not find Swift credentials in ${swift_config_option}"
     exit 1
   fi
-
-  os_token=$(grep -oP "x-subject-token:\s+\K\S+" "${headers}")
-  swift_url=$(echo "${response}" | jq -r '.token.catalog[] | select( .type == "object-store" and .name == "swift") | .endpoints[] | select(.interface == "public" and .region == "'"${swift_region}"'") | .url')
+  body=$(echo "${response}" | tr '\r\n' '!' | sed 's/!!/\n/g' | tail -n 1 | tr '!!' '\n')
+  os_token=$(echo "${response}" | grep -oP "x-subject-token:\s+\K\S+")
+  swift_url=$(echo "${body}" | jq -r '.token.catalog[] | select( .type == "object-store" and .name == "swift") | .endpoints[] | select(.interface == "public" and .region == "'"${swift_region}"'") | .url')
```
simonklb, line 226: This trick makes me a bit worried. Could there be a problem with existing exclamation marks in the response?

What about something like this?

```suggestion
{
  while read -d $'\r\n' header value; do
    [[ -z "${header}" ]] && break
    [[ "${header}" == "x-subject-token:" ]] && os_token="${value}"
  done

  swift_url=$(jq -r '.token.catalog[] | select( .type == "object-store" and .name == "swift") | .endpoints[] | select(.interface == "public" and .region == "'"${swift_region}"'") | .url')
} <<< "${response}"
```

This first reads all the headers, tries to find the token header and then passes the rest of the response (the body) to `jq`.

Ideally you would feed the output of `curl` directly to it rather than storing the response in a variable but don't know how much work that would be.

Note that I've not tested this so please try it out before committing it if you want to go with this solution instead. :)

tests/common/bats/update-ips.bash
```diff
@@ -134,7 +120,9 @@ update_ips.mock_swift() {
   update_ips.mock_minimal
 
   # GET /auth/tokens
-  mock_set_output "${mock_curl}" '{"token":{"catalog":[{"type": "object-store", "name": "swift", "endpoints": [{"interface":"public", "region": "swift-region", "url": "https://swift.foo.dev-ck8s.com:91011"}]}]}}' 1
+  mock_set_output "${mock_curl}" 'x-subject-token: 123456789
+
+  {"token":{"catalog":[{"type": "object-store", "name": "swift", "endpoints": [{"interface":"public", "region": "swift-region", "url": "https://swift.foo.dev-ck8s.com:91011"}]}]}}' 1
```
simonklb, line 125: I think it's good if we make this a bit more realistic so that we don't assume it works only because it passes a test with one header in it.

Something like this?

```suggestion
  mock_set_output "${mock_curl}" 'HTTP/2 200
date: Wed, 09 Oct 2024 14:14:40 GMT
expires: -1
cache-control: private, max-age=0
content-type: text/html; charset=ISO-8859-1
x-subject-token: 123456789
accept-ranges: none
vary: Accept-Encoding

  {"token":{"catalog":[{"type": "object-store", "name": "swift", "endpoints": [{"interface":"public", "region": "swift-region", "url": "https://swift.foo.dev-ck8s.com:91011"}]}]}}' 1
```

You could change it to headers that resemble the actual output better though.

simonklb (commented):
Love to see that the tests are improving! :)

bin/update-ips.bash
robinAwallace: Yeah, it is always abit iffy replacing characters. But I will try your solution and see if it works :+1:

tests/common/bats/update-ips.bash
robinAwallace: Good point, I will add them :+1:

tests/common/bats/update-ips.bash
```diff
@@ -63,7 +63,16 @@ update_ips.mock_maximal() {
...
-  mock_set_output "${mock_curl}" '{"token":{"catalog":[{"type": "object-store", "name": "swift", "endpoints": [{"interface":"public", "region": "swift-region", "url": "https://swift.foo.dev-ck8s.com:91011"}]}]}}' 1
+  mock_set_output "${mock_curl}" 'HTTP/2 200
+  date: Wed, 09 Oct 2024 14:14:40 GMT
+  expires: -1
+  cache-control: private, max-age=0
+  content-type: text/html; charset=ISO-8859-1
+  x-subject-token: 123456789
+  accept-ranges: none
+  vary: Accept-Encoding
+  \r\n
```
simonklb, line 74: Will this not print literal '\r\n'? Although I would not expect the test to succeed if that was the case, hmmm...

simonklb: Ah, I think we might get away with it because we never test what is getting fed to dig or curl. Could you add some `mock_get_call_args` to those mocks to make sure that the token and url is being parsed correctly?

robinAwallace: I have added one check here https://github.com/elastisys/compliantkubernetes-apps/blob/rw/make-get-swift-ip-simpler/tests/unit/general/bin-update-ips-swift.bats#L179. Would you like to have more of them?

simonklb: It would be nice to also verify that the swift URL is being parsed correctly. If it's done correctly it should eventually be passed as an argument to `dig`.

bin/update-ips.bash
```diff
@@ -225,8 +222,14 @@ get_swift_url() {
     exit 1
   fi
 
-  os_token=$(grep -oP "x-subject-token:\s+\K\S+" "${headers}")
-  swift_url=$(echo "${response}" | jq -r '.token.catalog[] | select( .type == "object-store" and .name == "swift") | .endpoints[] | select(.interface == "public" and .region == "'"${swift_region}"'") | .url')
+  {
+    # shellcheck disable=SC2162
+    while read -d $'\r\n' header value; do
```
simonklb, line 227: My bad missing this. I don't think there is any reason for not using the `-r` flag.

```suggestion
    while read -r -d $'\r\n' header value; do
```

simonklb (approved):
Great work!

PR #2310, opened 2024-10-14
config: move harbor backup schedule and make it configurable

viktor-f:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Default harbor backup schedule moved to after defaut velero schedule. This should ensure that images needed for resources in velero backup are present in harbor backup.

Also made the schedule configurable.

#### Checklist

- Change checks:
- Documentation checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -832,6 +832,9 @@ $defs:
         description: Value is the taint value the toleration matches to. If the operator is Exists, the value should be empty, otherwise just a regular string.
         type: string
     type: object
+  cronSchedule:
+    type: string
+    pattern: ^(((\*\/)?([0-5]?[0-9])((\,|\-|\/)([0-5]?[0-9]))*|\*)[^\S\r\n]+((\*\/)?((2[0-3]|1[0-9]|[0-9]|00))((\,|\-|\/)(2[0-3]|1[0-9]|[0-9]|00))*|\*)[^\S\r\n]+((\*\/)?([1-9]|[12][0-9]|3[01])((\,|\-|\/)([1-9]|[12][0-9]|3[01]))*|\*)[^\S\r\n]+((\*\/)?([1-9]|1[0-2])((\,|\-|\/)([1-9]|1[0-2]))*|\*|(jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec))[^\S\r\n]+((\*\/)?[0-6]((\,|\-|\/)[0-6])*|\*|00|(sun|mon|tue|wed|thu|fri|sat)))$|^@(annually|yearly|monthly|weekly|daily|hourly|reboot)$
 type: object
 required:
   - global
@@ -1698,8 +1701,9 @@ properties:
             description: |-
               `schedule` defines when the backup job for Harbor will run.
               This should be set to run shortly after Velero backups in the workload cluster, in order to ensure that images needed for Velero backups are backed up in Harbor.
+              Uses the Cron format, see https://en.wikipedia.org/wiki/Cron.
             default: "30 0 * * *"
-            type: string
+            $ref: '#/$defs/cronSchedule'
         type: object
       core:
         title: Core Config
@@ -1813,6 +1817,8 @@ properties:
             title: Schedule
             description: |-
               Defines a CRON schedule when the garbage collection job should run.
+              Uses a special Cron format that adds "seconds" as the first entry.
+              Order: "seconds, minutes, hours, day of month, month, day of week".
             default: 0 0 0 * * SUN
             type: string
         type: object
@@ -2318,9 +2324,11 @@ properties:
             default: false
           defaultSchedule:
             title: Rclone Sync Default Schedule
-            description: Default schedule to run the sync CronJobs.
-            type: string
+            description: |-
+              Default schedule to run the sync CronJobs.
+              Uses the Cron format, see https://en.wikipedia.org/wiki/Cron.
             default: 0 5 * * *
+            $ref: '#/$defs/cronSchedule'
           buckets:
             title: Rclone Sync Buckets
             description: Additional buckets to sync.
@@ -2343,7 +2351,10 @@ properties:
                 schedule:
                   type: string
                   title: Sync schedule for this bucket
-                  description: Defaults to `.objectStorage.sync.defaultSchedule`
+                  description: |-
+                    Defaults to `.objectStorage.sync.defaultSchedule`
+                    Uses the Cron format, see https://en.wikipedia.org/wiki/Cron.
+                  $ref: '#/$defs/cronSchedule'
                 sourceType:
                   type: string
                   title: Type of source
@@ -2565,8 +2576,11 @@ properties:
         default: false
       schedule:
         title: Velero Backup Schedule
+        description: |-
+          Schedule for Velero backup schedule "velero-daily-backup"
+          Uses the Cron format, see https://en.wikipedia.org/wiki/Cron.
         default: 0 0 * * *
-        type: string
+        $ref: '#/$defs/cronSchedule'
       retentionPeriod:
         title: Velero Backup Retention Period
         $ref: '#/$defs/goDuration'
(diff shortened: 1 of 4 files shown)


config/schemas/config.yaml
```diff
@@ -1693,6 +1693,13 @@ properties:
               `RetentionDays` defines how old a backup should be before deleting it.
             default: 7
             type: number
+          schedule:
+            title: Schedule for backup job
+            description: |-
+              `schedule` defines when the backup job for harbor will run.
+              This should be set to run shortly after velero backups in wc, in order to ensure that images needed for velero backups are backed up in harbor.
```
simonklb, line 1700:
```suggestion
              `schedule` defines when the backup job for Harbor will run.
              This should be set to run shortly after Velero backups in WC, in order to ensure that images needed for Velero backups are backed up in Harbor.
```

simonklb: Does this not also depend on how long time the Velero backup takes to complete?

It would be nice if they were triggered in sequence rather than on two different schedules. Issue worthy?

viktor-f: Yes, it does definitely depend on that. This mostly reflects that most velero backups should be done within 30 mins.
It would be very nice if that was triggered in sequence, but I'm not sure that it's worth making some controller that would be able to fix that. Especially since this is spread out across two different kubernetes clusters, harbor backups in sc and velero backups in wc

config/schemas/config.yaml
```diff
@@ -1693,6 +1693,13 @@ properties:
...
+            default: "30 0 * * *"
```
anders-elastisys, line 1701: It could be nice to have a pattern for this in the schema

anders-elastisys (approved):

config/schemas/config.yaml
simonklb: Since we are so hesitant to create new issues to prevent flooding the backlog (which I totally understand) could we find some way of still keep "nice to haves" around? I think even if this isn't something we want to add to the current backlog this _is_ something we want to solve in the future, i.e. that when you create a backup of your cluster all images that are currently in use should also be backed up.

viktor-f: i can bring that up to see what we can do :+1:

config/schemas/config.yaml
viktor-f: I did manage to find this pattern that seems to work in all general cases, but there might be some edge cases where it fails. But it is very complicated.

```
^((?<![\d\-\*])((\*\/)?([0-5]?[0-9])((\,|\-|\/)([0-5]?[0-9]))*|\*)[^\S\r\n]+((\*\/)?((2[0-3]|1[0-9]|[0-9]|00))((\,|\-|\/)(2[0-3]|1[0-9]|[0-9]|00))*|\*)[^\S\r\n]+((\*\/)?([1-9]|[12][0-9]|3[01])((\,|\-|\/)([1-9]|[12][0-9]|3[01]))*|\*)[^\S\r\n]+((\*\/)?([1-9]|1[0-2])((\,|\-|\/)([1-9]|1[0-2]))*|\*|(jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec))[^\S\r\n]+((\*\/)?[0-6]((\,|\-|\/)[0-6])*|\*|00|(sun|mon|tue|wed|thu|fri|sat)))$|^@(annually|yearly|monthly|weekly|daily|hourly|reboot)$
````

Is this something we want, or is it just confusing to add this?

Comment (simonklb) [reply config/schemas/config.yaml:1701]: The Kubernetes spec just has this:
```
        "schedule": {
          "description": "The schedule in Cron format, see https://en.wikipedia.org/wiki/Cron.",
          "type": "string"
        },
```

Comment (simonklb) [reply config/schemas/config.yaml:1701]: If we really want something this seems less complicated:
`/(@(annually|yearly|monthly|weekly|daily|hourly|reboot))|(@every (\d+(ns|us|µs|ms|s|m|h))+)|((((\d+,)+\d+|(\d+(\/|-)\d+)|\d+|\*) ?){5,7})/`
Source: https://stackoverflow.com/questions/14203122/create-a-regular-expression-for-cron-statement

Comment (viktor-f) [reply config/schemas/config.yaml:1702]: Added a link now to all the config options I could find that uses a cron format.

File: helmfile.d/values/harbor/harbor-backup.yaml.gotmpl [hunk 3]
```diff
@@ -7,6 +7,7 @@
 
 dbPassword: {{ .Values.harbor.internal.databasePassword }}
 retentionDays: {{ .Values.harbor.backup.retentionDays }}
+schedule: {{ .Values.harbor.backup.schedule }}
```
Zash, line 10: Consider passing through quote function, to be extra sure it stays a string.

config/schemas/config.yaml
Zash: Going with link only for now?

Zash (approved):

config/schemas/config.yaml
viktor-f: That's what I was thinking. But I'm open to the option of using the complicated regex as well.

helmfile.d/values/harbor/harbor-backup.yaml.gotmpl
viktor-f: Thanks, added now

config/schemas/config.yaml
OlleLarsson: I think I would like to have the regex to get some config validation before the k8s deploy stage. I don't really have a strong opinion on this though!

viktor-f: I have now added a regex pattern. I had to exclude "(?<![\d\-\*])" from the pattern I mentioned above because negative lookahead is not supported in the schema. But I think that is fine.

Here is some validation of the regex pattern that I used: https://regex101.com/r/l56Too/2

OlleLarsson (approved):

config/schemas/config.yaml
```diff
@@ -1693,6 +1693,15 @@ properties:
...
             type: number
+          schedule:
+            title: Schedule for backup job
+            description: |-
+              `schedule` defines when the backup job for harbor will run.
+              This should be set to run shortly after velero backups in wc, in order to ensure that images needed for velero backups are backed up in harbor.
+              Uses the Cron format, see https://en.wikipedia.org/wiki/Cron.
+            default: "30 0 * * *"
+            type: string
+            pattern: ^(((\*\/)?([0-5]?[0-9])((\,|\-|\/)([0-5]?[0-9]))*|\*)[^\S\r\n]+((\*\/)?((2[0-3]|1[0-9]|[0-9]|00))((\,|\-|\/)(2[0-3]|1[0-9]|[0-9]|00))*|\*)[^\S\r\n]+((\*\/)?([1-9]|[12][0-9]|3[01])((\,|\-|\/)([1-9]|[12][0-9]|3[01]))*|\*)[^\S\r\n]+((\*\/)?([1-9]|1[0-2])((\,|\-|\/)([1-9]|1[0-2]))*|\*|(jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec))[^\S\r\n]+((\*\/)?[0-6]((\,|\-|\/)[0-6])*|\*|00|(sun|mon|tue|wed|thu|fri|sat)))$|^@(annually|yearly|monthly|weekly|daily|hourly|reboot)$
```
aarnq, line 1704: This should probably be added as a definition in the root and reused.

aarnq (commented):
Just one thing else it looks good to me, also do update the capitalisation that Simon noted.

config/schemas/config.yaml
viktor-f: Moved this to a definition now. PTAL

aarnq (approved):
LGTM

PR #2314, opened 2024-10-16
Fix grafana datasources and tests for multi wc

Elias-elastisys:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Makes sure there is always a default datasource in grafana even with multiple WCs.
Fixes grafana datasource test to include all wc datasources.

#### Checklist

- Change checks:
- Documentation checks:
  - [x] The [public documentation](https://github.com/elastisys/compliantkubernetes) required no updates
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/helmfile.d/values/grafana/grafana-user.yaml.gotmpl
+++ b/helmfile.d/values/grafana/grafana-user.yaml.gotmpl
@@ -60,8 +60,12 @@ datasources:
       If only one workload cluster, don't bother naming the datasource and set to default.
       Otherwise set the name of the clusters so users know what they are looking at.
       */}}
-    {{- if eq (len .Values.global.clustersMonitoring) 1 }}
+    {{- with first .Values.global.clustersMonitoring }}
+    {{- if eq (len $.Values.global.clustersMonitoring) 1 }}
     - name: "Workload Cluster"
+    {{- else }}
+    - name: "Workload Cluster {{ . }}"
+    {{- end }}
       access: proxy
       basicAuth: false
       editable: false
@@ -71,11 +75,11 @@ datasources:
       url: http://grafana-label-enforcer:9090
       version: 1
       jsonData:
-        customQueryParameters: "tenant_id={{ first .Values.global.clustersMonitoring }}"
+        customQueryParameters: "tenant_id={{ . }}"
         manageAlerts: false
-    {{- else }}
-    {{- range .Values.global.clustersMonitoring }}
-    - name: "{{ . }}"
+    {{- end }}
+    {{- range rest .Values.global.clustersMonitoring }}
+    - name: "Workload Cluster {{ . }}"
       access: proxy
       basicAuth: false
       editable: false
@@ -87,7 +91,6 @@ datasources:
       jsonData:
         customQueryParameters: "tenant_id={{ . }}"
         manageAlerts: false
-    {{- end }}
     {{- end }}
     {{- end }}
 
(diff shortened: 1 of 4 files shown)


aarnq (approved):
Nice job!

helmfile.d/values/grafana/grafana-user.yaml.gotmpl
```diff
@@ -60,8 +60,8 @@ datasources:
       If only one workload cluster, don't bother naming the datasource and set to default.
       Otherwise set the name of the clusters so users know what they are looking at.
       */}}
-    {{- if eq (len .Values.global.clustersMonitoring) 1 }}
-    - name: "Workload Cluster"
+    {{- with first .Values.global.clustersMonitoring }}
+    - name: "Workload Cluster {{ . }}"
```
Xartos, line 64: For me I think this will look a bit strange for others using the cluster since this will be something like `Workload Cluster environment-name-wc` for basically all that have single WC.
In that case I'd probably be happier if we just used `{{ . }}` directly.

Also I think that we don't need the separate loops with this setup, they should be identical

aarnq: I'd argue that the user experience would be better with the "Workload Cluster", since we also expose the "Service Cluster" data source.
Additionally the split is mainly because we need to set one as default, else it will be wild west which Grafana will show first.

Xartos: Yea, I also think that that would be the best. If there's only one that is. So basically the same behaviour as we had it?

Right, I missed that one, so you're correct that we need them split

Elias-elastisys: I realized that this needs a migration script btw, grafana does not delete old data sources even if you remove them from the list

Xartos: > So should I try to rework it so that if the length is 1 that its just "Workload Cluster" or is it fine to have "Workload Cluster env-name" even with just 1 wc?

I think that would be nice if you could do (Keep "Workload Cluster" if it's only one)

AlbinB97 (approved):
👍

helmfile.d/values/grafana/grafana-user.yaml.gotmpl
Elias-elastisys: alright done, then those clusters won't need the migration step either

aarnq (approved):

anders-elastisys (approved):

tests/end-to-end/grafana/datasources.cy.js
```diff
@@ -109,10 +109,19 @@ describe("grafana dev datasources", function() {
...
 
   it("has workload cluster", function() {
-    cy.contains("Workload Cluster")
-      .should("exist")
-      .siblings()
-      .contains("default")
-      .should("exist")
+    cy.yqDigParse("sc", ".global.clustersMonitoring")
+      .then(([first_cluster, ...rest_clusters]) => {
+        cy.contains(`Workload Cluster ${first_cluster}`)
```
Xartos, line 114: Is this test working now? Doesn't this also need a check if it only is one WC or not?

Elias-elastisys: True, thought I tested but think i had the wrong naming still then. Fixed now, for some reason I also had to revert my change that added `parent()`. I tested this with both single and multi wc.

Xartos (approved):

PR #2343, opened 2024-11-14
New Dashboard: Simple Cluster Status Overview

### LINKED ISSUE #2325: Simple uptime and status dashboard [goto-monitoring]
### Proposed feature

Create a dashboard that gives a basic overview of the status of the cluster(s), answering questions like:

- [ ] Is the cluster running as it should?
- [ ] If not, what components might be having trouble?

### Proposed alternatives

- Tweak the 'Uptime and status' dashboard
- Tweak the 'Prometheus Blackbox Exporter' to provide status of Kubernetes itself

### Additional context

As a user, I would like to know if the cluster is up and running without having to hunt around many different dashboards.

### Definition of done

- [ ] I can look at a single dashboard and know if the cluster is okay.
- [ ] This new feature is covered by public documentation

Zash:
- apps: copy blackbox exporter dashboard as base for new simple status
- apps: rework cluster-status overview

> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Adds a new dashboard intended to provide a better *user-facing* overview of the overall state of the cluster.
It should answer the question "Is the cluster working?" and "Has there been recent problems?"
It's based on the Prometheus Blackbox Exporter Dashboard, with some panels at the top added and tweaked.

- Fixes #2325

#### Information to reviewers

- Does it tell you if the cluster is working?
- Does it tell you if the cluster is has had recent issues?
- Are there any other indicators

- Should this be merged back into the Prometheus Blackbox Exporter?

#### Checklist

- Change checks:
- Documentation checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:


helmfile.d/charts/grafana-dashboards/dashboards/cluster-status-dashboard.json
viktor-f, line 1: Question: I realized that this will show the probes that are present in the datasource one has selected. For user grafana this means that it will default to the probes in wc. That might be good and it is showing most of the important services in sc as well. But I wonder if it will be obvious to users that you could find more/other probes by switching to the service cluster datasource. 
Do you have any thoughts about this? Should there be a text box explaining this (and potentially other things about the dashboard)?

viktor-f (commented):
Thanks for adding this. I'm in contact with a user as well to get more feedback.

helmfile.d/charts/grafana-dashboards/dashboards/cluster-status-dashboard.json
viktor-f, line 1: Some user feedback (adding it as a file comment to make it a thread):

I took a look at the dashboard when I was running some database jobs last week. I think I came in with the wrong expectations, because the dashboard is clearly not targeting that use-case.

The latency-graphs for the systems that are in the dashboard seems really good for pinpointing/eliminating those systems as culprits for performance problems. The most immediate panel “in the same spirit”  that I felt was missing was S3. I don’t know how you probe so it’s possible a misbehaving storage solution could be seen through spiking times for Harbor or similar, but it would be better to have the status for the storage solution in the same way as for the other systems so one doesn’t have to infer the health of that.

A small thing is that the namespace selector at the top doesn’t seem to have any effect. I don’t know how hard it is to remove that and other UI elements that aren’t used, but I think that would make it slightly easier to use the dashboard.

Switching the dashboard to view 24 hours shows blue spikes (“resolve”) that are >1s for multiple systems. 1-2 seconds is not necessarily prohibitively long, but the blue part is frequently ~150ms, so I’m guessing some other service not in the dashboard is not responding in a timely manner for the spikes?

PR #2346, opened 2024-11-19
apps wc: exclude capi namespaces from hnc

lunkan93:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [x] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

The `hnc.excludedNamespaces` configuration option has been removed, you can use `hnc.excludedExtraNamespaces` if you need additional excluded namespaces.

...

- Fixes #

#### Checklist

- Change checks:
- Documentation checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -1739,37 +1739,16 @@ properties:
           - resource: networkpolicies
             group: networking.k8s.io
         type: array
-      excludedExtraNamespaces:
-        title: Excluded Extra Namespaces
-        description: |-
-          Extra excluded namespace, here we should add the namespaces that are more likely to change
-
-          [Including and excluding namespaces](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#including-and-excluding-namespaces-from-hnc)
-        items:
-          type: string
-        type: array
       excludedNamespaces:
         title: Excluded Namespaces
         description: |-
-          Excluded namespaces
+          Namespaces excluded by HNC, here you can configure a list of namespaces to exclude from HNC in addition to the default excluded namespaces.
+
+          [Including and excluding namespaces](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#including-and-excluding-namespaces-from-hnc)
         items:
           examples:
-            - alertmanager
-            - cert-manager
-            - falco
-            - fluentd
-            - fluentd-system
-            - gatekeeper-system
-            - hnc-system
-            - ingress-nginx
-            - kube-node-lease
-            - kube-public
-            - kube-system
-            - kured
-            - metallb-system
-            - monitoring
-            - rook-ceph
-            - velero
+            - mongodb-system
+            - sealed-secrets
           type: string
         type: array
       ha:
--- a/config/wc-config.yaml
+++ b/config/wc-config.yaml
@@ -172,27 +172,8 @@ hnc:
   ## Included namespaces, empty string will include all
   includedNamespacesRegex: ""
 
-  ## Excluded namespaces
-  excludedNamespaces:
-    - alertmanager
-    - cert-manager
-    - falco
-    - fluentd
-    - fluentd-system
-    - hnc-system
-    - ingress-nginx
-    - kube-node-lease
-    - kube-public
-    - kube-system
-    - kured
-    - monitoring
-    - rook-ceph
-    - velero
-    - gatekeeper-system
-    - metallb-system
-
   ## Extra excluded namespace, here we should add the namespaces that are more likely to change
-  excludedExtraNamespaces: []
+  excludedNamespaces: []
 
   ## Additional resources to enable opt-in propagation for.
   ## Objects that should be propagated must have one of the annotations listed here https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#limit-the-propagation-of-an-object-to-descendant-namespaces
--- a/helmfile.d/values/hnc/controller.yaml.gotmpl
+++ b/helmfile.d/values/hnc/controller.yaml.gotmpl
@@ -4,12 +4,30 @@ excludedNamespaces:
   {{- with .Values.hnc.excludedNamespaces }}
   {{- toYaml . | nindent 2 }}
   {{- end }}
-  {{- with .Values.hnc.excludedExtraNamespaces }}
-  {{- toYaml . | nindent 2 }}
-  {{- end }}
   {{- with (keys .Values.user.constraints) }}
   {{- toYaml . | nindent 2 }}
   {{- end }}
+  - alertmanager
+  - calico-apiserver
+  - calico-system
+  - cert-manager
+  - falco
+  - fluentd
+  - fluentd-system
+  - gatekeeper-system
+  - gpu-operator
+  - hnc-system
+  - ingress-nginx
+  - kube-node-lease
+  - kube-public
+  - kube-system
+  - kured
+  - metallb-system
+  - monitoring
+  - openstack-system
+  - rook-ceph
+  - tigera-operator
+  - velero
 
 unpropagatedAnnotations: {{- toYaml .Values.hnc.unpropagatedAnnotations | nindent 2 }}
 managedNamespaceAnnotations: {{- toYaml .Values.hnc.managedNamespaceAnnotations | nindent 2 }}
--- a/migration/v0.43/README.md
+++ b/migration/v0.43/README.md
@@ -82,6 +82,8 @@ As with all scripts in this repository `CK8S_CONFIG_PATH` is expected to be set.
     ./bin/ck8s upgrade wc v0.43 apply
     ```
 
+1. The config for `.hnc.excludedExtraNamespaces` has been moved to `.hnc.excludedNamespaces`. As such aliases used for `.hnc.excludedExtraNamespaces` in `$CK8S_CONFIG_PATH/wc-config.yaml` may be overwritten to the actual values, and need to be manually replaced with the alias again if desired.
+
 1. If Tekton is enabled, ensure to add appropriate network policies that allow traffic from Tekton to OpenSearch.
 
     To check if the tekton is enabled, run the following command
@@ -149,6 +151,11 @@ As with all scripts in this repository `CK8S_CONFIG_PATH` is expected to be set.
 
     ```bash
     ./migration/v0.43/prepare/10-capacity-alerts.sh
+
+1. Migrate `.hnc.excludedExtraNamespaces` to `.hnc.excludedNamespaces`:
+
+    ```bash
+    ./migration/v0.43/prepare/20-hnc-config.sh
     ```
 
 1. Update apps configuration:
--- a/migration/v0.43/prepare/20-hnc-config.sh
+++ b/migration/v0.43/prepare/20-hnc-config.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+HERE="$(dirname "$(readlink -f "${0}")")"
+ROOT="$(readlink -f "${HERE}/../../../")"
+
+# shellcheck source=scripts/migration/lib.sh
+source "${ROOT}/scripts/migration/lib.sh"
+
+if [[ "${CK8S_CLUSTER}" =~ ^(wc|both)$ ]]; then
+  log_info "operation on workload cluster"
+  if ! yq_null wc .hnc.excludedExtraNamespaces; then
+    if yq_null wc .hnc.excludedNamespaces; then
+      yq_move wc .hnc.excludedExtraNamespaces .hnc.excludedNamespaces
+    else
+      yq4 -i '.hnc.excludedNamespaces = ((.hnc.excludedExtraNamespaces | explode(.)) + (.hnc.excludedNamespaces | explode(.)))' "$CK8S_CONFIG_PATH"/wc-config.yaml
+      yq_remove wc .hnc.excludedExtraNamespaces
+    fi
+  fi
+fi


simonklb (commented):
Is there an equivalent to `openstack-system` in Azure clusters?

lunkan93:
> Is there an equivalent to `openstack-system` in Azure clusters?

I don't think so no, just had a look in a live environment.

simonklb (approved):

config/wc-config.yaml
viktor-f, line 1: Is it possible to move this to [config/k8s-installers/capi/](https://github.com/elastisys/compliantkubernetes-apps/tree/main/config/k8s-installers/capi)? That feels like it would be more fitting. Just make sure that the lists merge properly if you move it there.

lunkan93: Tested and it seems to work fine :+1:

config/k8s-installers/capi/wc-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - kube-system
+    - kured
+    - monitoring
+    - rook-ceph
+    - velero
+    - openstack-system
+    - gatekeeper-system
+    - gpu-operator
+    - metallb-system
+    - tigera-operator
```
robinAwallace, line 23: Could you try only adding the capi namespaces here and see if it merges correctly with default list? :slightly_smiling_face: 
```suggestion
    - openstack-system
    - gatekeeper-system
    - gpu-operator
    - metallb-system
    - tigera-operator
```

lunkan93: That was the first thing I tried and that did not work :(

viktor-f: Hmm, I dislike the fact that this is now duplicating information. But I understand that the other way didn't work.

Suggestion: Do we need to have this in the config? Could we perhaps instead move it to the values file and have an if-statement that adds the capi namespaces if capi=true? Then just use this config if we want to add something extra.

OlleLarsson (approved):

config/k8s-installers/capi/wc-config.yaml
viktor-f: Thanks, I like this solution better.

Suggestion: Could we perhaps move [this entire list](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/wc-config.yaml#L124-L140) to the same place? And just have the config option for extra namespaces that we might want to exclude. What do you think of that?

viktor-f: Hmm, I realize now that we also have the `excludedExtraNamespaces` config option. If we were to move the list of "standard" exclusions to the values file. Then perhaps we do not need both `excludedExtraNamespaces` and `excludedNamespaces`. I wonder if there are cases where we would rather have this in config so that it is possible to change for specific environments.

lunkan93: To me, it seems extremely unlikely that we would want to configure any of these namespaces to be hnc-managed. So to me it should suffice with only having the option to add namespaces with the `excludedExtraNamespaces`.

viktor-f: Maybe it is a bit out of scope for the original task. But I wonder if we should implement that change.
Although perhaps we should check with others before going ahead, to see if it is a good idea.
I'll @ the ones that have looked at this PR at least
@simonklb @Xartos @OlleLarsson @robinAwallace

simonklb: I'm always for less config and more streamlined defaults. :)

OlleLarsson: Let me know if I follow this correctly, you are suggesting moving [this](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/wc-config.yaml#L124) to [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile.d/values/hnc/controller.yaml.gotmpl#L3)?

viktor-f: @OlleLarsson yes, and removing the config option from your first link since we have another config option that can be used to exclude namespaces.

Xartos: Then I think it would make sense to also add this to public documentation, that these are "reserved" namespaces that the customer are not allowed to create. I'm mainly thinking of the namespaces `rook-ceph` and `metallb-system` that might not be part of the cluster, and also some kind of safeguard for customers to create these namespaces since they will not get the correct labels and stuff.

simonklb: Do we have no idea if the cluster is using Rook or MetalLB in Apps? I'm thinking that it would be fairly easy to make them conditional if we did (note I'm not asking for it to be introduced if it isn't already).

Xartos: If I remember correctly we have that for rook, but not metallb

OlleLarsson: Wouldn't it make sense to not conditional any namespaces based on k8s-installer or add-ons? That would make the transition of moving environments between k8s-installers or add-ons "easier"?

simonklb: Yea that is true, good thinking! While highly unlikely, a customer creating a namespace `metallb-system` would make it problematic for us to in the future start using that namespace. :smile: Let's just reserve all of them.

simonklb (changes requested):
@OlleLarsson convinced me that it's better to reserve all the possible namespaces that we might use to not have customers claim them and prevent us from migrating to different installers or addons.

Please do that instead of make it conditional or convince me why it's not a good idea! :smile:

lunkan93:
So I've been busy with other things, but just to recap:
- We want to add some Gatekeeper Policy as part of this, to prevent users from creating these reserved namespaces
- Update public docs with this information

Correct?

OlleLarsson:
> * We want to add some Gatekeeper Policy as part of this, to prevent users from creating these reserved namespaces

Can customers create subnamespaces even if they are part of the `excludedNamespaces` list?
If they can't then I don't think we need to add additional safeguard, expect making sure that they aren't created as a `userNamespace` as part of an application developer request.

lunkan93:
> > * We want to add some Gatekeeper Policy as part of this, to prevent users from creating these reserved namespaces
> 
> Can customers create subnamespaces even if they are part of the `excludedNamespaces` list? If they can't then I don't think we need to add additional safeguard, expect making sure that they aren't created as a `userNamespace` as part of an application developer request.

Yeah it would make sense if they couldn't, I'll verify that.

lunkan93:
Do we want to merge this? I probably won't have time to look at adding some safeguard in the near future, this was mostly just a quick fix for something I found while working at other stuff. I'm not 100% sure what it would like off the top of my head, but someone who has a clearer idea could create an issue to implement that.

OlleLarsson:
> Do we want to merge this? I probably won't have time to look at adding some safeguard for ourselves in the near future, this was mostly just a quick fix for something I found while working at other stuff. I'm not 100% sure what it would like off the top of my head, but someone who has a clearer idea could create an issue to implement that.

Make sense to merge as is. We could consider create a CI task for updating the HNC reserved namespaces.Let me know if you agree @viktor-f

helmfile.d/values/hnc/controller.yaml.gotmpl
```diff
@@ -1,15 +1,35 @@
...
+  - velero
+  - gatekeeper-system
+  - metallb-system
+  {{- if eq .Values.global.ck8sK8sInstaller "capi" }}
+  - calico-apiserver
+  - calico-system
+  - openstack-system
+  - gpu-operator
+  - tigera-operator
+  {{- end }}
```
simonklb, line 32: Was there a reason why we didn't just reserve these always?

```suggestion
  - calico-apiserver
  - calico-system
  - openstack-system
  - gpu-operator
  - tigera-operator
```

config/schemas/config.yaml
```diff
@@ -1515,34 +1515,13 @@ properties:
...
+          Extra namespaces excluded by HNC, here we should add the namespaces that are more likely to change
 
           [Including and excluding namespaces](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/docs/user-guide/how-to.md#including-and-excluding-namespaces-from-hnc)
-        items:
-          type: string
-        type: array
-      excludedNamespaces:
-        title: Excluded Namespaces
-        description: |-
-          Excluded namespaces
```
simonklb, line 1527: Probably not many are overriding this but since we are removing a configuration option that should probably be mentioned as a breaking change.

simonklb: Btw is it removed from the default config when you re-run `ck8s init`?

helmfile.d/values/hnc/controller.yaml.gotmpl
lunkan93: Right, I'll fix that :+1:

config/schemas/config.yaml
lunkan93: > Btw is it removed from the default config when you re-run `ck8s init`?

It should, but I can double check

lunkan93: Yes it is removed from default config

viktor-f:
> > Do we want to merge this? I probably won't have time to look at adding some safeguard for ourselves in the near future, this was mostly just a quick fix for something I found while working at other stuff. I'm not 100% sure what it would like off the top of my head, but someone who has a clearer idea could create an issue to implement that.
> 
> Make sense to merge as is. We could consider create a CI task for updating the HNC reserved namespaces.Let me know if you agree @viktor-f

I'm not sure I understand what the follow-up task is supposed to do. Could you explain it a bit more?

I'm ok with merging this. Since we are removing a config option I would have wanted a migration script that moved any existing use of `excludedNamespaces` to `excludedExtraNamespaces`. But you did at least add the admin note, so it could be ok without the migration script.

simonklb (approved):

robinAwallace (approved):

OlleLarsson:
> I'm not sure I understand what the follow-up task is supposed to do. Could you explain it a bit more?

Revisited this and I suppose that the main follow up would be to update the public docs with the list of reserved namespaces.

aarnq (commented):
I would see it as better choice to expose `excludedNamespaces` rather than `excludedExtraNamespaces` since that extra has lost its meaning.

viktor-f (approved):
Thanks for adding the text in the migration doc. That covers the last concern I have.

viktor-f:
> > I'm not sure I understand what the follow-up task is supposed to do. Could you explain it a bit more?
> 
> Revisited this and I suppose that the main follow up would be to update the public docs with the list of reserved namespaces.

Yes I think it would make sense to have that in the public docs. :+1:

lunkan93:
> I would see it as better choice to expose `excludedNamespaces` rather than `excludedExtraNamespaces` since that extra has lost its meaning.

While I do agree, I was looking at creating a migration script for this earlier and did not feel like it was worth the effort of trying to account for anchors/aliases without overwriting anything. We could of course make that a manual step too, but that step would have to be performed for almost all environments, not sure if that is preferable. One alternative could be to not make any changes to the exposed config right now and just add the namespaces, then we can have a follow-up to do the config changes.

lunkan93:
@aarnq Do you want to discuss this further or can this be merged as is?

aarnq:
> @aarnq Do you want to discuss this further or can this be merged as is?

Sorry it's been far down in my notifications, I'll review it later today.

helmfile.d/values/hnc/controller.yaml.gotmpl
```diff
@@ -1,15 +1,33 @@
...
+  - monitoring
+  - rook-ceph
+  - velero
+  - gatekeeper-system
+  - metallb-system
+  - calico-apiserver
+  - calico-system
+  - openstack-system
+  - gpu-operator
+  - tigera-operator
```
aarnq, line 30: Nit: sort

aarnq (commented):
> While I do agree, I was looking at creating a migration script for this earlier and did not feel like it was worth the effort of trying to account for anchors/aliases without overwriting anything. We could of course make that a manual step too, but that step would have to be performed for almost all environments, not sure if that is preferable. One alternative could be to not make any changes to the exposed config right now and just add the namespaces, then we can have a follow-up to do the config changes.

Are these config values typically aliased?
And shouldn't `.excludedNamespaces = explode(.excludedNamespaces) + explode(.excludedExtraNamespaces)` work?
With manual handling to restore potential aliases if so desired.

helmfile.d/values/hnc/controller.yaml.gotmpl
lunkan93: Yeah the excludedExtraNamespaces is aliased in most envs, but I will have a look at your suggestion.

aarnq: If `excludedNamespaces` is unset we could potentially just do a reassign, that should preserve the alias.

aarnq: Did it work well? As you have pushed the steps for it.

lunkan93: Yeah I did some testing with and without excludedNamespaces and both were merging correctly.

aarnq: OK, then I'm happy to accept it as is. :+1:

aarnq (approved):

PR #2361, opened 2024-12-02
Fixes for diagnostics script with namespace

anders-elastisys:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Noticed that running `diagnostics` with a `namespace` exited if no logs were found due to the exit code of `grep` and the script setting `set -euo pipefail`. This PR fixes this, and also adds a test for running `diagnostics namespace`.
Also added some additional patterns to `grep` logs for and simplified to get all configmaps in the namespace since we are already getting all pods in the namespace.

- Fixes #

#### Checklist

- Change checks:
- Documentation checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:
  - [x] The bug fix is covered by regression tests

### DIFF
--- a/bin/diagnostics.bash
+++ b/bin/diagnostics.bash
@@ -262,13 +262,13 @@ run_diagnostics_namespaced() {
     namespace="${1}"
     echo "Running in the ${namespace} namespace"
     # -- Pods --
-    echo -e "Fetching all pods <pods>"
+    echo -e "Fetching All Pods <pods>"
     "${here}/ops.bash" kubectl "${cluster}" get pods -n "${namespace}" -o yaml
     printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
 
     # -- Top --
     printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
-    echo "Fetching pods resources usage"
+    echo "Fetching Pods Resources Usage <top>"
     "${here}/ops.bash" kubectl "${cluster}" top pods -n "${namespace}"
     printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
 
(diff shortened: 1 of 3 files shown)


AlbinB97 (changes requested):
image
I tested it, and while normal diagnostics script seems to work fine, I get this error when trying to specify a namespace. Maybe I am doing it wrong? But the script says that `-n` should be a valid flag.

salehsedghpour:
I tested it, and while normal diagnostics script seems to work fine, I get this error when trying to specify a namespace. Maybe I am doing it wrong? But the script says that `-n` should be a valid flag.


I checked, to me it worked with using the keyword `namespace` instead of `-n`:
```bash
./bin/ck8s diagnostics wc namespace default
```

So we can either support `-n` or update the hint. I prefer the first one, as it aligns more with our experience with working with namespaces!

bin/diagnostics.bash
```diff
@@ -299,36 +299,18 @@ run_diagnostics_namespaced() {
...
+    "${here}/ops.bash" kubectl "${cluster}" get configmaps -n "${namespace}" -o yaml
     printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
 
     # -- Logs --
     printf '%*s\n' "${COLUMNS:-$(tput cols)}" '' | tr ' ' -
-    echo "Fetching Logs <logs>"
-    pods=$("${here}/ops.bash" kubectl "${cluster}" get pods -n "${namespace}" -o yaml | yq4 '.items[] | .metadata.name')
-    readarray pods_arr < <(echo "$pods" | yq4 e -o=j -I=0 '.[]')
-
+    echo "Fetching error Logs <logs>"
```
Zash, line 307: Consistent capitalization would be good.

bin/diagnostics.bash
```diff
@@ -299,36 +299,18 @@ run_diagnostics_namespaced() {
...
-    cfgs=($(echo "${cfgs[@]}" | tr [:space:] '\n' | awk '!a[$0]++'))
-
-    for cfg in "${cfgs[@]}"; do
-        if [ "$cfg" == "null" ]; then
-            :
-        else
-            configmap=$(sed -e 's/^"//' -e 's/"$//' <<<"$cfg")
-            "${here}/ops.bash" kubectl "${cluster}" get configmap -n "${namespace}" "$configmap" -o yaml
-        fi
-    done
```
Zash, line 320: Am I reading this correctly that it selects `configmaps` in active use?

Is that distinction not needed? If not then less code is better! 😄

Zash (commented):
A question and a nit.

Great to see comments explaining the why's :+1:

anders-elastisys:
@salehsedghpour @AlbinB97 PTAL, I changed it to only show the `--help` flag in the `ck8s` usage, since we also support other commands and flags not visible there, and adding them would make it to cluttered. https://github.com/elastisys/compliantkubernetes-apps/pull/2361/commits/15bd0f19d4ebe503b3e148cdc8eee6d573c91546

salehsedghpour (approved):

AlbinB97 (approved):
Nice work, thanks! 👍

bin/diagnostics.bash
anders-elastisys: It selected configmaps in active use for pods that are not `Ready`, which can have its uses, but I feel like we might be missing important information by not including ConfigMaps making it more difficult for whoever is going to look at the generated diagnostics to debug an issue. But I see merits with both approaches, as this change does generate more diagnostics, which might also be difficult to debug.

Elias-elastisys (approved):

bin/diagnostics.bash
Zash: Filtering locally may be faster than the round trip of asking for more information, so makes sense to err on the side of verbosity.

Zash (approved):

PR #2362, opened 2024-12-03
apps: use node-group label in node resource request alerts

### LINKED ISSUE #2349: [1] Change `{{ resource }}Request{{ percentage }}Percent{{ group }}` to use label instead of node name
### Description

Currently, [this alert](https://github.com/elastisys/compliantkubernetes-apps/blob/5f1e11f3fb012cf33f8ad7bb73c75b78a70fea82/helmfile.d/charts/prometheus-alerts/templates/alerts/cluster-capacity-management-alerts.yaml#L101) is filtering on node name instead of the new `label_elastisys_io_node_group` label that most metrics have and that other alerts in the same file are using.
We should change such that this alert also uses that label to select nodes instead of relying on node name.

### Additional context

_No response_

### Definition of done

- [ ] The linked alert uses the `label_elastisys_io_node_group` label to select nodes

lunkan93:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [x] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

The resource request capacity management alerts have been reworked to target the `elastisys.io/node-group` label instead of being based on node name patterns. As such, any previous override config for `prometheus.capacityManagementAlerts.requestLimit` will be invalid and needs to be reconfigured or removed to use the new defaults. 

Update the node resource request capacity alerts to use the `elastisys.io/node-group` label instead of relying on the node names.

- Fixes #2349 

#### Checklist

- Change checks:
  - [x] The change updates the config *and* the schema
- Documentation checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/common-config.yaml
+++ b/config/common-config.yaml
@@ -415,21 +415,10 @@ prometheus:
     nodeCpuLimit1h: 95
     nodeMemoryLimit1h: 85
     ## for each cpu and memory add the node pattern for which you want to create an alert
-    requestlimit:
-      cpu:
-        - name: worker
-          pattern: ".*-worker-.*"
-          limit: 80
-        - name: elastisys
-          pattern: ".*-elastisys-.*"
-          limit: 80
-      memory:
-        - name: worker
-          pattern: ".*-worker-.*"
-          limit: 80
-        - name: elastisys
-          pattern: ".*-elastisys-.*"
-          limit: 80
+    requestLimit:
+      cpu: 80
+      memory: 80
+    nodeGroupRequestsExcludePattern: ""
   diskAlerts:
     storage:
       predictLinear:
--- a/migration/v0.43/prepare/10-capacity-alerts.sh
+++ b/migration/v0.43/prepare/10-capacity-alerts.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+HERE="$(dirname "$(readlink -f "${0}")")"
+ROOT="$(readlink -f "${HERE}/../../../")"
+
+# shellcheck source=scripts/migration/lib.sh
+source "${ROOT}/scripts/migration/lib.sh"
+
+if [[ "${CK8S_CLUSTER}" =~ ^(sc|both)$ ]]; then
+  log_info "operation on service cluster"
+  yq_remove sc .prometheus.capacityManagementAlerts.requestLimit
+fi
+if [[ "${CK8S_CLUSTER}" =~ ^(wc|both)$ ]]; then
+  log_info "operation on workload cluster"
+  yq_remove wc .prometheus.capacityManagementAlerts.requestLimit
+fi
+
+yq_remove common .prometheus.capacityManagementAlerts.requestLimit
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -4076,7 +4076,7 @@ properties:
             description: Alert when a disk's usage reaches the limit in percent.
             type: number
             default: 75
-          requestlimit:
+          requestLimit:
             title: Capacity Management Alerts Request Limit
             description: Alert when a node's resource requests reaches the limits in percent.
             type: object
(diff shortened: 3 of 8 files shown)


config/common-config.yaml
```diff
@@ -418,17 +418,13 @@ prometheus:
...
         - name: elastisys
-          pattern: ".*-elastisys-.*"
           limit: 80
       memory:
         - name: worker
-          pattern: ".*-worker-.*"
           limit: 80
         - name: elastisys
-          pattern: ".*-elastisys-.*"
           limit: 80
```
viktor-f, line 428: If I understand correctly, this list determines which node groups we will look at. 
I wonder if it would be better to just look at all node groups that are available. The thing I dislike about this is that we would have to remember to add a new entry to this list if we create a new node group. I suspect that it will be rather common for wc workers to be split into several different groups.
But I also see that we might not want these alerts on e.g. AMS nodes or control plane nodes.

What do you think about this? 

Note that I do think your changes fulfill the task very well. So this might be a bit out of scope.

lunkan93: Yeah that is correct. I also had the same thoughts about this list, but opted against changing the behavior for the same reason you brought up. Not sure what is preferable :thinking:

Xartos: I'm not following the reason for why we wouldn't want this? Isn't the purpose of the `node-group` label specifically for these alerts, so then we shouldn't have multiple different if we don't want it. We have the `node-type` that we should use for scheduling purposes if that's necessary

And regarding us not wanting AMSes, couldn't we add a "excludeList" here. So that we can specify what node groups we don't want to get alerts for, alternatively we always ignore empty `node-group` and just set it to that if we don't want it.

But I might've misunderstood something completely here

lunkan93: I think what I would prefer the most is just, if you want capacity alerts for a node group you label it with `node-group`, since that is the only purpose of the label right now, and if not, don't label them. My hesitation was mostly just changing the behavior from only looking at worker and elastisys nodes, but maybe there wasn't a particular reason for that in the first place. So I'm fine with just making this look at any nodes with the `node-group` label, like the other capacity alerts. Then later down the line, if we find a need for picking and choosing which particular alerts are enabled for which `node-groups`, then maybe we can add some `excludeList`. Thoughts?

viktor-f: The label was originally made for the usage alerts, not these request alerts. They are very similar, but different enough that I thought it was worth discussing.

An exclude list could be a good way to work around this if we wanted to. I think that we should also always ignore empty `node-group`, but I think that there might be some situations where we might want to alert on high usage but not high requests (such as the AMS nodes that have very stable requests that we might not need alerts for, but usage alerts would still be good).

Xartos: Right, I was confusing this with the usage alerts and I agree that they are different. And in that case I would probably prefer to, as we said, go with all `node-group` and add an exclude list. I guess the only issue in that case is that we can't have only the resource alert and not the capacity alert. But I don't see any situation where we wouldn't want that.

lunkan93: Question: we will now lose the ability to configure the limit for when to alert for each node group, is that also fine? @viktor-f @Xartos

lunkan93:
Pushed some changes, will fix config and schema too if this is what we want.

Xartos (approved):
LGTM!
And it's a bit annoying maybe, but I don't think it's a big deal that we can't set the limit per node group. In a perfect world we want to have them the same for all

config/common-config.yaml
viktor-f: For now I think it can be ok to lose the ability to configure limit per group.
I also think that for requests we only need alert per group, not individual nodes.

config/common-config.yaml
viktor-f, line 1: (Adding a comment on this file to get a thread)

viktor-f (commented):
I'm happy with this solution. Well done :slightly_smiling_face:

anders-elastisys (approved):

config/common-config.yaml
lunkan93: Yeah I agree, not sure if there is a great way to translate previous config in a script for this as the behavior of these alerts and their config is changed. Perhaps adding a platform admin notice, a migration step and a script that removes any current override config so that it doesn't override the new defaults and break things?

viktor-f: Yes that should work

lunkan93: Added a platform admin notice and migration script now

migration/v0.43/prepare/10-capacity-alerts.sh
viktor-f, line 1: This should be mentioned in the list of manual steps in the migration doc. Just a step that tells you to run this script.

viktor-f (approved):

PR #2364, opened 2024-12-05
Improve readme steps for installation of apps

Elias-elastisys:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [x] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Tries to improve the readme so that it is possible to install apps solely based on it.

Decided in https://github.com/elastisys/ck8s-arch/issues/215

#### Checklist

- Change checks:
- Documentation checks:
  - [x] The [public documentation](https://github.com/elastisys/compliantkubernetes) required no updates
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,19 @@ This repository installs all the applications of ck8s on top of already created 
 To setup the clusters see [compliantkubernetes-kubespray][compliantkubernetes-kubespray].
 A service-cluster (sc) or workload-cluster (wc) can be created separately but all of the applications will not work correctly unless both are running.
 
-All config files will be located under `CK8S_CONFIG_PATH`.
+We follow the "configuration as code" principle which means that all configuration necessary to configure and operate the platform resides in the `CK8S_CONFIG_PATH` directory.
 There will be four config files: `common-config.yaml`, `wc-config.yaml`, `sc-config.yaml` and `secrets.yaml`.
+
+We strongly suggest to make your config directory part of a git repository so that it is stored safely and to allow you to rollback previously applied changes.
+We additionally suggest that you make Apps a submodule of your config repository in order to properly track which version of Apps you have applied and make your config repository the single source of truth of your environment.
+
+All operations are done through the `./bin/ck8s` command line tool. Run `./bin/ck8s help` for a complete set of possible commands.
+
+For more information please read our public documentation:
+
+- [Understand the basics](https://elastisys.io/welkin/operator-manual/understand-the-basics/)
+- [Understand Welkin](https://elastisys.io/welkin/operator-manual/understand-welkin/)
+
 See [Quickstart](#quickstart) for instructions on how to initialize the repo
 
 ### :cloud: Cloud providers :cloud:
@@ -95,27 +106,11 @@ If this is all new to you, here's a [link](https://riseup.net/en/security/messag
 
 ### Quickstart
 
-**You probably want to check the [compliantkubernetes-kubespray][compliantkubernetes-kubespray] repository first, since compliantkubernetes-apps depends on having two clusters already set up.**
-In addition to this, you will need to set up the following DNS entries (replace `example.com` with your domain).
+> [!NOTE]
+> **You probably want to check the [compliantkubernetes-kubespray][compliantkubernetes-kubespray] repository first, since compliantkubernetes-apps depends on having two clusters already set up.**
 
-There are two options when managing DNS records, manually or ExternalDNS.
-
-- Manually point these domains to the workload cluster ingress controller:
-
-    - `*.example.com`
-
-- Manually point these domains to the service cluster ingress controller:
-
-    - `*.ops.example.com`
-    - `dex.example.com`
-    - `grafana.example.com`
-    - `harbor.example.com`
-    - `opensearch.example.com`
-
-Assuming you already have everything needed to install the apps, this is what you need to do.
-
-The other option is to let ExternalDNS manage your DNS records, currently only AWS Route 53 is supported.
-You configure ExternalDNS later in the process.
+> [!NOTE]
+> Depending on your configuration of the clusters and OIDC, you might not have access to workload cluster before installing Dex in the service cluster. You would then have to install Apps in each cluster separately, starting with the service cluster.
 
 1. Decide on a name for this environment, the cloud provider to use as well as the flavor and set them as environment variables:
     Note that these will be later kept as global values in the common defaults config to prevent them from being inadvertently changed, as they will affect the default options of the configuration when generated or updated.
@@ -132,10 +127,12 @@ You configure ExternalDNS later in the process.
     export CK8S_K8S_INSTALLER=[kubespray|capi] # set this to whichever installer was used for the kubernetes layer
     ```
 
-    > [!NOTE]
-    > The `air-gapped` flavor has a lot of the same settings as the `prod` flavor but with some additional variables that you need to configure yourself (these are set to `set-me`).
+> [!NOTE]
+> The `air-gapped` flavor has a lot of the same settings as the `prod` flavor but with some additional variables that you need to configure yourself (these are set to `set-me`).
 
-1. Then set the path to where the ck8s configuration should be stored and the PGP fingerprint of the key(s) to use for encryption:
+<!-- markdownlint-disable MD029 -->
+
+2. Then set the path to where the ck8s configuration should be stored and the PGP fingerprint of the key(s) to use for encryption:
 
     ```bash
     export CK8S_CONFIG_PATH=${HOME}/.ck8s/my-ck8s-cluster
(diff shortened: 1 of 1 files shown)


bin/update-ips.bash
```diff
@@ -161,8 +161,8 @@ get_dns_ips() {
   local -a ips
   mapfile -t ips < <(dig +short "${domain}" | grep '^[.0-9]*$')
   if [ ${#ips[@]} -eq 0 ]; then
-    log_error "No IPs for ${domain} was found"
-    exit 1
+    log_error "No IPs for ${domain} was found. Will block all IPs"
+    echo "0.0.0.0"
```
simonklb, line 165: Will this not the opposite, i.e. allow all IPs? This function is used to resolve a domain name to an IP so that it can be added to the allow list, no?

simonklb (changes requested):
No issues found in the README but the change in the `update-ips` script looks dangerous. :o

bin/update-ips.bash
Elias-elastisys: This eventually sets the CIDR in the NP to 0.0.0.0/32, i.e. just the 0.0.0.0 ip which would never be a real IP meaning it blocks everything.

Elias-elastisys: The `process_ips_to_cidrs` seems to be the one adding the `/32`

simonklb: Ah of course :facepalm: 

I'm still a bit worried that we might at some point rewrite this and not all IPs go through that same IP to CIDR transformation function.

Also, why would you want it not to fail if the domain can't be resolved? Is it not better to become aware of it?

Elias-elastisys: So, when you use a LoadBalancer Service for ingress-nginx you can't create the domains before installing ingress-nginx. And before you could not run update-ips since it would fail and the schema validation would complain if you tried to apply anything. This seemed like a better approach then having to manually set all NPs to something first. Not perfect but I'm not sure if there is a better solution.

Although I just realized I forgot to mention you would have to re-run update-ips again after setting up the DNS.

aarnq: Could it not be left emptied instead to disallow everything that way? :thinking:
(IIRC the associated rules _shouldn't_ be rendered in that case, because empty netpols allow everything, but we have a split between the old and new netpol still.)

README.md
```diff
@@ -114,6 +114,21 @@ There are two options when managing DNS records, manually or ExternalDNS.
 
 Assuming you already have everything needed to install the apps, this is what you need to do.
 
+> [!NOTE]
```
cristiklein, line 117: I'm missing a section giving an overview of Apps. This is stuff that we already know and have in our DNA, which is why it's so hard to write it. Someone completely new to the project should first learn that:

- Welkin follows the "configuration as code" principle and expects all configuration to be in a so-called configuration folder.
- Welkin strongly recommends to have this configuration folder in `.git`. Apps may be a submodule, so that the configuration folder is the source of truth for how an environment is set up.
- Welkin Apps' functionality is accessed threw the `./bin/ck8s` command-line tool.

Makes sense?

cristiklein (changes requested):
Overall, I like the improvements, but I'm missing a section which discusses "the forest" before discussing "the trees".

bin/update-ips.bash
simonklb: Maybe split the script change and the documentation changes into separate PRs, both good for changelog purposes and review purposes

README.md
```diff
@@ -132,10 +147,10 @@ You configure ExternalDNS later in the process.
     export CK8S_K8S_INSTALLER=[kubespray|capi] # set this to whichever installer was used for the kubernetes layer
     ```
 
-    > [!NOTE]
-    > The `air-gapped` flavor has a lot of the same settings as the `prod` flavor but with some additional variables that you need to configure yourself (these are set to `set-me`).
+> [!NOTE]
+> The `air-gapped` flavor has a lot of the same settings as the `prod` flavor but with some additional variables that you need to configure yourself (these are set to `set-me`).
 
-1. Then set the path to where the ck8s configuration should be stored and the PGP fingerprint of the key(s) to use for encryption:
+2. Then set the path to where the ck8s configuration should be stored and the PGP fingerprint of the key(s) to use for encryption:
```
lucianvlad, line 153: Can you please clarify why we are switching from automatic numbering to manual numbering and then back again to automatic?
It is a bit confusing and I would like to know the reason for my learning purpose.

README.md
```diff
@@ -205,6 +225,33 @@ You configure ExternalDNS later in the process.
     )
     ```
 
+1. Update Network Policies
+
+    ```bash
+    ./bin/ck8s update-ips both apply
```
lucianvlad, line 231: Could we first run the `dry-run` so we see if there are any IP's or ports being removed that are not supposed to be removed?
I am thinking of the many cases where the `update-ips` script removed object storage ports or CIDR subnets that affected the environment afterwards.

lucianvlad (changes requested):

README.md
Elias-elastisys: In order for the notes to actually render correctly they can't be indented and when I moved them back the automatic numbering broke.

bin/update-ips.bash
Elias-elastisys: > Could it not be left emptied instead to disallow everything that way? 🤔

If I leave it empty instead the CIDR ends up as just `/32` which fails in the apply

README.md
Elias-elastisys: Pushed a change. Is that satisfactory? Hard to know where to exactly place it.

README.md
```diff
@@ -114,6 +120,21 @@ There are two options when managing DNS records, manually or ExternalDNS.
 
 Assuming you already have everything needed to install the apps, this is what you need to do.
 
+> [!NOTE]
+> Depending on your infrastructure, you might utilize a Service of type LoadBalancer for the ingress controller. This means you will not have an IP for the domains before installing the ingress controller. After configuring and validating the config, you can install just the ingress controller before the rest of apps with the following command
+>
+> ```bash
+> ./bin/ck8s ops helmfile sc apply -lapp=ingress-nginx --include-transitive-needs
+> ./bin/ck8s ops helmfile wc apply -lapp=ingress-nginx --include-transitive-needs
```
anders-elastisys, line 128: I am wondering if this README somewhere and somehow should state that you might not be able to run commands for `wc` if you have not installed Dex in `sc` first, since we usually assume for Welkin that `wc` is configured to point to Dex in `sc` :thinking:

anders-elastisys (commented):
~We should maybe include that the nodes should be labelled with the `elastisys.io/node-group` based on what group it belongs to (e.g. `worker` or `control-plane`) for our node capacity alerts to work properly~
I missed that we already had this section: https://github.com/elastisys/compliantkubernetes-apps?tab=readme-ov-file#capacity-management

README.md
```diff
@@ -114,6 +120,21 @@ There are two options when managing DNS records, manually or ExternalDNS.
 
 Assuming you already have everything needed to install the apps, this is what you need to do.
 
+> [!NOTE]
```
aarnq, line 123: It feels like we need to rethink the structure here.
Because first we tell them to setup domains to something, which in the normal case (with infrastructure providers supporting LB SVC) won't work.
Then second we tell them to setup ingress controller without first setting up the environment directory, which you have to remember to go back to.

I feel like we need to better identify the happy path, and make the structure better at being referable should we need to jump somewhere.

README.md
Elias-elastisys: Added links to those

README.md
Elias-elastisys: Tried to move the DNS steps to a later stage, is that better?

README.md
Elias-elastisys: Added a note about that, hard to find a good place to put it.

OlleLarsson:
Nice work!

As the instructions now are longer than before I'm tempted to suggest moving the Setup steps to `docs/`. Let me know what you think of that.

README.md
```diff
@@ -96,26 +107,9 @@ If this is all new to you, here's a [link](https://riseup.net/en/security/messag
 ### Quickstart
 
 **You probably want to check the [compliantkubernetes-kubespray][compliantkubernetes-kubespray] repository first, since compliantkubernetes-apps depends on having two clusters already set up.**
```
anders-elastisys, line 109: Nit: IMO this would look cleaner if we put it in a GitHub alert, either merge it with the one after this or as its own:
```suggestion
> [!NOTE]
> You probably want to check the [compliantkubernetes-kubespray][compliantkubernetes-kubespray] repository first, since compliantkubernetes-apps depends on having two clusters already set up.
```

anders-elastisys (commented):
I think the current changes looks good, just have some nits.

There is a link to compliantkubernetes.io in the README, could you update it to https://elastisys.io/welkin (we apparently don't have welkin.io :cry:) as part of this PR?

README.md
Elias-elastisys: :+1: done

anders-elastisys (approved):
LGTM, feel like you have made major improvements and that we can always continuously improve on this more in the future

lucianvlad (approved):

Elias-elastisys:
> As the instructions now are longer than before I'm tempted to suggest moving the Setup steps to `docs/`. Let me know what you think of that.

I agree that it's a bit long but also feel like I prefer having docs in the main readme, especially Setup instructions. If anything, the additional config could maybe be moved? (but I'd also be very happy to finally merge this) @OlleLarsson

Elias-elastisys:
@cristiklein Do you still have changes to this you'd want?

cristiklein:
> @cristiklein Do you still have changes to this you'd want?

I'm all good, thanks for checking.

cristiklein (approved):
No comments from my end. Thanks for brushing this up.

PR #2367, opened 2024-12-11
Add extra Application Developer RBAC config

anders-elastisys:
> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

As an Platform Administrator, I want a declarative approach for providing additional RBAC to Application Developers without extending the current RBAC provided by Welkin. This can be used for special cases where an application developer might request a service that requires some additional RBAC that they themselves can't delegate.

This also fixes so that  the `user-crds` ClusterRoleBinding is not created unless any userCRD resources have been specified, so now it will be the same check as for the [`user-crds` ClusterRole](https://github.com/elastisys/compliantkubernetes-apps/blob/v0.42.0/helmfile.d/charts/user-rbac/templates/clusterroles/user-crds.yaml#L1).

- Fixes #

#### Checklist

- Change checks:
- Documentation checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -832,6 +832,91 @@ $defs:
         description: Value is the taint value the toleration matches to. If the operator is Exists, the value should be empty, otherwise just a regular string.
         type: string
     type: object
+  io.k8s.api.rbac.v1.Subject:
+    title: Subject
+    description: Subject contains a reference to the object or user identities a role
+      binding applies to.  This can either hold a direct API object reference, or a
+      value for non-objects such as user and group names.
+    properties:
+      apiGroup:
+        title: Kind
+        description: APIGroup holds the API group of the referenced subject. Defaults
+          to "" for ServiceAccount subjects. Defaults to "rbac.authorization.k8s.io"
+          for User and Group subjects.
+        type: string
+      kind:
+        title: Kind
+        description: Kind of object being referenced. Values defined by this API group
+          are "User", "Group", and "ServiceAccount". If the Authorizer does not recognized
+          the kind value, the Authorizer should report an error.
+        type: string
+        enum:
+          - Group
+          - ServiceAccount
+          - User
+      name:
+        title: Name
+        description: Name of the object being referenced.
+        type: string
+      namespace:
+        title: Namespace
+        description: Namespace of the referenced object.  If the object kind is non-namespace,
+          such as "User" or "Group", and this value is not empty the Authorizer should
+          report an error.
+        type: string
+    required:
+    - kind
+    - name
+    type: object
+  io.k8s.api.rbac.v1.PolicyRule:
+    title: PolicyRule
+    description: PolicyRule holds information that describes a policy rule, but does
+      not contain information about who the rule applies to or which namespace the rule
+      applies to.
+    properties:
+      apiGroups:
+        title: APIGroups
+        description: APIGroups is the name of the APIGroup that contains the resources.  If
+          multiple API groups are specified, any action requested against one of the
+          enumerated resources in any API group will be allowed.
+        items:
+          type: string
+        type: array
+      nonResourceURLs:
+        title: NonResourceURLs
+        description: NonResourceURLs is a set of partial urls that a user should have
+          access to.  *s are allowed, but only as the full, final step in the path Since
+          non-resource URLs are not namespaced, this field is only applicable for ClusterRoles
+          referenced from a ClusterRoleBinding. Rules can either apply to API resources
+          (such as "pods" or "secrets") or non-resource URL paths (such as "/api"),  but
+          not both.
+        items:
+          type: string
+        type: array
+      resourceNames:
+        title: ResourceNames
+        description: ResourceNames is an optional white list of names that the rule
+          applies to.  An empty set means that everything is allowed.
+        items:
+          type: string
+        type: array
+      resources:
+        title: Resources
+        description: Resources is a list of resources this rule applies to.  ResourceAll
+          represents all resources.
+        items:
+          type: string
+        type: array
+      verbs:
+        title: Verbs
+        description: Verbs is a list of Verbs that apply to ALL the ResourceKinds and
+          AttributeRestrictions contained in this rule.  VerbAll represents all kinds.
+        items:
+          type: string
+        type: array
+    required:
+    - verbs
+    type: object
   cronSchedule:
     type: string
     pattern: ^(((\*\/)?([0-5]?[0-9])((\,|\-|\/)([0-5]?[0-9]))*|\*)[^\S\r\n]+((\*\/)?((2[0-3]|1[0-9]|[0-9]|00))((\,|\-|\/)(2[0-3]|1[0-9]|[0-9]|00))*|\*)[^\S\r\n]+((\*\/)?([1-9]|[12][0-9]|3[01])((\,|\-|\/)([1-9]|[12][0-9]|3[01]))*|\*)[^\S\r\n]+((\*\/)?([1-9]|1[0-2])((\,|\-|\/)([1-9]|1[0-2]))*|\*|(jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec))[^\S\r\n]+((\*\/)?[0-6]((\,|\-|\/)[0-6])*|\*|00|(sun|mon|tue|wed|thu|fri|sat)))$|^@(annually|yearly|monthly|weekly|daily|hourly|reboot)$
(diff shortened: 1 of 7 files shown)


helmfile.d/values/userCRDs/common.yaml.gotmpl
viktor-f, line 1: I think that this is a reasonable feature to add. :+1: 
But I'm not sure if userCRDs is the best chart to add it with. My initial expectation when thinking about it would be to add it via the userRBAC chart. I also think that it is a bit unexpected that these extra roles can only be added if [gatekeeper.allowUserCRDs.enabled=true](https://github.com/elastisys/compliantkubernetes-apps/blob/6cbd652bd9abe355ecace8ddf859ce793d9330f4/helmfile.d/stacks/rbac.yaml#L44)
But I understand that it was very convenient to add it via this chart, so I'm not sure if my concerns are worth acting on. What do you think of this?
Another alternative could be to add a new chart, called something like "userCustomRBAC".

anders-elastisys: Yes I do agree, but I feel like it is nice to use the RBAC templating of the `userCRDs` chart, personally I think it is the name for the `user-crds` chart that is a bit misleading as it is essentially just a chart for generating RBAC. I got a bit carried away in [this PR](https://github.com/elastisys/compliantkubernetes-apps/pull/2354) to try to address this and change some of the values file and charts to make the configuration of user CRDs a bit more clear, but I have not had the time to go back to it. I guess this would be something to maybe discuss and bring up with @elastisys/goto-access-control

viktor-f: Yes, I agree that the name for the chart is a bit misleading and that we could probably restructure it significantly.
I think I missed that PR earlier, that kind of rework could be nice. But perhaps a bit out of scope and something that could be handed over /discussed with access control goto like you said.
Since you are trying to add this I assume that this is something you have need for. Is that correct? (You don't need to/should mention any details here).

I think I'd be ok with making these changes in the simplest way possible now and then reworking it later. The only thing that I might see as a true blocker is the oddity of locking this behind `gatekeeper.allowUserCRDs.enabled=true`.

anders-elastisys: How about having it as a separate chart? https://github.com/elastisys/compliantkubernetes-apps/pull/2367/commits/450a7d138790ba9d6a2510ffd6e46ae472386d0a
This removes the dependency on `gatekeeper.allowUserCRDs.enabled=true`.
And then addressing the name of the chart can be a separate PR, and yes this templating is something we are using atm for cases where we need to add e.g. a single ClusterRoleBinding for an app developer without having to add it to our templating directly.

helmfile.d/stacks/rbac.yaml
```diff
@@ -51,7 +51,21 @@ templates:
...
       - values/userCRDs/flux/fluxv2.yaml.gotmpl
+      - values/userCRDs/mongodbcommunity/mongodb.yaml.gotmpl
+      - values/userCRDs/strimzi/kafka.yaml.gotmpl
+
+  extra-dev-rbac:
+    inherit: [ template: rbac ]
+    condition: ck8sWorkloadCluster.enabled
+    chart: charts/user-crds
+    version: 1.0.0
+    name: extra-dev-rbac
```
aarnq, line 65: Nit: I would prefer `dev-rbac-extra` as name so they would be naturally sorted together.

config/schemas/config.yaml
```diff
@@ -1473,6 +1473,18 @@ properties:
...
+        type: object
+      extraRoleBindings:
+        title: Configure extra RoleBindings for application developers
+        type: object
+      extraClusterRoles:
+        title: Configure extra ClusterRoles for application developers
+        type: object
+      extraClusterRoleBindings:
+        title: Configure extra ClusterRoleBindings for application developers
+        type: object
```
aarnq, line 1487: Describe how these are supposed to be use, because right now one would have to look at the values templates for user-crds to see how it is setup.
And that the resulting roles and rolebindings are in all application developer namespaces.

helmfile.d/values/userCRDs/common.yaml.gotmpl
viktor-f: Yes, I like this version better.
I'm happy with this after you fix the comments from aarnq.

anders-elastisys:
@aarnq PTAL https://github.com/elastisys/compliantkubernetes-apps/pull/2367/commits/05cf393c96698ff9294556d86cd3d5b5ae6b4f67

config/schemas/config.yaml
```diff
@@ -1473,6 +1473,84 @@ properties:
...
+          The Roles are added to all application developer namespaces configured in user.namespaces
+        type: object
+        properties: {}
+        additionalProperties:
+          properties:
+            rules:
+              title: PolicyRules for this Role
+              type: array
+              default: []
+          additionalProperties: false
```
viktor-f, line 1489: Question: What is the reason for adding this in `additionalProperties` instead of `properties`? AFAIK the common thing would be to add all options in `properties`. But I'm no expert on the config schema yet, so I might just be missing something. 
Same question for the other things you added further down.

config/schemas/config.yaml
```diff
@@ -1473,6 +1473,84 @@ properties:
...
+          Configure extra Roles for application developers
+          The Roles are added to all application developer namespaces configured in user.namespaces
+        type: object
+        properties: {}
+        additionalProperties:
+          properties:
+            rules:
+              title: PolicyRules for this Role
+              type: array
+              default: []
```
viktor-f, line 1488: The items in this list is just the same rules format that is added to roles, right? Could we perhaps link to some kubernetes documentation to make it very clear what to put here?
Same for the other similar things further down.

config/schemas/config.yaml
anders-elastisys: Since what you configure are objects with "variable/custom" names I did this way as it is also how we do it for configuring constraints, see [here](https://github.com/elastisys/compliantkubernetes-apps/blob/v0.42.0/config/schemas/config.yaml#L1336-L1354). But maybe there is a better way to handle it? Maybe @aarnq has an opinion on this?

config/schemas/config.yaml
```diff
@@ -1473,6 +1473,99 @@ properties:
             title: Enable Kafka
             type: boolean
         additionalProperties: false
+      extraRoles:
+        title: Configure extra Roles for application developers
```
aarnq, line 1477: Prefer to keep the title short like "Extra Roles", as that will be generated as "Extra Roles Schema" in the config, or "Extra Application Developer Roles" if you want to retain that in the title. It is fine with me if that is in the description instead.

Similar for the other titles.

config/schemas/config.yaml
aarnq: Yes, this is the way to express `map[string]object`, as the key will become the name of the resource.

config/schemas/config.yaml
anders-elastisys: Like this https://github.com/elastisys/compliantkubernetes-apps/pull/2367/commits/6ccbd8e6dd8136858ffcaa6d0009853c7aee0372?

aarnq: Looks good, well done :+1:

config/wc-config.yaml
```diff
@@ -73,6 +73,11 @@ user:
   kafka:
     enabled: false
```
davidumea, line 75: Can we get some examples here? For me it was not obvious how this should be configured
```suggestion
# Creates extra RBAC
# Examples:
# extraRoles:
#   example:
#     name: example
#     rules:
#     - apiGroups:
#       - '*'
#       resources:
#       - '*'
#       verbs:
#       - '*'
# extraRoleBindings:
#   example:
#     name: example
#    roleRef:
#       name: example
#     subjects:
#     - kind: ServiceAccount
#       name: example
#       namespace: example
```

davidumea (approved):

davidumea:
Nice work, I like this feature!

aarnq (approved):
LGTM!

PR #2372, opened 2024-12-17
Don't get Cluster API resource metrics in workload cluster

anders-elastisys:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

Noticed a bunch of error logs in my Cluster API workload cluster from kube-state-metrics pod:

```console
E1217 14:47:13.509297       1 reflector.go:158] "Unhandled Error" err="pkg/mod/k8s.io/client-go@v0.31.2/tools/cache/reflector.go:243: Failed to watch cluster.x-k8s.io/v1beta1, Kind=Machine: failed to list cluster.x-k8s.io/v1beta1, Kind=Machine: the server could not find the requested resource" logger="UnhandledError"
```

I do not believe we ever have Cluster API resources in the workload clusters so this PR removes the configuration for getting metrics for Cluster API resources in the workload cluster.

- Fixes #

#### Checklist

- Change checks:
- Documentation checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/helmfile.d/stacks/monitoring-prometheus.yaml.gotmpl
+++ b/helmfile.d/stacks/monitoring-prometheus.yaml.gotmpl
@@ -14,7 +14,7 @@ templates:
     needs:
       - monitoring/networkpolicy
       - monitoring/podsecuritypolicy
-      {{- if .Values | get "clusterApi.enabled" false }}
+      {{- if and (.Values | get "ck8sManagementCluster.enabled" false) (.Values | get "clusterApi.enabled" false) }}
       - monitoring/kube-state-metrics-extra-resource-metrics
       {{- end }}
     values:
@@ -90,6 +90,7 @@ templates:
 
   kube-state-metrics-extra-resources:
     inherit: [ template: prometheus ]
+    condition: ck8sManagementCluster.enabled
     installed: {{ .Values | get "clusterApi.enabled" "false" }}
     chart: charts/kube-state-metrics-extra-resource-metrics
     version: 0.1.0
--- a/helmfile.d/values/kube-prometheus-stack-wc.yaml.gotmpl
+++ b/helmfile.d/values/kube-prometheus-stack-wc.yaml.gotmpl
@@ -22,47 +22,6 @@ prometheusOperator:
 
 kube-state-metrics:
   resources: {{- toYaml .Values.kubeStateMetrics.resources | nindent 4 }}
-  {{- if .Values.clusterApi.enabled }}
-  rbac:
-    extraRules:
-      - apiGroups:
-          - apiextensions.k8s.io
-        resources:
-          - customresourcedefinitions
-        verbs:
-          - list
-          - watch
-      - apiGroups:
-          - cluster.x-k8s.io
-        resources:
-          - clusters
-          - machinedeployments
-          - machinesets
-          - machines
-          - machinehealthchecks
-          - machinepools
-        verbs:
-          - get
-          - list
-          - watch
-      - apiGroups:
-          - controlplane.cluster.x-k8s.io
-        resources:
-          - kubeadmcontrolplanes
-        verbs:
-          - get
-          - list
-          - watch
-  volumeMounts:
-    - mountPath: /etc/config
-      name: kube-state-metrics-clusterapi-volume
-  volumes:
-    - configMap:
-        name: kube-state-metrics-clusterapi
-      name: kube-state-metrics-clusterapi-volume
-  extraArgs:
-    - "--custom-resource-state-config-file=/etc/config/clusterapi-metrics.yaml"
-  {{- end }}
 
   metricLabelsAllowlist:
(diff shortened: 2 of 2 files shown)


helmfile.d/stacks/monitoring-prometheus.yaml
```diff
@@ -14,9 +14,11 @@ templates:
     needs:
       - monitoring/networkpolicy
       - monitoring/podsecuritypolicy
+      {{- if .Values | get "ck8sManagementCluster.enabled" false }}
```
Zash, line 17: Sounds like [`needs`](https://helmfile.readthedocs.io/en/latest/#dag-aware-installationdeletion-ordering-with-needs) is primarily about ordering, are these conditionals needed at all?

Xartos, line 17: If, like @Zash is questioning, this is needed. Can we combine the two if statements with an `and` instead of having 2?

anders-elastisys: From testing it does not seem like this check is actually needed so I will remove them

anders-elastisys: Removed the check as it was redundant

anders-elastisys: Nvm, seems like our unit tests complain if it is listed in needs but is not installed :thinking: 
@aarnq as you wrote [these tests](https://github.com/elastisys/compliantkubernetes-apps/blob/main/tests/unit/templates/releases.bash), do you have any input about us changing the conditionals for `needs`?

aarnq: AFAIK, needs in helmfile is not just for ordering, it will check that the needed helm releases are installed before it install this release.

anders-elastisys: It did seem like helmfile ignored the release for `wc` although it was included in needs but it did not follow the `condition`, but would need to do more testing to ensure this, I also feel like it would be better to make it work with our current tests, and then potentially look into this more in the future if we feel that we need to.

anders-elastisys: Added the check back and combined it into one: https://github.com/elastisys/compliantkubernetes-apps/pull/2372/commits/17466d0e9d95e733e1eb6ca066e61259844f737b

helmfile.d/stacks/monitoring-prometheus.yaml.gotmpl
```diff
@@ -90,6 +90,7 @@ templates:
 
   kube-state-metrics-extra-resources:
     inherit: [ template: prometheus ]
+    condition: ck8sManagementCluster.enabled
```
Xartos, line 93: Question: How does this work for existing clusters? I know that `installed` typically is the better to use since it will actually uninstall it if the release already exists. But the `condition` will just skip to check completely. 
So I guess we'd either need to have some migration script for this or move this check to `installed` as well.

anders-elastisys: When I tested applying on workload clusters with kube-state-metrics-extra-resources already deployed with this change, the release got removed so it would seem that releases gets uninstalled based on the condition, so should not need a migration script.

Xartos (approved):

helmfile.d/stacks/monitoring-prometheus.yaml.gotmpl
```diff
@@ -14,7 +14,7 @@ templates:
     needs:
       - monitoring/networkpolicy
       - monitoring/podsecuritypolicy
-      {{- if .Values | get "clusterApi.enabled" false }}
+      {{- if and (.Values | get "ck8sManagementCluster.enabled" false) (.Values | get "clusterApi.enabled" false) }}
```
OlleLarsson, line 17: Would this work? Will we be in the situation where `clusterApi.enabled: false` and `ck8sManagementCluster.enabled: true`?
```suggestion
      {{- if (.Values | get "ck8sManagementCluster.enabled" false) }}
```

anders-elastisys: For kubespray environments we will have `clusterApi.enabled: false` and `ck8sManagementCluster.enabled: true` so both values need to be checked or it will break our unit tests (see the discussion [here](https://github.com/elastisys/compliantkubernetes-apps/pull/2372#discussion_r1890044266)).

Elias-elastisys (approved):

OlleLarsson (approved):

Issue #2287, opened 2024-09-23
Create template migration snippets


aarnq:
### Proposed feature

When we update a larger set of components we most likely need a migration snippet to do some special steps for that stack.

So, we should add custom upgrade snippets that can be reused between versions, as many of those can be reused between versions.

Additionally, it would then be possible to add rollback support there, which then doesn't have to be reinvented between versions.

### Proposed alternatives

One example is OpenSearch, as it has requirements for when things are upgraded so certain configuration jobs run after the core components have been updated.

### Additional context

Note, that these migration snippets _must never do unnecessary changes to a cluster_!

Example, it should not assume it needs to always rerun opensearch-configurer because opensearch _might have_ been upgraded, it should be clear conditions that reruns it _when it has_ upgraded.

### Definition of done

- [ ] Migration template includes predefined snippets
  - Fluentd
  - Grafana
  - Harbor
  - Monitoring
  - OpenSearch
  - Thanos
- [ ] This new feature is covered by contributor/migration documentation

PR #2373, opened 2024-12-18
docs: Update migration handling for next version

aarnq:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change 
- [ ] kind/dev-change   
- [ ] kind/security     
- [ ] [kind/adr]()      

This implements the idea of change for migrations that was proposed before to combat the divergence around releases, and to simplify contribution.
The idea is that the `migration/main` is the one under development, and during release it is promoted to the release series, example `migration/v0.44`.

Hopefully this change in process will also be helpful to realise https://github.com/elastisys/compliantkubernetes-apps/issues/2287, with reusable snippets naturally as we go.

This is in a POC phase, as I've implemented the change in migration structure, but yet to implement the handling in the bin scripts, as I want it to be finalised first.
The process documents supporting migration contribution and release will also need more updates once completed.

#### Information to reviewers

`./migration/promote.sh v0.44`
It is also re-runnable.

#### Checklist

- Change checks:
- Documentation checks:
- Metrics checks:
- Logs checks:
- Pod Security Policy checks:
- Network Policy checks:
- Audit checks:
- Falco checks:
- Bug checks:


aarnq:
Note to self: Update public docs maintenance page.

migration/main/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+    > _Non-disruptive, done before maintenance window._
+
+    ```bash
+    ./bin/ck8s upgrade <target-cluster> main prepare
+
+    # Check if any NetworkPolicy IPs need to be updated
+    ./bin/ck8s update-ips <target-cluster> dry-run
+    # Apply the changes if you agree with them
+    ./bin/ck8s update-ips <target-cluster> apply
```
anders-elastisys, line 61: Personally I would prefer to use an environment variable here for specifying the cluster or clusters to apply the upgrade to, to make it more "copy-paste-able". So maybe have a separate step for setting a variable, e.g `CK8S_CLUSTER` which we are using in the manual steps:
```suggestion
    ./bin/ck8s upgrade ${CK8S_CLUSTER} main prepare

    # Check if any NetworkPolicy IPs need to be updated
    ./bin/ck8s update-ips ${CK8S_CLUSTER} dry-run
    # Apply the changes if you agree with them
    ./bin/ck8s update-ips ${CK8S_CLUSTER} apply
```

migration/main/README.md
```diff
@@ -0,0 +1,128 @@
...
+    ```
+
+1. Apply upgrade
+
+    > **Disruptive, done during maintenance window.**
+
+    ```bash
+    ./bin/ck8s upgrade both <target-cluster> apply
```
anders-elastisys, line 69: Same thing here:
```suggestion
    ./bin/ck8s upgrade both ${CK8S_CLUSTER} apply
```

migration/README.md
```diff
@@ -1,17 +1,51 @@
...
+    - Prepare steps:
+      - are placed in the `prepare/` directory,
+      - may **only** modify the configuration of the environment,
+      - may **not** modify the state of the environment,
+      - steps are run in order of their names use two digit prefixes.
+    - Apply steps:
+      - are placed in the `apply/` directory,
+      - may **only** modify the state of the environment,
+      - may **not** modify the configuration of the environment,
+      - are run in order of their names use two digit prefixes,
```
anders-elastisys, line 26: small nit
```suggestion
      - are run in order of their names using two digit prefixes,
```

migration/README.md
```diff
@@ -1,17 +1,51 @@
...
+      - are run with the argument `execute` on upgrade and should return 1 on failure and 2 on successful internal rollback,
+      - are rerun with the argument `rollback` on execute failure and should return 1 on failure.
+
+For prepare the init step is given.
```
anders-elastisys, line 30: Maybe expand this part a bit.
```suggestion
For the prepare part, a init step is always included in each migration which runs the `ck8s init` command.
```

migration/README.md
```diff
@@ -1,17 +1,51 @@
...
+For apply the bootstrap and the apply steps are given, it is expected that releases upgraded in custom steps are excluded from the apply step.
```
anders-elastisys, line 31: Could maybe expand this a bit more and clarify how and where you are supposed to exclude certain steps, just to make it easier for people that are writing their first migration script :)

anders-elastisys (commented):
Great work, really good that this is getting finally getting addressed as I know that the old system could be problematic if migrations steps did not end up in the proper release folder before a release was in progress. Left some comments, but overall I think it looks really good, have not tested the scripts yet however

migration/main/README.md
aarnq: I've made targeting a separate step, so once that is done the rest is copypasteable.

migration/README.md
aarnq: This has been expanded on.

migration/README.md
aarnq: I'm not sure how to rewrite it, because the hope is that we will have more and more persistent migration snippets, so it wont just be init and apply.
I want to remove it, or maybe just state that one should look for the current ones, and give init and apply as examples as a prepare and apply snippet respectively.
:thinking:

anders-elastisys: Maybe have a link to the respective folders and then provide the init and apply snippets as examples, but I would not say this is blocking in any way

migration/README.md
```diff
@@ -1,17 +1,84 @@
...
+        - Apply steps:
+            - are placed in the `apply/` directory,
+            - may **only** modify the state of the environment,
+            - may **not** modify the configuration of the environment,
+            - are run in order of their names using two digit prefixes,
+            - are run with the argument `execute` on upgrade and should return 1 on failure and 2 on successful internal rollback,
+            - are rerun with the argument `rollback` on execute failure and should return 1 on failure.
+
+[Prepare snippets](main/prepare) are supposed to update the environment configuration to the target version, an example snippet is the init step which runs the `ck8s init` command.
+[Apply snippets](main/apply) are supposed to update the environment application to the target version, and example sippet is the apply step which run `helmfile upgrade` on all releases that have changed.
```
Zash, line 32: Typo:

```suggestion
[Apply snippets](main/apply) are supposed to update the environment application to the target version, and example snippet is the apply step which run `helmfile upgrade` on all releases that have changed.
```

migration/promote.sh
```diff
@@ -0,0 +1,106 @@
...
+
+  local target_version="v${ver["next-major"]}.${ver["next-minor"]}"
+  local target_migration="${ROOT}/migration/${target_version}"
+
+  if [[ -d "${target_migration}" ]]; then
+    if log.continue "migration $(esc.ylw "${target_version}") already exists, do you want to recreate it?"; then
+      log.warn "removing migration $(esc.ylw "${target_version}")"
+      rm -r "${target_migration}"
+    else
+      log.warn "skipping"
```
Zash, line 33: What is it meant to skip?

Should it maybe `exit` here?

If not, is this expected?:

``` console
$ bash ./migration/promote.sh v0.43 
[ck8s/promote.sh] migration v0.43 already exists, do you want to recreate it? - [y/N]: 
[ck8s/promote.sh] skipping
Error: bad file '/tmp/temp2072476512': yaml: did not find expected comment or line break
```

I guess `yq4` has trouble with the missing front matter section in existing migration `README`s

aarnq: It is meant to skip the full recreation of the migration folder, so a complete wipe and copy.
It will still replace references and update text, but maybe it should skip that and just continue to prune. :thinking: 

And yes I assume it is the missing front matter, I would not run it on v0.43 and older which is using the old template.

aarnq: Now that error cannot happen due to old readmes, pruning will still kick in but that is separate. :slightly_smiling_face:

migration/README.md
```diff
@@ -1,17 +1,84 @@
...
-## Migration template and script
+## Writing migration steps for contributions
 
-> [!NOTE]
-> Only major releases are supported by the script
+Notice to developers on writing migration steps:
 
-There is a script that can help you create the initial migration folder and `upgrade-apps.md` document. You can use the following command to run it from the root of compliantkubernetes-apps directory:
+- Migration steps:
+    - are written per major and minor version and placed in a subdirectory of the migration directory with the name `main/`,
```
OlleLarsson, line 13: Nit; feel free to disregard!
```suggestion
    - are written for every major and minor version, and they are placed in a subdirectory of the migration directory with the name `main/`,
```

migration/README.md
```diff
@@ -1,17 +1,84 @@
...
+        - during release this is promoted to the next release series `vX.Y/`,
+    - are written to be idempotent and usable no matter which patch version you are upgrading from and to,
+    - are documented in `main/README.md` to be able to run them manually,
+    - are divided into prepare and apply steps:
+        - Prepare steps:
+            - are placed in the `prepare/` directory,
+            - may **only** modify the configuration of the environment,
+            - may **not** modify the state of the environment,
```
OlleLarsson, line 21: Not sure if we want to use `an environment` instead of `the environment`.
I'll let you decide.

```suggestion
            - may **only** modify the configuration of an environment,
            - may **not** modify the state of an environment,
```

migration/README.md
```diff
@@ -1,17 +1,84 @@
...
+            - steps are run in order of their names using two digit prefixes.
+        - Apply steps:
+            - are placed in the `apply/` directory,
+            - may **only** modify the state of the environment,
+            - may **not** modify the configuration of the environment,
+            - are run in order of their names using two digit prefixes,
+            - are run with the argument `execute` on upgrade and should return 1 on failure and 2 on successful internal rollback,
+            - are rerun with the argument `rollback` on execute failure and should return 1 on failure.
+
+[Prepare snippets](main/prepare) are supposed to update the environment configuration to the target version, an example snippet is the init step which runs the `ck8s init` command.
```
OlleLarsson, line 31: Nit;
```suggestion
[Prepare snippets](main/prepare) are supposed to update the environment's configuration to the target version, an example snippet is the init step which runs the `ck8s init` command.
```

migration/README.md
```diff
@@ -1,17 +1,84 @@
...
+[Apply snippets](main/apply) are supposed to update the environment application to the target version, and example snippet is the apply step which run `helmfile upgrade` on all releases that have changed.
```
OlleLarsson, line 32: Nit;
```suggestion
[Apply snippets](main/apply) are supposed to update the environment's applications to the target versions, and example snippet is the apply step which run `helmfile upgrade` on all releases that have changed.
```

migration/README.md
```diff
@@ -1,17 +1,84 @@
...
+> During runtime these are `AND`:ed together.
+
+Upgrades of components that are dependent on each other should be done within the same snippet to easily manage the upgrade to a working state and to be able to rollback to a working state.
+
+Steps should use the `scripts/migration/lib.sh` which will provide helper functions, see the file for available helper functions.
+This script expects the `ROOT` environment variable to be set pointing to the root of the repository.
+As with all scripts in this repository `CK8S_CONFIG_PATH` is expected to be set.
+
+> [!important]
+> Since the migration snippets should be idempotent, and usable from and to any patch version, it is important that they only perform actions when needed, especially in the case of destructive actions!
```
OlleLarsson, line 68: Nit;
```suggestion
> Since the migration snippets should be idempotent and usable from and to any patch version, it is important that they only perform actions when needed, especially in the case of destructive actions!
```

migration/README.md
```diff
@@ -1,17 +1,84 @@
...
 
 ```bash
-./migration/create-migration-document.sh v0.66 v0.67
+./migration/promote.sh vX.Y
 ```
+
+This script will create a new directory for the `vX.Y` release series from the `main` directory, and prune old directories to only include the latest five releases.
+
+Once the new directory is created you must manually remove snippets and steps that was specific for the new release, be sure to remove any exclusions from them set in the `80-apply-sh` snippet based on the tip in the previous section.
+General migration steps, such as snippets for upgrading OpenSearch, Prometheus, or Thanos, may be retained as long as they verify beforehand they run that they are needed, and fall back to a generic upgrade steps if not required.
```
OlleLarsson, line 84: I'm a bit confused here and I hope you can clarify it for me :) Once the new release migration directory is created, which snippets and steps should I remove? Everything in `main/`?

OlleLarsson (commented):
Looks great! I added some nits and a question.

migration/README.md
aarnq: No, definitely not all.

The idea is to only remove those snippets that are specific to the version that will never be used again, so we can build up a library of reusable ones that we can iteratively improve, and potentially implement proper rollback steps for.

As an example, [`v0.44/apply/02-migrate-gpu-operator-ns.sh`](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration/v0.44/apply/20-migrate-gpu-operator-ns.sh) is only ever going to make sense once.

While [`v0.44/apply/10-kube-prometheus-stack.sh`](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration/v0.44/apply/10-kube-prometheus-stack.sh) is something we've seen in a few versions, and would be better served with a persistent snippet that would be made a bit more generic.
In its current state it already looks good enough to be kept, as it doesn't even use hardcoded versions.

Perhaps the way to go would be to add some sort of tag to the ones we know should be deleted, so that part of the process is clearer. :thinking:

OlleLarsson: Right, I think I understand now!

Maybe the snippets we want to keep could be moved to a `library/` directory?

aarnq: Went with a different option to add suffixes for those that should be deleted, as I think that would require special handling in the migration machinery to keep them in a separate directory.

migration/README.md
```diff
@@ -1,17 +1,87 @@
...
 
 ```bash
-./migration/create-migration-document.sh v0.66 v0.67
+./migration/promote.sh vX.Y
 ```
+
+This script will create a new directory for the `vX.Y` release series from the `main` directory, and prune old directories to only include the latest five releases.
+
+Once the new directory is created you must manually remove snippets and steps that was specific for the new release named with the suffix `-version-specific.sh`, be sure to remove any exclusions from them set in the `80-apply.sh` snippet based on the tip in the previous section.
+General migration steps, such as snippets for upgrading Grafana, OpenSearch, Prometheus, or Thanos, may be retained as long as they verify beforehand they run that they are needed, and fall back to a generic upgrade steps if not required.
```
OlleLarsson, line 87: Maybe something like that?
```suggestion
General migration steps, such as snippets for upgrading Grafana, OpenSearch, Prometheus, or Thanos, may be retained as long as they are verified beforehand that they are needed and run when required, and fall back to a generic upgrade steps if not required.
```

migration/README.md
viktor-f: So the idea right now is that as long as a migration script that is added is version-generic it should be kept in `main/` and would then be part of all future releases (until it is actively removed)?
My feeling right now is that it might end up being a bit messy if we have like 10-20 generic scripts in `main/` that is added to every release, but only some of them are actually needed in any particular release. But maybe it will not grow to be that big.

I sort of like the idea Olle had that we can move the generic scripts to `library/` after the release is done, then copy them back to `main/` later when they are needed for a release again. One disadvantage I see with that is that we need to consider if unused scripts should be removed from the library at some point.

What are your thoughts on this? Should we reconsider using the library option?

migration/promote.sh
```diff
@@ -0,0 +1,118 @@
...
+  [[ -n "${series}" ]] || log.fatal "missing required argument $(esc.ylw "<release-series>"), $(usage)"
+
+  ver.parse next "${series}"
+
+  local target_version="v${ver["next-major"]}.${ver["next-minor"]}"
+  local target_migration="${ROOT}/migration/${target_version}"
+
+  create
+
+  prune
```
simonklb, line 115: Not a fan of relying on global variables like this, to me it's always been a footgun as soon as you start refactoring things. Not a blocker though, I'm fine with both.

simonklb (changes requested):
Nothing major but some things I'd like to be explained and/or fixed before approving the merge! Nice job!

migration/main/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+    > _Non-disruptive, done before maintenance window._
+
+    ```bash
+    ./bin/ck8s upgrade "${CK8S_CLUSTER}" main prepare
+
+    # Check if any NetworkPolicy IPs need to be updated
+    ./bin/ck8s update-ips "${CK8S_CLUSTER}" dry-run
+    # Apply the changes if you agree with them
+    ./bin/ck8s update-ips "${CK8S_CLUSTER}" apply
```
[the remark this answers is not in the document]
simonklb: Uhm, not sure why I was thinking this was actually applying anything to the cluster, I even worked on this just yesterday.. :smile:

migration/README.md
simonklb: This also stumped me a bit, I'm worried that having to remember naming things correctly and removing the correct files etc might be missed or done incorrectly.

Improving the machinery over trusting humans to do the manual process correctly is the way to go IMO.

migration/main/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+1. Apply upgrade
+
+    > **Disruptive, done during maintenance window.**
+
+    ```bash
+    ./bin/ck8s upgrade both "${CK8S_CLUSTER}" apply
+    ```
+
+## Manual method
```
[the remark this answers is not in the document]
aarnq: Maybe we are at the point that we can remove the manual steps. It would be _very_ nice.
I will revise and figure out some helping text to run steps manually if needed, but let the automated way be _the way_ to do it.

migration/README.md
aarnq: > So the idea right now is that as long as a migration script that is added is version-generic it should be kept in `main/` and would then be part of all future releases (until it is actively removed)?

Yes, that is my idea. Because then we can reuse and improve migration snippets over time.
As as per how migration snippets work they will only perform upgrade steps when needed, and will degrade to a generic apply if there are regular changes.

> My feeling right now is that it might end up being a bit messy if we have like 10-20 generic scripts in `main/` that is added to every release, but only some of them are actually needed in any particular release. But maybe it will not grow to be that big.

The idea here is for things like opensearch, it has a common flow when you upgrade it, same for kube-prometheus-stack.
Those would make sense to keep, until we don't have opensearch or kube-prometheus-stack, or we've changed their deployment model so the generic apply is enough in all cases.

> I sort of like the idea Olle had that we can move the generic scripts to `library/` after the release is done, then copy them back to `main/` later when they are needed for a release again. One disadvantage I see with that is that we need to consider if unused scripts should be removed from the library at some point.

I would be against this as it would promote bad snippets that always operates, without being required to.
And then it would be the same as just pulling it from an old version.

> What are your thoughts on this? Should we reconsider using the library option?

The only way I would consider that is if we have a linked in library, that are always sourced, but not as a storage place for snippets not in use with the reasoning above.

> This also stumped me a bit, I'm worried that having to remember naming things correctly and removing the correct files etc might be missed or done incorrectly.
> 
> Improving the machinery over trusting humans to do the manual process correctly is the way to go IMO.

How would we make a machinery to evaluate what humans script are required only for one version in specific?
Or what is your idea?

aarnq: @OlleLarsson @simonklb @viktor-f I'm still waiting on suggestions here.

viktor-f: Forgot to answer.
I think that what you say about the generic migration scripts (e.g. for opensearch) makes sense. So I now lean towards your solution being better than the `library/` solution.

simonklb: > > This also stumped me a bit, I'm worried that having to remember naming things correctly and removing the correct files etc might be missed or done incorrectly.
> > Improving the machinery over trusting humans to do the manual process correctly is the way to go IMO.
> 
> How would we make a machinery to evaluate what humans script are required only for one version in specific? Or what is your idea?

My idea is that you would keep the generic snippets that should always run in one place that is kept around for every migration and version specific snippets in a different location so that you don't have to think about the naming or removing some files and keeping some files around.

For example:
migration/include-always/01-opensearch.bash
migration/include-always/02-prometheus.bash
migration/include-always/03-apply.bash
migration/v0.45/01-specific-version-migration.bash

Then it's up to the person introducing the snippets to put them in their correct spot and not the one promoting to the next version.

aarnq: Yeah, but you have to think of the naming, always, because that is how order is determined, and because of that I see that it is best to keep them in the same directory so it is more intuitive for contributors and reviewers.
Because then what they see is what happens, what is in main will be the next release.

If it is only to identify which snippets are version specific, I rather have a marker in its name because then we can easily identify and automatically prune it when we promote the next version.

```sh
# example
./migration/main/apply/60-prometheus.sh
./migration/main/apply/69-openserch-migrate-snapshots-only-once.sh
./migration/main/apply/70-openserch.sh
./migration/main/apply/80-apply.sh
```

And to add to this, if we must pre-create the migration directory for specific releases when developing specific migrations, then this update won't change anything for what prompted the change in the first place.

simonklb: I think I'm fine with this if we enforce naming so that you have to make the active choice when you introduce a new snippet and then we have automation around what should be left and what should be removed after a promotion. When you promote you should not have to manually pick and choose, it should just be one command and preferably be done as part of cutting a release.

```
./migration/main/apply/60-prometheus-keep.sh
./migration/main/apply/69-openserch-migrate-snapshots-only-once.sh
./migration/main/apply/70-openserch-keep.sh
./migration/main/apply/80-apply-keep.sh
```

Not sure about the naming though.

simonklb: Also, then you can have a linter that enforces the naming scheme so that you are forced to make the choice.

aarnq: Enforced it with suffixes through pre-commit, chose "all-versions" and "one-version" as "only-once" might make the contributor believe that it will ever only be run once, which isn't the case with patches.

scripts/common.sh
```diff
@@ -0,0 +1,166 @@
...
+  log.error.no_newline "${*}\n"
+}
+log.fatal() {
+  log.error "${*}"
+  exit 1
+}
+log.fatal.nested() {
+  shopt -s extdebug
+  local -a function
+  readarray -t function <<<"$(declare -F "${FUNCNAME[1]}" | tr ' ' '\n')"
```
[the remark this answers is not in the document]
aarnq: It is not reserved enough :sweat_smile:

migration/README.md
```diff
@@ -1,17 +1,85 @@
...
+Upgrades of components that are dependent on each other should be done within the same snippet to easily manage the upgrade to a working state and to be able to rollback to a working state.
+
+Steps should use the `scripts/migration/lib.sh` which will provide helper functions, see the file for available helper functions.
+This script expects the `ROOT` environment variable to be set pointing to the root of the repository.
+As with all scripts in this repository `CK8S_CONFIG_PATH` is expected to be set.
+
+> [!important]
+> Since the migration snippets should be idempotent and usable from and to any patch version, it is important that they only perform actions when needed, especially in the case of destructive actions!
+>
+> Example: if a major upgrade of a component requires that another component is temporarily removed, then it should **only** remove the component if there is a major upgrade in the first place, if there is a minor upgrade then it should be applied without removing the component.
```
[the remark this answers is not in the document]
aarnq: To the component it could be yes, if it is really breaking anything, but that isn't very clear.
I'll reword so `major -> major and minor` and `minor -> patch` to make it clearer, as it is in the context of a patch release we're mainly talking about, so on that we should focus.

simonklb (approved):

viktor-f (approved):
Great work! I think it is excellent that we have a clear way to upgrade to current main  and I think the way to keep some migration scripts for future releases turned out really well.
Note that I have not checked the logic of all the scripts very closely.

PR #2399, opened 2025-01-17
Incremental version checks during migration

### LINKED ISSUE #2315: [3] Add applied apps version to cluster state
### Proposed feature

To reduce the risk of forgetting or missing to run the apply migration step we should add some state to the cluster such that it knows what was last applied.

Then when you run the prepare command it could check and make sure that both the config and the cluster state has a valid version.

### Proposed alternatives

Since we would like to be able to catch if any of the migration scrips were missed we should probably have this apply be done in the migration apply step and not directly in apps.

### Additional context

Some thing to consider:
- How to implement? ConfigMap, labels, etc.
- How to handle when you are not on a tag?
- Should this only be checked when preparing an upgrade or also e.g. when running `bin/ck8s apply`?

### Definition of done

- [ ] The current applied apps version is stored in the environment
- [ ] Some prepare step should validate that the cluster is using a valid version before continuing

Zash:
> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

There are now safeguards in place to ensure `ck8s upgrade prepare` and `ck8s upgrade apply` are run in the correct order.

There's also a new `ck8s version` command that reports the verison of apps in the config and cluster(s).

- [x] Stores the current version of apps on first install (when running `ck8s apply`)
- [x] Stores the current version of apps in the cluster **after** `ck8s upgrade`
- [x] Ensure that `ck8s upgrade prepare` is run before `ck8s upgrade apply`
- [x] Ensure that `ck8s upgrade apply` is run before `ck8s apply`
- [x] Ensure that apps version and cluster matches on `ck8s apply`
- [x] Ensure the same configuration is used for `ck8s upgrade prepare` and `ck8s upgrade apply`

```mermaid
flowchart TD;
	init[ck8s init];
	apply[ck8s apply];
	uprep[ck8s upgrade prepare];
	uapply[ck8s upgrade apply];
	unlock[ck8s upgrade unlock];

	init --x init
	init --> apply
	init --> uprep
	init --x uapply
	init --x unlock

	apply --x init
	apply --> apply 
	apply --> uprep
	apply --x uapply
	apply --x unlock

	uprep --x init
	uprep --x apply
	uprep --> uprep
	uprep --> uapply
	uprep --> unlock

	uapply --> init
	uapply --> apply
	uapply --> uprep
	uapply --> uapply
	uapply --x unlock
```

This uses a pair of ConfigMaps in the `kube-system` namespace to keep track of the current version of apps as well as progress during apps upgrades.

- `welkin-apps-meta` stores the current version of apps
	- `.data.version` the version number itself
- `welkin-apps-upgrade` stores details about an in-progress upgrade
	- its existence signifies that `ck8s upgrade ... apply` has started
	- `.data.version` records the version of apps being upgraded to
	- `.data.timestamp` a timestamp also stored in the config at `.global.ck8sLastChange`
	- `.data.last_step` records the last snippet run by `ck8s upgrade ...`
	- configmap is deleted to signify a completed migration

- Fixes #2315

#### Information to reviewers

Please have a look at the tests and consider whether they have good coverage.

#### integration tests

```bash
time make -C tests run-integration/general/upgrade.bats
```

Note: test \# 3 "no ck8s apply without ck8s upgrade" does not behave correctly for me, insisting that the configmap either exists when it should not or doesn't exist after having been successfully created. eventual consistency or caches?

#### unit tests:

Covers only the `ck8s version` command.

```bash
make -C tests run-unit/general/bin-version.bats
```

#### Checklist

- Change checks:
		- [x] The change affects migration itself :)
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/bin/apply.bash
+++ b/bin/apply.bash
@@ -28,6 +28,12 @@ update_ips_dryrun() {
   fi
 }
 
+check_upgrade() {
+  if get_upgrade_status "${1}" &>/dev/null; then
+    log_fatal "Upgrade ongoing, try again when it has completed or use 'ck8s upgrade unlock'"
+  fi
+}
+
 apps_apply() {
   local suppress
 
@@ -49,6 +55,15 @@ apps_apply() {
 
   if (with_kubeconfig "${config["kube_config_$1"]}" helmfile -f "${here}/../helmfile.d/" -e "$2" "${action}" "${concurrency}" "${suppress:-}"); then
     log_info "---"
+    local current_version
+    current_version="$(get_apps_version "${2}" 2>/dev/null || true)"
+    if [ -z "${current_version}" ]; then
+      current_version="$(get_repo_version)"
+      if [[ "${current_version}" == v* ]]; then
+        set_apps_version "${2}" "${current_version%.*}"
+      fi
+    fi
+    log_info "Current version is ${current_version}"
     log_info "Successful Apps ${action} on ${2/_/ }!"
   else
     log_error "---"
@@ -73,5 +88,7 @@ esac
 
 check_node_label "$1" elastisys.io/node-group
 update_ips_dryrun "$1" "${environment}"
+check_upgrade "$1"
 config_load "$1"
+[[ -z "${CK8S_CI_SKIP_APPLY:-}" ]] || exit 0 # Improve mockability in the future
 apps_apply "$1" "${environment}" "${@:2}"
--- a/bin/ck8s
+++ b/bin/ck8s
@@ -35,8 +35,10 @@ usage() {
   echo "  update-ips <wc|sc|both> <apply|dry-run>         Automatically fetches and applies the IPs for network policies" 1>&2
   echo "  upgrade <wc|sc|both> <vX.Y> apply               runs all apply steps upgrading the environment" 1>&2
   echo "  upgrade <wc|sc|both> <vX.Y> prepare             runs all prepare steps upgrading the configuration" 1>&2
+  echo "  upgrade <wc|sc|both> unlock                     reset an upgrade to allow trying again" 1>&2
   echo "  validate <wc|sc>                                validates config files" 1>&2
   echo "  check-requirements                              check the locally installed tool versions" 1>&2
+  echo "  version <all|both|config|sc|wc>                 shows apps version" 1>&2
   exit 1
 }
 
@@ -83,11 +85,13 @@ dry-run)
   ;;
 upgrade)
   [[ "${2}" =~ ^(wc|sc|both)$ ]] || usage
-  [[ "${3}" =~ ^(v[0-9]+\.[0-9]+)$ ]] || usage
-  [[ "${4}" =~ ^(prepare|apply)$ ]] || usage
+  [[ "${3}" =~ ^(v[0-9]+\.[0-9]+|unlock)$ ]] || usage
+  if [[ "${3}" != "unlock" ]]; then
+    [[ "${4}" =~ ^(prepare|apply)$ ]] || usage
+  fi
   check_tools
   export CK8S_CLUSTER="${2}"
-  "${here}/upgrade.bash" "${3}" "${4}"
+  "${here}/upgrade.bash" "${3}" "${4:-}"
   ;;
 team)
   case "${2}" in
@@ -158,6 +162,19 @@ diagnostics)
   [[ "${2}" =~ ^(wc|sc)$ ]] || usage
   shift
   "${here}/diagnostics.bash" "${@}"
+  ;;
+version) # all|both|sc|wc
+  shift
+  [[ "${1}" =~ ^(all|both|config|wc|sc)$ ]] || usage
+  if [[ "${1}" =~ ^(all|config)$ ]]; then
+    log_info "config version:" "$(yq '.global.ck8sVersion' "${CK8S_CONFIG_PATH}/defaults/common-config.yaml")"
+  fi
+  if [[ "${1}" =~ ^(all|both|sc)$ ]]; then
+    log_info "sc version:" "$(get_apps_version "sc")"
+  fi
+  if [[ "${1}" =~ ^(all|both|wc)$ ]]; then
+    log_info "wc version:" "$(get_apps_version "wc")"
+  fi
   ;;
 *) usage ;;
 esac
--- a/bin/common.bash
+++ b/bin/common.bash
@@ -314,7 +314,8 @@ load_config() {
   fi
 }
 
-version_get() {
+# Retrieve version from git
+get_repo_version() {
   pushd "${root_path}" >/dev/null || exit 1
   git describe --exact-match --tags 2>/dev/null || git rev-parse HEAD
   popd >/dev/null || exit 1
@@ -324,7 +325,7 @@ version_get() {
 # TODO: Simple hack to make sure version matches, we need to have a proper way
 #       of making sure that the version is supported in the future.
 validate_version() {
-  version=$(version_get)
+  version=$(get_repo_version)
   if [[ "${1}" == "sc" ]]; then
     merged_config="${config[config_file_sc]}"
   elif [[ "${1}" == "wc" ]]; then
@@ -334,14 +335,27 @@ validate_version() {
     exit 1
   fi
   ck8s_version=$(yq '.global.ck8sVersion' "${merged_config}")
+  cluster_version=$(get_apps_version "${1}" 2>/dev/null || true)
   if [[ -z "$ck8s_version" ]]; then
     log_error "ERROR: No version set. Run init to generate config."
     exit 1
   elif [ "${ck8s_version}" != "any" ] &&
     [ "${version}" != "${ck8s_version}" ]; then
-    log_error "ERROR: Version mismatch. Run init to update config."
-    log_error "Config version: ${ck8s_version}"
-    log_error "Welkin Apps version: ${version}"
+    log_error "ERROR: Version mismatch. Run upgrade to update config."
+    log_error "Welkin Apps cluster version:    ${cluster_version}"
+    log_error "Welkin Apps config version:     ${ck8s_version}"
+    log_error "Welkin Apps repository version: ${version}"
+    exit 1
+  fi
+  if [[ -z "${cluster_version}" ]]; then
+    log_warning "Welkin Apps cluster version:    Unknown"
+    log_warning "Welkin Apps config version:     ${ck8s_version}"
+    log_warning "Welkin Apps repository version: ${version}"
+  elif [[ "${cluster_version}" != "${version%.*}" ]]; then
+    log_error "ERROR: Version mismatch. Run upgrade to update cluster."
+    log_error "Welkin Apps cluster version:    ${cluster_version}"
+    log_error "Welkin Apps config version:     ${ck8s_version}"
+    log_error "Welkin Apps repository version: ${version}"
     exit 1
   fi
 }
@@ -666,4 +680,21 @@ check_node_label() {
       ask_abort
     fi
   fi
+}
+
+# Store apps version to configmap
+# Usage: set_apps_version
+set_apps_version() {
+  "${here}/ops.bash" kubectl "${1}" create configmap --namespace kube-system welkin-apps-meta \
+    --from-literal "version=${2}" >/dev/null
+}
+
+# Retrieve apps version from configmap
+get_apps_version() {
+  "${here}/ops.bash" kubectl "${1}" get --namespace kube-system configmap welkin-apps-meta \
+    --output jsonpath --template='{.data.version}'
+}
+
+get_upgrade_status() {
+  "${here}/ops.bash" kubectl "${1}" get --namespace kube-system configmap welkin-apps-upgrade
 }
--- a/bin/init.bash
+++ b/bin/init.bash
@@ -397,7 +397,7 @@ mkdir -p "${default_config_path}"
 #       for generating user configurations and not store it as a part of
 #       the ck8s configuration.
 mkdir -p "${CK8S_CONFIG_PATH}/user"
-CK8S_VERSION=$(version_get)
+CK8S_VERSION=$(get_repo_version)
 export CK8S_VERSION
 
 generate_default_config "${config[default_common]}"
--- a/bin/upgrade.bash
+++ b/bin/upgrade.bash
@@ -5,6 +5,8 @@
 here="$(readlink -f "$(dirname "${0}")")"
 
 ROOT="$(readlink -f "${here}/../")"
+# Allow overriding from test suite
+MIGRATION_ROOT="${MIGRATION_ROOT:-"${ROOT}/migration"}"
 
 CK8S_STACK="$(basename "$0")"
 export CK8S_STACK
@@ -20,7 +22,7 @@ snippets_list() {
     log_fatal "usage: snippets_list <prepare|apply>"
   fi
 
-  echo "${ROOT}/migration/${CK8S_TARGET_VERSION}/${1}/"* | sort
+  echo "${MIGRATION_ROOT}/${CK8S_TARGET_VERSION}/${1}/"* | sort
 }
 
 snippets_check() {
@@ -59,14 +61,21 @@ prepare() {
     if [[ "$(basename "${snippet}")" == "00-template.sh" ]]; then
       continue
     fi
-
-    log_info "prepare snippet \"${snippet##"${ROOT}/migration/"}\":"
+    log_info "prepare snippet \"${snippet##"${MIGRATION_ROOT}/"}\":"
     if "${snippet}"; then
       log_info "prepare snippet success\n---"
     else
       log_fatal "prepare snippet failure"
     fi
   done
+
+  # Record, in the cluster, that an upgrade has been prepared so that nobody else starts the apply.
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(sc|both)$ ]]; then
+    record_upgrade_prepare_done sc
+  fi
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(wc|both)$ ]]; then
+    record_upgrade_prepare_done wc
+  fi
 
   config_validate secrets
   if [[ "${CK8S_CLUSTER:-}" =~ ^(sc|both)$ ]]; then
@@ -101,9 +110,15 @@ apply() {
       continue
     fi
 
-    log_info "apply snippet \"${snippet##"${ROOT}/migration/"}\":"
+    log_info "apply snippet \"${snippet##"${MIGRATION_ROOT}/"}\":"
     if "${snippet}" execute; then
       log_info "apply snippet success\n---"
+      if [[ "${CK8S_CLUSTER:-}" == "both" ]]; then
+        record_upgrade_apply_step "sc" "${snippet}"
+        record_upgrade_apply_step "wc" "${snippet}"
+      else
+        record_upgrade_apply_step "${CK8S_CLUSTER}" "${snippet}"
+      fi
     else
       local return="${?}"
       log_error "apply snippet execute failure"
@@ -121,6 +136,13 @@ apply() {
       fi
     fi
   done
+
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(sc|both)$ ]]; then
+    record_upgrade_done sc
+  fi
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(wc|both)$ ]]; then
+    record_upgrade_done wc
+  fi
 }
 
 usage() {
@@ -138,12 +160,17 @@ usage() {
 }
 
 main() {
+  if [[ "${1}" == "unlock" ]]; then
+    unlock
+    return
+  fi
+
   local version="${1}"
   local action="${2}"
 
   local pass="true"
   for dir in "" "prepare" "apply"; do
-    if [[ ! -d "${ROOT}/migration/${version}/${dir}" ]]; then
+    if [[ ! -d "${MIGRATION_ROOT}/${version}/${dir}" ]]; then
       log_error "error: migration/${version}/${dir} is not a directory, did you specify the correct version?"
       pass="false"
     fi
@@ -171,6 +198,19 @@ main() {
   "${action}"
 
   log_info "${action} complete"
+}
+
+unlock() {
+  check_config
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(sc|both)$ ]]; then
+    config_load "sc"
+    unlock_upgrade "sc"
+  fi
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(wc|both)$ ]]; then
+    config_load "wc"
+    unlock_upgrade "wc"
+  fi
+  log_info "Cluster upgrade unlocked. You can now retry the upgrade"
 }
 
 main "${@}"
--- a/scripts/local-clusters/configs/secrets.yaml
+++ b/scripts/local-clusters/configs/secrets.yaml
@@ -2,3 +2,8 @@ objectStorage:
   s3:
     accessKey: minioaccesskey
     secretKey: miniosecretkey
+alerts:
+  opsGenie:
+    apiKey: not-used
+  slack:
+    apiUrl: not-used
--- a/scripts/migration/lib.sh
+++ b/scripts/migration/lib.sh
@@ -256,8 +256,8 @@ check_config() {
 
 # usage: check_version <sc|wc> <prepare|apply>
 check_version() {
-  if [[ ! "${1:-}" =~ ^(sc|wc)$ ]] || [[ ! "${2:-}" =~ ^(prepare|apply)$ ]]; then
-    log_fatal "usage: check_version <sc|wc> <prepare|apply>"
+  if [[ ! "${1:-}" =~ ^(sc|wc)$ ]] || [[ ! "${2:-}" =~ ^(prepare|apply|unlock)$ ]]; then
+    log_fatal "usage: check_version <sc|wc> <prepare|apply|unlock>"
   elif [ -z "${CK8S_TARGET_VERSION:-}" ]; then
     log_fatal "error: \"CK8S_TARGET_VERSION\" is unset"
   fi
@@ -282,6 +282,48 @@ check_version() {
       fi
     fi
   fi
+
+  cluster_version="$(get_apps_version "${1}" >/dev/null 2>&1 || true)"
+  case "${2:-}" in
+  prepare)
+    # Ensure not in the middle of upgrading
+    if get_prepared_version "${1}" &>/dev/null; then
+      log_warn "Migration already in progress"
+      if [[ -t 1 ]]; then
+        log_warn_no_newline "Do you want to continue [y/N]: "
+        read -r reply
+        if [[ "${reply}" != "y" ]]; then
+          exit 1
+        fi
+      else
+        # Don't automatically proceed
+        exit 1
+      fi
+    fi
+
+    if [[ -z "${cluster_version}" ]]; then
+      log_warning "Welkin Apps cluster version:     Unknown"
+      log_warning "Welkin Apps config version:      ${VERSION["${1}-config"]%.*}"
+    elif [[ "${cluster_version}" != "${VERSION["${1}-config"]%.*}" ]]; then
+      log_warning "Version mismatch!"
+      log_warning "Welkin Apps cluster version:     ${cluster_version}"
+      log_warning "Welkin Apps config version:      ${VERSION["${1}-config"]%.*}"
+      if [[ -t 1 ]]; then
+        log_warn_no_newline "Do you want to continue [y/N]: "
+        read -r reply
+        if [[ "${reply}" != "y" ]]; then
+          exit 1
+        fi
+      else
+        # On mismatch, don't automatically proceed
+        exit 1
+      fi
+    fi
+    ;;
+  apply)
+    ensure_upgrade_prepared "${1}"
+    ;;
+  esac
 
   if [[ ! "${VERSION["${1}-config"]}" =~ v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
     log_warn "reducing version validation of ${1}-config for version \"${VERSION["${1}-config"]}\""
@@ -367,6 +409,115 @@ append_trap() {
 
   # shellcheck disable=SC2064
   trap "$(new_trap)" "${signal}"
+}
+
+# usage: [[ "$(get_apps_version)" == "0.x" ]]
+get_apps_version() {
+  kubectl_do "${1}" get configmap --namespace kube-system welkin-apps-meta --output=jsonpath --template="{.data.version}"
+}
+
+unlock_upgrade() {
+  kubectl_do "${1}" delete configmap --namespace kube-system welkin-apps-upgrade &>/dev/null ||
+    log_warn "Could not unlock upgrade or upgrade not locked"
+}
+
+# Get currently prepared version, $CK8S_TARGET_VERSION (vX.Y) given when preparing an upgrade
+get_prepared_version() {
+  kubectl_do "${1}" get configmap --namespace kube-system welkin-apps-upgrade --output=jsonpath --template="{.data.version}"
+}
+
+check_prepared_version() {
+  local prepared_version
+  prepared_version="$(get_prepared_version "${1}" || true)"
+  if [ -z "${prepared_version}" ]; then
+    log_fatal "'prepare' step does not appear to have been run, do so first"
+  fi
+
+  if [[ "${prepared_version}" != "${CK8S_TARGET_VERSION}" ]]; then
+    log_fatal "'prepare' step in ${1} appears to have been run for version ${prepared_version}, not ${CK8S_TARGET_VERSION}"
+  fi
+}
+
+# Usage: record_upgrade_prepare_done sc|wc
+# Records that upgrade has been prepared by storing a timestamp in the config and in the cluster along with the target
+# version.
+record_upgrade_prepare_done() {
+  local apps_config_timestamp
+  apps_config_timestamp="$(date -uIs)"
+
+  # This ConfigMap should only exist while doing an upgrade.
+  # Abort if it already exists
+  log_info "Locking cluster ${1} for upgrade"
+  if kubectl_do "${1}" create configmap --dry-run=client --output=yaml \
+    --namespace kube-system welkin-apps-upgrade \
+    --from-literal "version=${CK8S_TARGET_VERSION}" \
+    --from-literal "timestamp=${apps_config_timestamp}" |
+    yq '.metadata.labels["app.kubernetes.io/managed-by"] = "welkin-apps-upgrade"' - |
+    kubectl_do "${1}" create --filename -; then
+
+    ts="${apps_config_timestamp}" \
+      yq_add common '.global.ck8sConfigSerial' 'strenv(ts)'
+    log_info "Cluster ${1} locked for upgrade"
+    return 0
+  else
+    log_warn "prepare already started in ${1} ('ck8s upgrade ${1} unlock' to try again)"
+  fi
+}
+
+# Usage: ensure_upgrade_prepared sc|wc
+# Ensures that the timestamp and version recorded in the config matches that recorded in the cluster.
+ensure_upgrade_prepared() {
+  local apps_upgrade
+  local apps_version
+  local apps_cluster_timestamp
+  local apps_config_timestamp
+
+  if apps_upgrade="$(kubectl_do "${1}" get --namespace kube-system configmap welkin-apps-upgrade --output=yaml 2>/dev/null)"; then
+    apps_version="$(yq '.data.version' <<<"${apps_upgrade}")"
+  else
+    log_fatal "Upgrade has not been prepared, run 'ck8s upgrade prepare' first"
+  fi
+
+  if [[ "${apps_version}" != "${CK8S_TARGET_VERSION}" ]] >/dev/null; then
+    log_fatal "Version mismatch, upgrading to ${CK8S_TARGET_VERSION} but cluster ${1} was prepared for ${apps_version}"
+  fi
+
+  apps_config_timestamp="$(yq '.data.timestamp' <<<"${apps_upgrade}")"
+  apps_cluster_timestamp="$(yq_dig common '.global.ck8sConfigSerial')"
+  if [[ "${apps_config_timestamp}" != "${apps_cluster_timestamp}" ]]; then
+    log_fatal "Config timestamp mismatch, ${apps_cluster_timestamp} in ${1} but ${apps_config_timestamp} in config"
+  fi
+}
+
+# Usage: record_upgrade_apply_step sc|wc step-description
+# Records the last migration snippet that ran successfully, to allow where a failed migration stopped
+record_upgrade_apply_step() {
+  local apps_upgrade
+  apps_upgrade="$(kubectl_do "${1}" get --namespace kube-system configmap welkin-apps-upgrade --output=yaml)"
+  if ! yq --exit-status 'select(.data.version == strenv(CK8S_TARGET_VERSION))' <<<"${apps_upgrade}" >/dev/null; then
+    log_fatal "version mismatch, upgrading to ${CK8S_TARGET_VERSION} but cluster ${1} was prepared for $(
+      yq '.data.version' <<<"${apps_upgrade}"
+    )"
+  fi
+  apps_upgrade="$(last_step="${2##*/}" yq --exit-status '.data.last_apply_step = strenv(last_step)' <<<"${apps_upgrade}")"
+  log_info "Recording upgrade checkpoint"
+  if ! kubectl_do "${1}" replace --filename - <<<"${apps_upgrade}" >/dev/null; then
+    log_fatal "could not record completed upgrade step in ${1}"
+  fi
+}
+
+# Usage: record_upgrade_done sc|wc
+# Records that the upgrade has been completed and records the version upgraded to.
+record_upgrade_done() {
+  # Record the upgraded-to version. Create if it does not already exist.
+  log_info "Recording new apps version in cluster"
+  if ! kubectl_do "${1}" patch --namespace kube-system configmap welkin-apps-meta --type=merge --patch "$(yq --null-input --output-format json '.data.version = strenv(CK8S_TARGET_VERSION)')"; then
+    if ! kubectl_do "${1}" create configmap --namespace kube-system welkin-apps-meta --from-literal "version=${CK8S_TARGET_VERSION}"; then
+      log_fatal "could not record new apps version in ${1}"
+    fi
+  fi
+  # Complete the upgrade.
+  kubectl_do "${1}" delete configmap --namespace kube-system welkin-apps-upgrade >/dev/null
 }
 
 # shellcheck source=scripts/migration/helm.sh
--- a/tests/common/bats/env.bash
+++ b/tests/common/bats/env.bash
@@ -106,6 +106,9 @@ env.init() {
     yq.set 'sc' '.alerts.opsGenieHeartbeat.name' '"example-heartbeat-name"'
   fi
 
+  yq.set 'secrets' '["alerts"]["opsGenie"]["apiKey"]' '"ZXhhbXBsZS1hcGlrZXkx"'
+  yq.set 'secrets' '["alerts"]["slack"]["apiUrl"]' '"https://slack.example/channel/ZXhhbXBsZS1jaGFubmVs"'
+
   yq.set 'wc' '.opa.imageRegistry.URL' '["harbor.ck8s.example.com"]'
 
   yq.set 'wc' '.user.adminUsers' '["user@example.com"]'
--- a/tests/integration/general/upgrade.bats
+++ b/tests/integration/general/upgrade.bats
@@ -1,0 +1,160 @@
+#!/usr/bin/env bats
+
+# bats file_tags=static,general,bin:upgrade
+
+setup_file() {
+  export BATS_NO_PARALLELIZE_WITHIN_FILE=true
+  export CK8S_AUTO_APPROVE=true
+  export CK8S_CI_SKIP_APPLY=true # quicken the apply step
+
+  MIGRATION_ROOT="${ROOT}/tests/integration/general/migration"
+  export MIGRATION_ROOT
+
+  load "../../bats.lib.bash"
+  load_common "local-cluster.bash"
+  local_cluster.configure_selfsigned
+}
+
+setup() {
+  load "../../bats.lib.bash"
+  load_common "local-cluster.bash"
+  load_common "git-version.bash"
+  load_common "yq.bash"
+  load_assert
+  load_file
+  load_mock
+
+  env.private
+
+  gitversion.setup_mocks
+
+  # consistent state at start
+  run yq -i '.global.ck8sVersion="v0.41.0"' "${CK8S_CONFIG_PATH}/defaults/common-config.yaml"
+  ck8s ops kubectl sc delete -n kube-system configmap welkin-apps-upgrade --ignore-not-found --wait
+  ck8s ops kubectl sc apply -f - <<EOF
+apiVersion: v1
+data:
+  version: v0.41
+kind: ConfigMap
+metadata:
+  name: welkin-apps-meta
+  namespace: kube-system
+EOF
+}
+
+teardown() {
+  env.teardown
+}
+
+@test "upgrading works" {
+  # test 1, the happy path where everything goes well
+
+  run ck8s version config
+  assert_success
+  assert_output --partial "v0.41"
+
+  gitversion.mock_static "v0.42.0"
+  run ck8s upgrade sc "v0.42" prepare
+  assert_success
+  assert_output --partial "locked for upgrade"
+
+  gitversion.mock_static "v0.42.0"
+  run ck8s upgrade sc "v0.42" apply
+  assert_success
+
+  run ck8s apply sc
+  assert_success
+
+  run ck8s version sc
+  assert_success
+  assert_output --partial "v0.42"
+}
+
+@test "no upgrade apply without upgrade prepare" {
+  # test 2
+
+  gitversion.mock_static "v0.42.0"
+  run ck8s version config
+  assert_success
+  assert_output --partial "v0.41"
+  run ck8s upgrade sc "v0.42" apply
+  assert_failure # because trying to apply 0.42 when config says 0.41
+  assert_output --partial "Upgrade has not been prepared, run 'ck8s upgrade prepare' first"
+}
+
+@test "prevent apply without upgrade apply after upgrade prepare" {
+  # test 3
+
+  gitversion.mock_static "v0.42.0"
+  run ck8s upgrade sc "v0.42" prepare
+  assert_success
+  assert_output --partial "locked for upgrade"
+
+  run ck8s ops kubectl sc get -n kube-system configmap welkin-apps-upgrade -o jsonpath --template='{.data.version}'
+  assert_success
+  assert_output --partial "v0.42"
+
+  # Oops, forgot to ck8s upgrade apply
+
+  run ck8s apply sc
+  assert_failure
+  assert_output --partial "Upgrade ongoing"
+}
+
+@test "prevent a ck8s apply after ck8s upgrade prepare but prepare is not merged" {
+  # test 4
+
+  gitversion.mock_static "v0.42.0"
+  run ck8s version config
+  assert_success
+  assert_output --partial "v0.41"
+
+  run ck8s upgrade sc "v0.42" prepare
+  assert_success
+
+  # someone changes the config
+  run yq -i '.global.ck8sConfigSerial="something"' "${CK8S_CONFIG_PATH}/common-config.yaml"
+
+  run ck8s upgrade sc "v0.42" apply
+  assert_failure
+  assert_output --partial "Config timestamp mismatch"
+
+}
+
+@test "prevent a ck8s apply on an old config after ck8s upgrade apply" {
+  run ck8s ops kubectl sc patch -n kube-system configmap welkin-apps-meta --type=merge --patch '{"data":{"version":"v0.42"}}'
+
+  # accidentally downgrade apps
+  gitversion.mock_static "v0.41.0"
+  run ck8s apply sc
+  assert_failure
+  assert_output --partial "Version mismatch. Run upgrade to update cluster."
+}
+
+@test "prevent prepare when migration already started" {
+  run ck8s version config
+  assert_success
+  assert_output --partial "v0.41"
+
+  gitversion.mock_static "v0.42.0"
+  run ck8s upgrade sc "v0.42" prepare
+  assert_success
+  assert_output --partial "locked for upgrade"
+
+  gitversion.mock_static "v0.42.0"
+  run ck8s upgrade sc "v0.42" prepare
+  assert_failure
+  assert_output --partial "Migration already in progress"
+}
+
+@test "prevent prepare on a different version" {
+  gitversion.mock_static "v0.42.0"
+  run ck8s upgrade sc "v0.42" prepare
+  assert_success
+  assert_output --partial "locked for upgrade"
+
+  gitversion.mock_static "v0.43.0"
+  run ck8s upgrade sc "v0.43" apply
+  assert_failure
+  assert_output --partial "Version mismatch, upgrading to v0.43 but cluster sc was prepared for v0.42"
+}
--- a/tests/unit/general/bin-version.bats
+++ b/tests/unit/general/bin-version.bats
@@ -1,0 +1,74 @@
+#!/usr/bin/env bats
+
+# bats file_tags=static,general,bin:version
+
+setup_file() {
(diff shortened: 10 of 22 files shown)


aarnq:
What is the motivation for keeping the phase of prepare in-cluster?

Isn't it critical to track the apply phase?

Zash:
> What is the motivation for keeping the phase of prepare in-cluster?

Seemed a natural starting point for these checks.
 
> Isn't it just critical to track the apply phase?

Yes.

Zash:
https://github.com/elastisys/compliantkubernetes-apps/issues/2315#issuecomment-2416989660 is probably where I got that from, to track the whole upgrade process from the start.

bin/upgrade.bash
```diff
@@ -52,6 +52,18 @@ prepare() {
...
+  # Create a configmap. This fails if done twice.
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(sc|both)$ ]]; then
+    if ! record_migration_prepare_begin sc; then
+      log_fatal "prepare already started in sc (kubectl delete -n kube-system configmap apps-upgrade to try again)"
+    fi
+  fi
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(wc|both)$ ]]; then
+    if ! record_migration_prepare_begin wc; then
+      log_fatal "prepare already started in wc (kubectl delete -n kube-system configmap apps-upgrade to try again)"
+    fi
```
simonklb, line 64: suggestion: Move the `log_fatal` inside the function instead of assuming what a non-zero exit code means.

bin/upgrade.bash
```diff
@@ -93,6 +113,28 @@ apply() {
...
+  local prepared_version
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(sc|both)$ ]]; then
+    prepared_version="$(get_prepared_version sc || true)"
+    if [ -z "${prepared_version}" ]; then
+      log_fatal "'prepare' step does not appear to have been run, do so first"
+    fi
+
+    if [[ "${prepared_version}" != "${CK8S_TARGET_VERSION}" ]]; then
+      log_fatal "'prepare' step in sc appears to have been run for version ${prepared_version}, not ${CK8S_TARGET_VERSION}"
+    fi
```
simonklb, line 125: suggestion: Extract into a separate function and reuse for both sc and wc.

scripts/migration/lib.sh
```diff
@@ -367,6 +376,66 @@ append_trap() {
...
+
+# Usage: record_migration_prepare_done sc|wc
+record_migration_prepare_done() {
+  # assert that the above begin has been done and from the expected version
+  # yq4 should trip pipefail on version mismatch
+  kubectl_do "${1}" get -n kube-system cm apps-upgrade -o yaml 2>/dev/null |
+    yq4 --exit-status 'select(.data.prepare == strenv(CK8S_TARGET_VERSION)) |
+      .data.prepared = strenv(CK8S_TARGET_VERSION) |
+      del(.data.last_prepare_step)' |
+    kubectl_do "${1}" replace -f - >/dev/null 2>/dev/null
```
simonklb, line 402: suggestion: While one-lines are nice I think reducing the magic a bit here could help with readability and make it less error-prone, something like:
```
current_version=$(kubectl get configmap apps-upgrade -o jsonpath='{.data.prepare}')
[ $current_version == $CK8S_TARGET_VERSION ] || log_fatal "version mismatch"
kubectl patch configmap apps-upgrade --type=json -p='[{"op": "replace", "path": "/data/prepared", "value":"$CK8S_TARGET_VERSION"}]'
kubectl patch configmap apps-upgrade --type=json -p='[{"op": "remove", "path": "/data/last_prepare_step"}]'
```

simonklb (approved):
This LGTM!

Regarding the tracking of the prepare phase, I like that we are able to make sure that the config has been properly updated and get rid of the `ck8sVersion` from the config on the other hand if there is no Kubernetes cluster to verify against then we still have to rely on something external.

No strong opinion here though. It doesn't hurt but might add some extra complexity.

scripts/migration/lib.sh
Zash: Does that resist [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issues? I was hoping that doing this as a single pipeline would ensure that the version could not be changed between the test and the patching. Sadly the JSON Patch subset used by kubernetes does not support `"op":"test"`  😭

simonklb: As long as `record_migration_begin` isn't suffering from it I think we should be fine, right? Since that is essentially acting as a lock.

If we aren't worried that the version is changed out of bounds that is, but then we have bigger issues than a race condition between the test and the patch. :smile:

simonklb: Speaking of resisting issues. Tests would be nice!

simonklb (approved):
Still approved after removing the prepare phase tracking and keep relying on `ck8sVersion` in the config file. However, I still do think that getting rid of the `ck8sVersion` config value and only rely on the `apps-meta` config map would have been neat.

bin/upgrade.bash
```diff
@@ -83,6 +83,18 @@ apply() {
     config_validate wc
   fi
 
+  # Create a configmap. This fails if done twice.
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(sc|both)$ ]]; then
+    if ! record_migration_begin sc; then
+      log_fatal "prepare already started in sc (kubectl delete -n kube-system configmap apps-upgrade to try again)"
```
aarnq, line 89:
```suggestion
      log_fatal "apply already started in sc (kubectl delete -n kube-system configmap apps-upgrade to try again)"
```

Also, would it make sense to add a `ck8s upgrade <prefix> unlock-apply` to do this?
(With some warning+confirm that you probably shouldn't unless you know what you are doing.)

bin/upgrade.bash
```diff
@@ -83,6 +83,18 @@ apply() {
...
+    fi
+  fi
+  if [[ "${CK8S_CLUSTER:-}" =~ ^(wc|both)$ ]]; then
+    if ! record_migration_begin wc; then
+      log_fatal "prepare already started in wc (kubectl delete -n kube-system configmap apps-upgrade to try again)"
```
aarnq, line 94:
```suggestion
      log_fatal "apply already started in wc (kubectl delete -n kube-system configmap apps-upgrade to try again)"
```

bin/upgrade.bash
```diff
@@ -93,6 +113,28 @@ apply() {
...
+    prepared_version="$(get_prepared_version wc || true)"
+    if [ -z "${prepared_version}" ]; then
+      log_fatal "'prepare' step does not appear to have been run, do so first"
+    fi
+
+    if [[ "${prepared_version}" != "${CK8S_TARGET_VERSION}" ]]; then
+      log_fatal "'prepare' step in wc appears to have been run for version ${prepared_version}, not ${CK8S_TARGET_VERSION}"
+    fi
+  fi
+
```
Zash: Note from @Xartos - record beginning of apply step

scripts/migration/lib.sh
Zash: Tests **would** be nice! Figuring out how to test whole migrations feels like a whole task in itself however, unless we already have it somewhere? Maybe @aarnq or @viktor-f have thoughts on that?

simonklb: Actually testing migrations would be awesome but feels like too big of an undertaking here.

I think something as simple as testing that running two migrations concurrently is prohibited and that migrations are actually being tracked. The migrations in the tests could just be dummies.

simonklb:
> > Regarding the tracking of the prepare phase, I like that we are able to make sure that the config has been properly updated and get rid of the `ck8sVersion` from the config on the other hand if there is no Kubernetes cluster to verify against then we still have to rely on something external.
> 
> But if it is in cluster it doesn't say anything, as the config isn't attached to it. Then you must apply some sort of checksumming to assert that it is _that_ config you mean. Or we have to move the entire config in cluster. Else if someone has done prepare, then someone else _without that config_ can run apply.

Yes you're right, the inherent issue of working with distributed configuration files still applies. At least now we can make sure that the cluster has been migrated to a new version which is a big win.

I think my main concern with the `ck8sVersion` config is that it leads to a bad habit of thinking that just running `ck8s init` and you're good to go. I've seen it countless of times where the configuration validation succeeds and apply breaks because the migration was never run.

scripts/migration/lib.sh
Elias-elastisys: I also agree that it might be better to split this into separate lines, probably in all the new functions if so.
Also am I wrong in assuming TOCTOU could happen with this pipeline anyways if the version is changed by someone else?

Zash: Try this:

``` shell
$ kubectl create confimap -n kube-system atomic-test --from-literal=version=v1

# save this version for later (pretend this is the start of the pipeline)
$ kubectl get confimap -n kube-system atomic-test -o yaml > atomic-test.yaml

# pretend this is a separate process
$ kubectl patch confimap -n kube-system atomic-test --type=json -p='[{"op":"replace","path":"/data/version","value":"v3"}]'

# continuation of the pipeline
$ yq4 -i '.data.version="v2"' atomic-test.yaml
$ kubectl apply -f atomic-test.yaml
```

For me the `apply` fails, complaining that the object has been modified (which it was, by the `patch` command), which was the property I was aiming for here.

simonklb: Yea I think it's worth keeping the begin atomic but I'm fine keeping the rest as-is as well. Just a suggestion!

bin/upgrade.bash
Zash: I pushed a thing with `--force`, would a separate sub-command be better?

scripts/migration/lib.sh
Zash: Would have been much nicer if the [`test` operation](https://datatracker.ietf.org/doc/html/rfc6902#section-4.6) was supported.

I pushed some refactoring (untested!!), PTAL

simonklb: Looks good! @Elias-elastisys feel free to re-open if you're not satisfied!

scripts/migration/lib.sh
```diff
@@ -367,6 +376,83 @@ append_trap() {
...
+
+# Get currently prepared version
+get_prepared_version() {
+  kubectl_do "${1}" get cm -n kube-system apps-upgrade -o jsonpath --template="{.data.prepared}" 2>/dev/null
+}
+
+# Usage: record_migration_prepare_step sc|wc step-description
+record_migration_prepare_step() {
+  local apps_upgrade
+  apps_upgrade="$(kubectl_do "${1}" get -n kube-system cm apps-upgrade -o yaml 2>/dev/null)"
```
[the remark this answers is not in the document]
Zash: Worth reconsidering. I believe I wanted to get rid of the success messages as well as replace the errors with `log_pipefail`. Is there a `--quiet` option?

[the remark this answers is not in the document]
simonklb: If by success messages you mean for example `namespace/foo created` then that is printed to stdout which I think is fine to pipe to /dev/null.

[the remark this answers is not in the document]
Zash: Yes, the `created` and `patched` messages. So, drop all the `>/dev/null` ones or ?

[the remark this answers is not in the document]
simonklb: No I'm fine with keeping the `>/dev/null` to silence the `created` and `patched` messages but remove the `2>/dev/null` to ensure that errors are printed.

bin/upgrade.bash
aarnq: I think so, because else some people might just always run with `--force` and that would defeat the purpose.

bin/ck8s
```diff
@@ -153,5 +154,17 @@ diagnostics)
...
+  case "${1:-both}" in
+  both)
+    log_info "sc version:" "$(get_version "sc")"
+    log_info "wc version:" "$(get_version "wc")"
+    ;;
+  sc | wc)
+    log_info "${1} version:" "$(get_version "${1}")"
+    ;;
+  esac
+  ;;
```
viktor-f, line 168: Nice to have this command. I think it could be even better if it also printed the version in the config, so you can easily see if they are the same. Does that sound good to you as well?

scripts/migration/lib.sh
```diff
@@ -281,6 +282,14 @@ check_version() {
...
   fi
 
+  cluster_version="$(get_apps_version "${1}" >/dev/null 2>&1 || true)"
+  if [ -n "${cluster_version}" ]; then
+    log_info "Currently running ${cluster_version} according to cluster"
+    if [ "${common_override}" != "null" ] && [ "${cluster_version%.*}" != "${common_override}" ]; then
+      log_warn "Version mismatch, cluster ${cluster_version}, config ${common_override}"
+    fi
+  fi
+
```
viktor-f, line 292: Question: Could you elaborate on why we have this block? If I understand it correctly then it will warn if there is a version override in the common config and that is not the same as the applied version. But it will not exit and is not part of a prompt, so it would just be informational when looking at what is happening.

Could it perhaps make sense to merge this check into the prompt above it? Then you would still get this information, but it can be used as more context to decide if you want to abort or not. What do you think of that?

bin/ck8s
Zash: Sounds good.

Should just be one version in `defaults/common-config.yaml`, or are there cases where wc and sc may differ?

scripts/migration/lib.sh
Zash: Hm, probably missed that it's the _override_ config it checks against. The intent was probably to check the pre-upgrade config version (`defaults/common-config.yaml → .global.ck8sVersion`) against the in-cluster version (if available). E.g. if you try to upgrade to `v0.43` and your config says `v0.42` but your cluster says `v0.41` or similar. This gets complicated and hard to think about since the config is changed by `ck8s init`.

Maybe just drop this, or what do you think is the way forward?

scripts/migration/lib.sh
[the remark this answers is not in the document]
Zash: Pushed some updates, PTAL

bin/upgrade.bash
Zash: Pushed a `ck8s reset-upgrade` command to bypass some of the argument related logic that `ck8s upgrade ...` does. How's it look?

scripts/migration/lib.sh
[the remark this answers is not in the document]
simonklb: Schweet!

aarnq:
> > > Regarding the tracking of the prepare phase, I like that we are able to make sure that the config has been properly updated and get rid of the `ck8sVersion` from the config on the other hand if there is no Kubernetes cluster to verify against then we still have to rely on something external.
> > 
> > 
> > But if it is in cluster it doesn't say anything, as the config isn't attached to it. Then you must apply some sort of checksumming to assert that it is _that_ config you mean. Or we have to move the entire config in cluster. Else if someone has done prepare, then someone else _without that config_ can run apply.
> 
> Yes you're right, the inherent issue of working with distributed configuration files still applies. At least now we can make sure that the cluster has been migrated to a new version which is a big win.
> 
> I think my main concern with the `ck8sVersion` config is that it leads to a false sense of security and a bad habit of thinking that just running `ck8s init` and you're good to go. I've seen it countless of times where the configuration validation succeeds and apply breaks because the migration was never run.

If that is the concern then we should change how `ck8sVersion` is handled, because I do agree that this is an issue because init is not supposed to be _the_ prepare step.

bin/ck8s
viktor-f: I think that the only logical place to override the version is in common. It might be technically possible to have it in sc or wc, but that is not intended.

So just comparing with common should be fine, but doing the regular merge of sc with common and wc with common might be safest.

scripts/migration/lib.sh
viktor-f: I'm not sure what the best way forward is.
Merging it with the block above might still be best, so that you get a chance to abort. Because I think that there is some value to having this check.

Zash: I moved the block above the abort check but removed the mismatch check since it would be comparing the to-be-upgraded-to version recorded by `ck8s init` to the currently running version, which wasn't the intent and it would always mismatch unless try to migrate a second time.

Zash:
> > I think my main concern with the `ck8sVersion` config is that it leads to a false sense of security and a bad habit of thinking that just running `ck8s init` and you're good to go. I've seen it countless of times where the configuration validation succeeds and apply breaks because the migration was never run.
> 
> If that is the concern then we should change how `ck8sVersion` is handled, because I do agree that this is an issue because init is not supposed to be _the_ prepare step.

Would it be enough to add something here that checks in the apply steps that `config-version == cluster version`, or do we need to rethink migrations from `ck8s init` onwards?

aarnq:
> > > I think my main concern with the `ck8sVersion` config is that it leads to a false sense of security and a bad habit of thinking that just running `ck8s init` and you're good to go. I've seen it countless of times where the configuration validation succeeds and apply breaks because the migration was never run.
> > 
> > 
> > If that is the concern then we should change how `ck8sVersion` is handled, because I do agree that this is an issue because init is not supposed to be _the_ prepare step.
> 
> Would it be enough to add something here that checks in the apply steps that `config-version == cluster version`, or do we need to rethink migrations from `ck8s init` onwards?

Or just move the update of ck8sVersion to be done as a final step when upgrade prepare completes.
Init should still be able to set it in a new config though.

bin/ck8s
```diff
@@ -35,7 +35,9 @@ usage() {
   echo "  update-ips <wc|sc|both> <apply|dry-run>         Automatically fetches and applies the IPs for network policies" 1>&2
   echo "  upgrade <wc|sc|both> <vX.Y> apply               runs all apply steps upgrading the environment" 1>&2
   echo "  upgrade <wc|sc|both> <vX.Y> prepare             runs all prepare steps upgrading the configuration" 1>&2
+  echo "  unlock-upgrade <wc|sc|both>                     reset a migration to allow trying again" 1>&2
```
aarnq, line 38: Nit, could this be `upgrade-unlock` instead? Makes similar commands stay together.
Or perhaps `upgrade <wc|sc|both> unlock` to be more similar in structure to the rest of the upgrade commands.

Zash: > Nit, could this be `upgrade-unlock` instead?

Sure

> Or perhaps `upgrade <wc|sc|both> unlock`

Finding a good way to bypass the existing path that requires a version number was a bit awkward.

bin/upgrade.bash
Zash: 1877e26f

bin/ck8s
```diff
@@ -83,6 +85,11 @@ upgrade)
   export CK8S_CLUSTER="${2}"
   "${here}/upgrade.bash" "${3}" "${4}"
   ;;
+upgrade-unlock)
+  [[ "${2}" =~ ^(wc|sc|both)$ ]] || usage
+  export CK8S_CLUSTER="${2}"
+  "${here}/upgrade.bash" "unlock"
+  ;;
```
Zash, line 92:
```bash
  echo "  upgrade <wc|sc|both> unlock                     reset a migration to allow trying again" 1>&2
####
upgrade)
  [[ "${2}" =~ ^(wc|sc|both)$ ]] || usage
  if [[ "${3}" == "unlock" ]]; then
    export CK8S_CLUSTER="${2}"
    "${here}/upgrade.bash" "unlock"
    exit 0
  fi
  [[ "${3}" =~ ^(v[0-9]+\.[0-9]+)$ ]] || usage
  [[ "${4}" =~ ^(prepare|apply)$ ]] || usage
  check_tools
  export CK8S_CLUSTER="${2}"
  "${here}/upgrade.bash" "${3}" "${4}"
  ;;
```

I guess this would work, tho feels a bit awkward to jump out in the middle here. @aarnq ?

aarnq: Depends on how you write it, you don't have to jump in the middle.

```sh
[[ "${2}" =~ ^(wc|sc|both)$ ]] || usage
[[ "${3}" =~ ^(v[0-9]+\.[0-9]+|unlock)$ ]] || usage
if [[ "${3}" != "unlock" ]]; then
  [[ "${4}" =~ ^(prepare|apply)$ ]] || usage
fi
check_tools
export CK8S_CLUSTER="${2}"
"${here}/upgrade.bash" "${3}" "${4:-}"
```

simonklb: Is this resolved?

aarnq:
Could you schedule a time and we can look through it in detail?

bin/ck8s
Zash: This was about whether the `unlock` command should be part of `upgrade`, i.e. `upgrade <wc|sc|both> prepare` vs `upgrade-unlock <wc|sc|both>` and how to handle the different number of in-the-middle arguments.

bin/apply.bash
```diff
@@ -47,6 +47,17 @@ apps_apply() {
   log_info "---"
   log_info "Start Apps ${action} on ${2/_/ }"
 
+  validate_version "${1}"
+  if get_migration_status "${1}" &>/dev/null; then
+    log_fatal "Migration ongoing, try again when it has completed or 'ck8s upgrade unlock'"
+  fi
```
Zash, line 53: TODO move into function and call before `apps_apply()`

tests/migration/README.md
```diff
@@ -0,0 +1,7 @@
+# Migration testing migrations
```
Zash, line 1: move into `integration/general`

tests/common/bats/migration-override.bash
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/bash
+
+migration.override_path() {
+  MIGRATION_ROOT="${ROOT}/tests/migration"
```
Zash, line 4: move into ... setup?

scripts/local-cluster.sh
```diff
@@ -228,6 +228,8 @@ resolve() {
             resolvectl dns "${link}" 127.0.64.43
           fi
         done
+      elif command -v unbound-control &>/dev/null; then
+        sudo unbound-control forward_add +i test.dev-ck8s.com 127.0.64.43
```
Zash, line 232: ask nicely first

scripts/migration/lib.sh
```diff
@@ -265,6 +265,18 @@ check_version() {
     return
   fi
 
+  if [ "${2}" == "prepare" ]; then
+    echo TODO >/dev/null
+    # config version should not yet have been touched
+    # how to ensure this?
```
Zash, line 271: obsolete?

scripts/migration/lib.sh
```diff
@@ -265,6 +265,18 @@ check_version() {
...
+  elif [ "${2}" == "apply" ]; then
+    # ensure config version == target version, set by init during prepare DONE? DONE?
+    if [ "${VERSION["${1}-config"]%.*}" != "${CK8S_TARGET_VERSION}" ]; then
+      log_fatal "Apps version mismatch, '${VERSION["${1}-config"]%.*}' in config but upgrading to '${CK8S_TARGET_VERSION}'"
```
Zash, line 275: so, redundant?

scripts/migration/lib.sh
```diff
@@ -367,6 +419,111 @@ append_trap() {
   trap "$(new_trap)" "${signal}"
 }
 
+# usage: [[ "$(get_apps_version)" == "0.x" ]]
+get_apps_version() {
+  kubectl_do "${1}" get cm -n kube-system apps-meta -o jsonpath --template="{.data.version}"
+}
+
+unlock_migration() {
+  kubectl_do "${1}" delete configmap -n kube-system apps-upgrade >/dev/null
```
Zash, line 428: error handling

scripts/migration/lib.sh
```diff
@@ -367,6 +419,111 @@ append_trap() {
...
+
+# Usage: record_migration_prepare_done sc|wc
+# Records that migration has been prepared by storing a timestamp in the config and in the cluster along with the target
+# version.
+record_migration_prepare_done() {
+  local apps_config_timestamp
+  apps_config_timestamp="$(date +uIs)"
+
+  ts="${apps_config_timestamp}" \
+    yq4 -i '.global.ck8sLastChange=strenv(ts)' "${CK8S_CONFIG_PATH}/defaults/common-config.yaml"
```
Zash, line 456: `s/defaults//`

bin/apply.bash
Zash: Turns out to need a larger refactor to allow than what seems in scope for this PR, changing to using a variable instead to make tests faster

bin/upgrade.bash
```diff
@@ -56,15 +58,22 @@ prepare() {
     if [[ "$(basename "${snippet}")" == "00-template.sh" ]]; then
       continue
     fi
-
-    log_info "prepare snippet \"${snippet##"${ROOT}/migration/"}\":"
+    log_info "prepare snippet \"${snippet##"${MIGRATION_ROOT}/"}\":"
```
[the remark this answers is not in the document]
Zash: line 9, for overriding during tests

bin/apply.bash
```diff
@@ -28,6 +28,12 @@ update_ips_dryrun() {
   fi
 }
 
+check_migration() {
+  if get_migration_status "${1}" &>/dev/null; then
+    log_fatal "Migration ongoing, try again when it has completed or 'ck8s upgrade unlock'"
+  fi
+}
```
aarnq, line 35: Would it make sense to instead say upgrade instead of migration in the cli? It feels more logical when that is the command is upgrade.

bin/init.bash
```diff
@@ -328,7 +328,8 @@ backup_file() {
 
   log_info "Creating backup ${backup_config_path}/${backup_name}"
 
-  cp "${file}" "${backup_config_path}/${backup_name}"
+  # --force to workaround issue in integration test where file already exists
+  cp --force --backup "${file}" "${backup_config_path}/${backup_name}"
```
aarnq, line 332: Is this still needed, or was it because it ran multiple ones concurrently?

scripts/local-cluster.sh
```diff
@@ -226,12 +226,12 @@ resolve() {
   if [[ -z "${action}" ]]; then
     log.usage
   fi
+  if [[ -z "${domain}" ]]; then
+    log.usage
+  fi
```
aarnq, line 231: This will require changes where it is used.

bin/apply.bash
Zash: Makes sense, with upgrade being the overall action and migrations being individual changes applied.

Zash:
```suggestion
check_migration() {
  if get_migration_status "${1}" &>/dev/null; then
    log_fatal "Upgrade ongoing, try again when it has completed or 'ck8s upgrade unlock'"
  fi
}
```

bin/init.bash
Zash: Will check 🤔

scripts/local-cluster.sh
Zash: Aware of any uses besides these?

https://github.com/elastisys/compliantkubernetes-apps/blob/d3ff4df7a14131991514b6d2fbdbb15e339850a2/.github/workflows/tests.yml#L93

https://github.com/elastisys/compliantkubernetes-apps/blob/d3ff4df7a14131991514b6d2fbdbb15e339850a2/.github/workflows/tests.yml#L139

bin/init.bash
Zash: Tests pass with that reverted, so doesn't seem necessary anymore

bin/apply.bash
```diff
@@ -28,6 +28,12 @@ update_ips_dryrun() {
   fi
 }
 
+check_upgrade() {
+  if get_upgrade_status "${1}" &>/dev/null; then
+    log_fatal "Upgrade ongoing, try again when it has completed or 'ck8s upgrade unlock'"
```
aarnq, line 33: Suggestion:
```suggestion
    log_fatal "Upgrade ongoing, try again when it has completed or use 'ck8s upgrade unlock'"
```

bin/upgrade.bash
```diff
@@ -5,6 +5,8 @@
 here="$(readlink -f "$(dirname "${0}")")"
 
 ROOT="$(readlink -f "${here}/../")"
+# Allow overriding from test suite
+MIGRATION_ROOT="${MIGRATION_ROOT:-$ROOT/migration}"
```
aarnq, line 9:
```suggestion
MIGRATION_ROOT="${MIGRATION_ROOT:-"${ROOT}/migration"}"
```

bin/apply.bash
Zash: Accidentally a word 🤣

scripts/migration/lib.sh
```diff
@@ -281,6 +281,46 @@ check_version() {
     fi
   fi
 
+  # Make sure **config** and **cluster** is on the same (old) version
+  cluster_version="$(get_apps_version "${1}" >/dev/null 2>&1 || true)"
+  if [ -z "${cluster_version}" ]; then
+    log_warn "Unknown cluster version!"
+    # TODO ensure this configmap exists on first-run
```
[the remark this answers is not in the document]
Zash: Haven't added that yet, did we decide whether it could be another PR?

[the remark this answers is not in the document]
aarnq: Isn't it required for this to work? :thinking:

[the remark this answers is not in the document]
Zash: It wasn't supposed to be required for the upgrade checks, but now I see the `get_prepared_version` call is in the `elif` block after here and that should still run after this. 😬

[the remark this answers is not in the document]
aarnq: Then this needs a manual step in the next migration to add it?

bin/apply.bash
```diff
@@ -49,6 +55,7 @@ apps_apply() {
 
   if (with_kubeconfig "${config["kube_config_$1"]}" helmfile -f "${here}/../helmfile.d/state.yaml" -e "$2" "${action}" "${concurrency}" "${suppress:-}"); then
     log_info "---"
+    set_apps_version "${2}" || log_info "Current version is $(get_apps_version "${2}")"
```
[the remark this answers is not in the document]
Zash: `&>/dev/null` all the things, or better to check if the version configmap exists first?

scripts/migration/lib.sh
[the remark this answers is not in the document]
Zash: Added something to store the current version for a _fresh install_. It was already added at the end of a migration.

scripts/migration/lib.sh
```diff
@@ -367,6 +407,112 @@ append_trap() {
...
+
+# Usage: record_upgrade_prepare_done sc|wc
+# Records that upgrade has been prepared by storing a timestamp in the config and in the cluster along with the target
+# version.
+record_upgrade_prepare_done() {
+  local apps_config_timestamp
+  apps_config_timestamp="$(date -uIs)"
+
+  ts="${apps_config_timestamp}" \
+    yq_add common '.global.ck8sConfigSerial' 'strenv(ts)'
```
Elias-elastisys, line 445: What if you were to run upgrade twice for some reason? Would this not override the timestamp, and fail on the cm apply thus leaving you in a weird state, not allowing apply?

bin/ck8s
```diff
@@ -83,11 +85,14 @@ dry-run)
   ;;
 upgrade)
   [[ "${2}" =~ ^(wc|sc|both)$ ]] || usage
-  [[ "${3}" =~ ^(v[0-9]+\.[0-9]+)$ ]] || usage
-  [[ "${4}" =~ ^(prepare|apply)$ ]] || usage
+  [[ "${3}" =~ ^(v[0-9]+\.[0-9]+|unlock)$ ]] || usage
+  if [[ "${3}" != "unlock" ]]; then
+    [[ "${4}" =~ ^(prepare|apply)$ ]] || usage
+  fi
+  [[ "${2}" =~ ^(wc|sc|both)$ ]] || usage
```
Elias-elastisys, line 92: This line is here twice, seems unnecessary or?

Zash: Good catch, probably because this has gone trough a few revisions.

scripts/migration/lib.sh
Zash: Hm, yeah. I'll move it into the `if` block below, that should make more sense.

The prepare steps themselves should be fine to run again and there's the unlock command if nothing else.

scripts/migration/lib.sh
```diff
@@ -367,6 +407,112 @@ append_trap() {
   trap "$(new_trap)" "${signal}"
 }
 
+# usage: [[ "$(get_apps_version)" == "0.x" ]]
+get_apps_version() {
+  kubectl_do "${1}" get cm --namespace kube-system apps-meta --output=jsonpath --template="{.data.version}"
+}
+
+unlock_upgrade() {
+  kubectl_do "${1}" delete configmap --namespace kube-system apps-upgrade &>/dev/null ||
```
Zash, line 416: Perhaps this should remove `.global.ck8sConfigSerial` ?

Elias-elastisys: Maybe its interesting to be able to see when last upgrade was done just from config?

scripts/local-cluster.sh
```diff
@@ -250,6 +250,10 @@ resolve() {
             resolvectl dns "${link}" 127.0.64.43
           fi
         done
+      elif test -x /usr/sbin/unbound-control; then
```
aarnq, line 253:
```suggestion
      elif command -v unbound-control &>/dev/null; then
```

scripts/local-cluster.sh
```diff
@@ -262,6 +266,8 @@ resolve() {
     else
       if command -v resolvectl &>/dev/null && log.continue "restart systemd-resolved to reset dns servers?"; then
         systemctl restart systemd-resolved.service
+      elif test -x /usr/sbin/unbound-control && log.continue "remove forward for ${domain}?"; then
```
aarnq, line 269:
```suggestion
      elif command -v unbound-control &>/dev/null && log.continue "remove forward for ${domain}?"; then
```

scripts/local-cluster.sh
Zash: While it is sub-optimal to depend on the tool being in a specific location, it also might not be in `$PATH` when running as a regular user. 🤔

Zash: Perhaps `command -v || [[ -x /usr/sbin/... ]]` ? "Why not both" 😁

bin/ck8s
Zash: f59273f892057bb15b5f9097f07397a53349613a

scripts/migration/lib.sh
Zash: Will leave as is then.

scripts/migration/lib.sh
Zash: dde8bc9bd2e6a0ab4cecf45f605cadac2dcdfa69 does this look sensible?

bin/apply.bash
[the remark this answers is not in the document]
Zash: Does 0d27cd5ef2d26e2a8176338e16bbd38edae865da look like it resolves this?

scripts/migration/lib.sh
Elias-elastisys: Yes!

scripts/migration/lib.sh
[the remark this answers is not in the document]
Zash: Will have to think about this some. Obviously the in-cluster version won't exist before this is merged, so there has to be a before → after transition or migration somewhere.

But then this is the pre-migration in-cluster version. It uses the target version for the checks during the migration itself, so I think the pre-migration in-cluster version is only used at this point.

scripts/local-cluster.sh
Zash: So how about ff6b1062d384ace1be0115942f119fda41509c90

Otherwise I am not sure how to deal with the sudoers `secure_path` setting that is used in some places to override the `$PATH` you get when running `sudo`

bin/apply.bash
```diff
@@ -49,6 +55,14 @@ apps_apply() {
 
   if (with_kubeconfig "${config["kube_config_$1"]}" helmfile -f "${here}/../helmfile.d/state.yaml" -e "$2" "${action}" "${concurrency}" "${suppress:-}"); then
     log_info "---"
+    local current_version
+    current_version="$(get_apps_version "${2}" 2>/dev/null || true)"
+    if [ -n "${current_version}" ]; then
+      log_info "Current version is ${current_version}"
+    else
+      set_apps_version "${2}" "$(get_repo_version)"
+      log_info "Current version is ${current_version}"
```
aarnq, line 64: The else will not print the version.

scripts/local-cluster.sh
aarnq: I'm fine with `command -v || [[ -x /usr/sbin/... ]]`.

scripts/migration/lib.sh
```diff
@@ -281,6 +281,43 @@ check_version() {
     fi
   fi
 
+  # Make sure **config** and **cluster** is on the same (old) version
+  cluster_version="$(get_apps_version "${1}" >/dev/null 2>&1 || true)"
+  if [[ "${2}" == "prepare" ]]; then
+    if [[ "" != "${cluster_version}" ]] && [[ "${cluster_version}" != "${VERSION["${1}-config"]%.*}" ]]; then
```
aarnq, line 287:
```suggestion
    if [[ -n "${cluster_version}" ]] && [[ "${cluster_version}" != "${VERSION["${1}-config"]%.*}" ]]; then
```

scripts/migration/lib.sh
```diff
@@ -281,6 +281,43 @@ check_version() {
     fi
   fi
 
+  # Make sure **config** and **cluster** is on the same (old) version
+  cluster_version="$(get_apps_version "${1}" >/dev/null 2>&1 || true)"
+  if [[ "${2}" == "prepare" ]]; then
```
aarnq, line 286: If we are in an apply do we not want to do some validation as well here?

.github/workflows/tests.yml
```diff
@@ -90,7 +90,7 @@ jobs:
         run: make run-regression/${{ matrix.tag }}
       - name: delete local resolve
         if: always()
-        run: ./scripts/local-cluster.sh resolve delete
+        run: ./scripts/local-cluster.sh resolve delete test.dev-ck8s.com
```
simonklb, line 93: This looks unrelated, could you explain why this is relevant for this change?

.github/workflows/tests.yml
```diff
@@ -136,7 +136,7 @@ jobs:
         run: make run-integration/${{ matrix.tag }}
       - name: delete local resolve
         if: always()
-        run: ./scripts/local-cluster.sh resolve delete
+        run: ./scripts/local-cluster.sh resolve delete test.dev-ck8s.com
```
simonklb, line 139: Also looks unrelated

bin/init.bash
```diff
@@ -328,7 +328,7 @@ backup_file() {
 
   log_info "Creating backup ${backup_config_path}/${backup_name}"
 
-  cp "${file}" "${backup_config_path}/${backup_name}"
+  cp -b "${file}" "${backup_config_path}/${backup_name}"
```
simonklb, line 331: Also looks unrelated

scripts/local-cluster.sh
```diff
@@ -250,6 +250,8 @@ resolve() {
             resolvectl dns "${link}" 127.0.64.43
           fi
         done
+      elif command -v unbound-control &>/dev/null; then
+        sudo unbound-control forward_add +i test.dev-ck8s.com 127.0.64.43
```
simonklb, line 254: Also looks unrelated

scripts/local-cluster.sh
```diff
@@ -262,6 +264,8 @@ resolve() {
     else
       if command -v resolvectl &>/dev/null && log.continue "restart systemd-resolved to reset dns servers?"; then
         systemctl restart systemd-resolved.service
+      elif command -v unbound-control &>/dev/null; then
+        sudo unbound-control forward_remove +i test.dev-ck8s.com
```
simonklb, line 268: Also looks unrelated

scripts/local-clusters/configs/secrets.yaml
```diff
@@ -2,3 +2,8 @@ objectStorage:
   s3:
     accessKey: minioaccesskey
     secretKey: miniosecretkey
+alerts:
+  opsGenie:
+    apiKey: not-used
+  slack:
+    apiUrl: not-used
```
simonklb, line 9: Also looks unrelated

tests/common/bats/env.bash
```diff
@@ -106,6 +106,9 @@ env.init() {
     yq.set 'sc' '.alerts.opsGenieHeartbeat.name' '"example-heartbeat-name"'
   fi
 
+  yq.set 'secrets' '["alerts"]["opsGenie"]["apiKey"]' '"ZXhhbXBsZS1hcGlrZXkx"'
+  yq.set 'secrets' '["alerts"]["slack"]["apiUrl"]' '"https://slack.example/channel/ZXhhbXBsZS1jaGFubmVs"'
+
```
simonklb, line 111: Also looks unrelated

tests/common/bats/env.bash
```diff
@@ -114,6 +117,8 @@ env.init() {
   mkdir -p "${CK8S_CONFIG_PATH}/.state"
   touch "${CK8S_CONFIG_PATH}/.state/kube_config_sc.yaml"
   touch "${CK8S_CONFIG_PATH}/.state/kube_config_wc.yaml"
+
+  mkdir -m 777 "${CK8S_CONFIG_PATH}/backups"
```
simonklb, line 121: Also looks unrelated

simonklb (commented):
Commented on a few changes that seemed unrelated to the original issue. Please squash into separate commits or create separate PRs to make it easier to review.

bin/init.bash
Zash: Was a workaround for some issue in tests, hopefully resolved by not running them concurrently

tests/common/bats/env.bash
Zash: gone

tests/common/bats/env.bash
Zash: I think this was needed by the tests because otherwise it complains about "set-me". Do you remember @aarnq ?

scripts/local-clusters/configs/secrets.yaml
Zash: Or was this enough that the `yq.set` were not needed? 64e5c5d73b321c5f73d70e11fd47eba340cbc1e9 made it seem so

scripts/migration/lib.sh
Zash: What validation would we run here for apply?

tests/common/bats/env.bash
simonklb: Feels like it should have failed before this as well then? It's not a biggie though but curious.

tests/integration/general/upgrade.bats
```diff
@@ -0,0 +1,186 @@
...
+  name: apps-meta
+  namespace: kube-system
+EOF
+
+  # downgrade
+  gitversion.mock_static "v0.41.0"
+  run ck8s apply sc
+  assert_failure
+  assert_output --partial "Version mismatch. Run upgrade to update config."
+}
```
simonklb, line 166: The test title is "prevent apply using older config" but it looks like it's testing when the config is newer (v0.42.0) and the cluster is older (v0.41)?

tests/common/bats/env.bash
aarnq: It is because the migration lib (which hasn't been covered in tests before) is stricter than bin scripts in that it doesn't like somelongsecrets.

scripts/migration/lib.sh
aarnq: What validation is done up to this part in the apply case?

bin/apply.bash
aarnq: Then you can move the log out of the if.

tests/unit/general/bin-version.bats
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bats
```
simonklb, line 1: I think these tests would be more worth-while with mocks so that they actually showed versions in the output and not just the partials.

simonklb (changes requested):
Overall looks really good! A few comments that left me thinking there might be work still yet to be done or just leftovers that we have decided to skip for now.

Also some minor things in the tests but the coverage looks great so I think once the final touch ups are in place this is ready to go!

Amazing work, I know it's been quite a ride! :smile:

scripts/migration/lib.sh
```diff
@@ -283,6 +283,43 @@ check_version() {
...
+        if [[ "${reply}" != "y" ]]; then
+          exit 1
+        fi
+      else
+        # Don't automatically proceed
+        exit 1
+      fi
+    fi
+
+    # Make sure that **.global.ck8sVersion** is not updated until prepare is done
```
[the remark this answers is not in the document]
Zash: Leftover of an earlier misunderstanding I think.

`.global.ck8sVersion` is set by one of the `prepare` steps

```suggestion
    # line intentionally left blank 
```

tests/integration/general/upgrade.bats
```diff
@@ -0,0 +1,186 @@
...
+  run yq -i '.global.ck8sVersion="v0.42.0"' "${CK8S_CONFIG_PATH}/defaults/common-config.yaml"
+  ck8s ops kubectl sc apply -f - <<EOF
+apiVersion: v1
+data:
+  version: v0.41
+kind: ConfigMap
+metadata:
+  name: apps-meta
+  namespace: kube-system
+EOF
```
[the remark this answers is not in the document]
Zash: replaced by patching the existing one to pretend we upgraded

scripts/migration/lib.sh
```diff
@@ -283,6 +283,43 @@ check_version() {
...
+  fi
+
+  # **Config** must be previous minor or patch in the same minor
+  # The following?
```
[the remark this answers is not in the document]
Zash: removed

bin/ck8s
Zash: `ck8s version` can now show version from config and from cluster(s)

tests/integration/general/upgrade.bats
simonklb: I don't think the configmap patch makes any difference in that test. The error message you get is from the version in the config file mismatches from the git repository.

If you actually want to test the apply failing because of the diff in the configmap the configuration file needs to match v0.41 as well, no?

tests/integration/general/upgrade.bats
```diff
@@ -0,0 +1,186 @@
...
+# not:
+#   ck8s apply
+#   ck8s init ?
+#
+# ck8s upgrade apply
+# next:
+#   ck8s apply ?
+#
+# any other combo must throw error
+
```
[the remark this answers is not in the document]
simonklb: I think moving it to the README makes sense. I wonder if a state diagram would make more sense instead of a relational diagram between the commands though?

scripts/migration/lib.sh
```diff
@@ -283,6 +283,48 @@ check_version() {
...
+      fi
+    fi
+    ;;
+    # apply)
+    # In this case, what do we know?
+    # git: new version
+    # config: new version
+    # cluster: old version
+    # cluster: new version
+    # ;;
```
simonklb, line 325: Is this left for the future? If so at least write a TODO comment explaining what needs to be done. An issue would be even better.

tests/integration/general/upgrade.bats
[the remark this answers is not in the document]
simonklb: Just to be clear, this is completely optional. No need to do anything with this now or ever.

tests/integration/general/upgrade.bats
simonklb: I think this still needs to be addressed before I can approve.

scripts/migration/lib.sh
```diff
@@ -369,6 +411,112 @@ append_trap() {
   trap "$(new_trap)" "${signal}"
 }
 
+# usage: [[ "$(get_apps_version)" == "0.x" ]]
+get_apps_version() {
+  kubectl_do "${1}" get cm --namespace kube-system apps-meta --output=jsonpath --template="{.data.version}"
```
simonklb, line 416:
```suggestion
  kubectl_do "${1}" get configmap --namespace kube-system apps-meta --output=jsonpath --template="{.data.version}"
```

scripts/migration/lib.sh
```diff
@@ -369,6 +411,112 @@ append_trap() {
...
+}
+
+unlock_upgrade() {
+  kubectl_do "${1}" delete configmap --namespace kube-system apps-upgrade &>/dev/null ||
+    log_warn "Could not unlock upgrade or upgrade not locked"
+}
+
+# Get currently prepared version
+get_prepared_version() {
+  kubectl_do "${1}" get cm --namespace kube-system apps-upgrade --output=jsonpath --template="{.data.version}"
```
simonklb, line 426:
```suggestion
  kubectl_do "${1}" get configmap --namespace kube-system apps-upgrade --output=jsonpath --template="{.data.version}"
```

scripts/migration/lib.sh
```diff
@@ -369,6 +411,112 @@ append_trap() {
...
+  apps_config_timestamp="$(yq '.data.timestamp' <<<"${apps_upgrade}")"
+  apps_cluster_timestamp="$(yq_dig common '.global.ck8sConfigSerial')"
+  if [[ "${apps_config_timestamp}" != "${apps_cluster_timestamp}" ]]; then
+    log_fatal "Config timestamp mismatch, ${apps_cluster_timestamp} in ${1} but ${apps_config_timestamp} in config"
+  fi
+}
+
+# Usage: record_upgrade_apply_step sc|wc step-description
+record_upgrade_apply_step() {
+  # Was this needed?
```
simonklb, line 491: Was it? :smile:

simonklb (changes requested):
Still a few things left to do but it's very close now! :smile: 

I would also like that the new configmaps `apps-upgrade` and `apps-meta` as well as the new configuration option `ck8sConfigSerial` was described somewhere, perhaps `migration/README.md`?

scripts/migration/lib.sh
Zash: _Need_ is such a strong word. I was wondering if there was consensus on whether to keep or remove this. Easy enough to remove later anyway. Wrote a more descriptive comment.

scripts/migration/lib.sh
Zash: These are mostly notes for myself for something @aarnq wrote in another comment, but I wasn't sure yet what needed to be done.

tests/integration/general/upgrade.bats
Zash: Hm, right. How would you write the test?

scripts/migration/lib.sh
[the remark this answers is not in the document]
aarnq: I think at its current state this is fine if we add a warning message (to let admins know version checks is reduce if it is missing), and add an admin notice that this warning will pop up to the next version but as the version will be added the warning should not return as long as one is on a proper release tag.

scripts/migration/lib.sh
aarnq: I have reviewed the checks and I think they are fine, I have one suggestion on moving the ensure prepared to gather them in the same version check function, and to make it fail faster should it not be prepared.

bin/apply.bash
```diff
@@ -49,6 +55,17 @@ apps_apply() {
...
+    current_version="$(get_apps_version "${2}" 2>/dev/null || true)"
+    if [ -z "${current_version}" ]; then
+      current_version="$(get_repo_version)"
+      if [[ "${current_version}" == v* ]]; then
+        set_apps_version "${2}" "${current_version}"
+      else
+        current_version="git"
+      fi
+    fi
+    log_info "Current version is ${current_version}"
```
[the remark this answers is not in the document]
Zash: The issue was mostly about migrations so I didn't give much thought to e.g. following `main` or non-release branches. There should be some behavior corresponding to the `.global.ck8sVersion == "any"` presumably, if that isn't how it behaves already.

scripts/migration/lib.sh
```diff
@@ -283,6 +283,48 @@ check_version() {
...
+        log_warn_no_newline "Do you want to continue [y/N]: "
+        read -r reply
+        if [[ "${reply}" != "y" ]]; then
+          exit 1
+        fi
+      else
+        # Don't automatically proceed
+        exit 1
+      fi
+    fi
```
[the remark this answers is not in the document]
aarnq: If you link to specific commits please ensure that you don't force push them away, if I see that commits isn't contained in the branch then I won't look at them.

bin/common.bash
```diff
@@ -334,14 +335,28 @@ validate_version() {
...
+    log_error "Welkin Apps config version: ${ck8s_version}"
+    log_error "Welkin Apps repository version: ${version}"
+    exit 1
+  fi
+  if [[ -z "${cluster_version}" ]]; then
+    log_warning "Welkin Apps cluster version:     Unknown"
+    log_warning "Welkin Apps config version:      ${ck8s_version}"
+    log_warning "Welkin Apps repository version:  ${version}"
+  elif [[ "${cluster_version}" != "${version%.*}" ]]; then
+    # TODO double check whether patch version is included here
```
aarnq, line 355:
```suggestion
    # TODO double check whether patch version is included here
```
Is this still needed?

scripts/migration/lib.sh
```diff
@@ -283,6 +283,49 @@ check_version() {
...
+    ;;
+  apply)
+    # In this case, what do we know?
+    # git: new version
+    # config: new version
+    # cluster: old version
+    # cluster: new version
```
aarnq, line 324:
```suggestion
```

bin/common.bash
Zash: Would be good to check an extra time if all the `$version_*` variables include the patch version or not, so it's consistent.

scripts/migration/lib.sh
```diff
@@ -283,6 +283,44 @@ check_version() {
...
+        fi
+      else
+        # Don't automatically proceed
+        exit 1
+      fi
+    fi
+
+    # Make sure **config** and **cluster** is on the same (old) version
+    if [[ -n "${cluster_version}" ]] && [[ "${cluster_version}" != "${VERSION["${1}-config"]%.*}" ]]; then
+      log_warn "Version mismatch, cluster ${cluster_version}, config ${VERSION["${1}-config"]%.*}"
```
aarnq, line 306: I'm still missing the warning message here for when there is no cluster version.

Zash: Oh, thought I did that, where did that change go?

Oh, in `common.bash` https://github.com/elastisys/compliantkubernetes-apps/pull/2399/commits/4539515ef5ea206eed5d047bc31043cbd33f3149 I'll do it here too.

Zash: How about d4bcbec9 ?

bin/common.bash
```diff
@@ -334,14 +335,27 @@ validate_version() {
...
+  fi
+  if [[ -z "${cluster_version}" ]]; then
+    log_warning "Welkin Apps cluster version:     Unknown"
+    log_warning "Welkin Apps config version:      ${ck8s_version}"
+    log_warning "Welkin Apps repository version:  ${version}"
+  elif [[ "${cluster_version}" != "${version%.*}" ]]; then
+    log_error "ERROR: Version mismatch. Run upgrade to update cluster."
+    log_error "Welkin Apps cluster version:    ${cluster_version}"
+    log_error "Welkin Apps config version:     ${ck8s_version}"
+    log_error "Welkin Apps repository version: ${version}"
```
aarnq, line 358:
```suggestion
    log_error "ERROR: Version mismatch. Run upgrade to update config."
    log_error "Welkin Apps cluster version:    ${cluster_version}"
    log_error "Welkin Apps config version:     ${ck8s_version}"
    log_error "Welkin Apps repository version: ${version}"
    exit 1
  fi
  if [[ -z "${cluster_version}" ]]; then
    log_warning "Welkin Apps cluster version:    Unknown"
    log_warning "Welkin Apps config version:     ${ck8s_version}"
    log_warning "Welkin Apps repository version: ${version}"
  elif [[ "${cluster_version}" != "${version%.*}" ]]; then
    log_error "ERROR: Version mismatch. Run upgrade to update cluster."
    log_error "Welkin Apps cluster version:    ${cluster_version}"
    log_error "Welkin Apps config version:     ${ck8s_version}"
    log_error "Welkin Apps repository version: ${version}"
```
Nit

aarnq (approved):
A nit then I'm happy

Zash:
Phew, finally. Thanks for reviewing :)

simonklb:
image

PR #37, opened 2020-11-06
Upgrade Harbor chart to version 1.5.1

danielharr:
**What this PR does / why we need it**:
Upgrades Harbor to version 1.5.1.
This is required to access the `caBundleSecretName` Helm value, which is required to configure trust for a private CA.

**Which issue this PR fixes**:
Fixes #36 

**Public facing documentation PR** *(if applicable)*

**Special notes for reviewer**:
Tested an upgrade which succeeded. `ck8s test sc` passes as well.
Have not done any more extensive validation of Harbor functionality.

**Checklist:**

- [x] Added relevant notes to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md)
- [x] Proper commit message prefix on all commits
- [ ] Updated the [public facing documentation](https://github.com/elastisys/compliantkubernetes)
- Is this changeset backwards compatible for existing clusters? Applying:
    - [x] is completely transparent, will not impact the workload in any way.
    - [ ] requires running a migration script.
    - [ ] will create noticeable cluster degradation.
          E.g. logs or metrics are not being collected or Kubernetes API server
          will not be responding while upgrading.
    - [ ] requires draining and/or replacing nodes.
    - [ ] will change any APIs.
          E.g. removes or changes any CK8S config options or Kubernetes APIs.
    - [ ] will break the cluster.
          I.e. full cluster migration is required.

PR #2405, opened 2025-01-21
apps: Safeguard to prevent accidental cluster deletion

Eliastisys:
> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Adds a new gatekeeper policy that protects the resources of kinds `Cluster`, `OpenStackCluster` and `AzureCluster` from deletion. To delete the resources one must add the annotation `elastisys.io/ok-to-delete=""`

Since we don't ever want to get into the situation that we by accident delete a cluster, we want to have a gatekeeper policy to enforce this. This pr makes the `gatekeeper-constraints` chart install in sc (was previously only wc) and disables all constraints in sc config except for this newly created one.

- Fixes: [#37](https://github.com/elastisys/ck8s-cluster-api/issues/37)

#### Information to reviewers

Here's what the safeguard looks like in action:

```bash
➜ k delete cluster -n capi-cluster eliash-wc --dry-run=server
Error from server (Forbidden): admission webhook "validation.gatekeeper.sh" denied the request: [elastisys-prevent-accidental-deletion] Cluster deletion is not allowed.
To bypass the constraint, run:
kubectl annotate Cluster -n capi-cluster eliash-wc elastisys.io/ok-to-delete=""

➜ kubectl annotate Cluster -n capi-cluster eliash-wc elastisys.io/ok-to-delete=""
cluster.cluster.x-k8s.io/eliash-wc labeled

➜ k delete cluster -n capi-cluster eliash-wc --dry-run=server         
cluster.cluster.x-k8s.io "eliash-wc" deleted (server dry run)
```

@elastisys/goto-cluster-api do we want `capi terminate` to encounter this or should i include the labling part in that command?

#### Checklist

- Change checks:
    - [x] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
    - [x] The change does not cause any Pods to be blocked by Pod Security Standards or Policies
- NetworkPolicy checks:
    - [-] Any changed Pod is covered by Network Policies
    - [x] The change does not cause any dropped packets in the NetworkPolicy Dashboard
- Audit checks:
    - [x] The change requires changes to Kubernetes audit policy
- Falco checks:
- Bug checks:

### DIFF
--- a/helmfile.d/charts/gatekeeper/templates/policies/prevent-accidental-deletion.rego
+++ b/helmfile.d/charts/gatekeeper/templates/policies/prevent-accidental-deletion.rego
@@ -1,0 +1,38 @@
+package k8spreventaccidentaldeletion
+
+import future.keywords.in
+
+violation[{"msg": msg}] {
+
+    input.review.operation == "DELETE"
+    not correct_annotation
+    correct_kind
+
+    msg := sprintf(
+        "%v deletion is not allowed.\nTo bypass the constraint, run:\nkubectl annotate %v -n %v %v %v=anything",
+        [
+            input.review.object.kind,
+            input.review.object.kind,
+            input.review.object.metadata.namespace,
+            input.review.object.metadata.name,
+            input.parameters.annotation
+        ]
+    )
+}
+
+correct_annotation {
+
+    annotations := input.review.object.metadata.annotations
+    value := input.parameters.annotation
+
+    some key, _ in annotations
+    key == value
+}
+
+correct_kind {
+
+    kind := input.review.object.kind
+    kinds := input.parameters.kinds
+
+    kinds[_] == kind
+}
--- a/helmfile.d/charts/gatekeeper/templates/policies/tests/prevent-accidental-deletion.rego
+++ b/helmfile.d/charts/gatekeeper/templates/policies/tests/prevent-accidental-deletion.rego
@@ -1,0 +1,162 @@
+package k8spreventaccidentaldeletion
+
+import future.keywords.every
+
+prevented_kinds := ["Cluster", "OpenStackCluster", "AzureCluster"]
+
+test_delete_resource_without_annotation_is_denied {
+    every i in prevented_kinds {
+        count(violation) == 1 with input as {
+            "review": {
+                "operation": "DELETE",
+                "object": {
+                    "apiVersion": "*",
+                    "kind": i,
+                    "metadata": {
+                        "name": "test-resource",
+                        "namespace": "default",
+                        "annotations": {}
+                    }
+                }
+            },
+            "parameters": {
+                "annotation": "elastisys.io/ok-to-delete",
+                "kinds": prevented_kinds
+            }
+        }
+    }
+}
+
+test_delete_resource_with_annotation_is_allowed {
+    every i in prevented_kinds {
+        count(violation) == 0 with input as {
+            "review": {
+                "operation": "DELETE",
+                "object": {
+                    "apiVersion": "*",
+                    "kind": i,
+                    "metadata": {
+                        "name": "test-resource",
+                        "namespace": "default",
+                        "annotations": {
+                            "elastisys.io/ok-to-delete": "anything"
+                        }
+                    }
+                }
+            },
+            "parameters": {
+                "annotation": "elastisys.io/ok-to-delete",
+                "kinds": prevented_kinds
+            }
+        }
+    }
+}
+
+test_non_delete_operation_without_annotation_is_allowed {
+    every i in prevented_kinds {
+        count(violation) == 0 with input as {
+            "review": {
+                "operation": "UPDATE",
+                "object": {
+                    "apiVersion": "*",
+                    "kind": i,
+                    "metadata": {
+                        "name": "test-resource",
+                        "namespace": "default",
+                        "annotations": {}
+                    }
+                }
+            },
+            "parameters": {
+                "annotation": "elastisys.io/ok-to-delete",
+                "kinds": prevented_kinds
+            }
+        }
+    }
+}
+
+test_without_annotations_field {
+    count(violation) == 1 with input as {
+        "review": {
+            "operation": "DELETE",
+            "object": {
+                "apiVersion": "*",
+                "kind": "Cluster",
+                "metadata": {
+                    "name": "test-resource",
+                    "namespace": "default"
+                }
+            }
+        },
+        "parameters": {
+            "annotation": "elastisys.io/ok-to-delete",
+            "kinds": prevented_kinds
+        }
+    }
+}
+
+test_delete_other_kind_is_allowed {
+    count(violation) == 0 with input as {
+        "review": {
+            "operation": "DELETE",
+            "object": {
+                "apiVersion": "*",
+                "kind": "Pod",
+                "metadata": {
+                    "name": "test-pod",
+                    "namespace": "default",
+                }
+            }
+        },
+        "parameters": {
+            "annotation": "elastisys.io/ok-to-delete",
+            "kinds": prevented_kinds
+        }
+    }
+}
+
+test_some_other_annotation {
+    count(violation) == 0 with input as {
+        "review": {
+            "operation": "DELETE",
+            "object": {
+                "apiVersion": "*",
+                "kind": "Cluster",
+                "metadata": {
+                    "name": "test-resource",
+                    "namespace": "default",
+                    "annotations": {
+                        "elastisys.io/something-else": "test"
+                    }
+                }
+            }
+        },
+        "parameters": {
+            "annotation": "elastisys.io/something-else",
+            "kinds": prevented_kinds
+        }
+    }
+}
+
+test_bad_annotation {
+    count(violation) == 1 with input as {
+        "review": {
+            "operation": "DELETE",
+            "object": {
+                "apiVersion": "*",
+                "kind": "Cluster",
+                "metadata": {
+                    "name": "test-resource",
+                    "namespace": "default",
+                    "annotations": {
+                        "elastisys.io/bad-input": "test"
+                    }
+                }
+            }
+        },
+        "parameters": {
+            "annotation": "elastisys.io/something-else",
+            "kinds": prevented_kinds
+        }
+    }
+}
--- a/helmfile.d/values/gatekeeper/constraints.yaml.gotmpl
+++ b/helmfile.d/values/gatekeeper/constraints.yaml.gotmpl
@@ -13,10 +13,16 @@ disallowedTags:
 rejectLoadBalancerService:
     enable: {{ .Values.opa.rejectLoadBalancerService.enabled }}
     enforcementAction: {{ .Values.opa.rejectLoadBalancerService.enforcement }}
+preventAccidentalDeletion:
+    enable: {{ .Values.opa.preventAccidentalDeletion.enabled }}
+    enforcementAction: {{ .Values.opa.preventAccidentalDeletion.enforcement }}
 {{- if eq .Environment.Name "workload_cluster" }}
 allowUserCRDs:
     enable: {{ .Values.gatekeeper.allowUserCRDs.enabled }}
     enforcementAction: {{ .Values.gatekeeper.allowUserCRDs.enforcement }}
+disallowLocalhostSeccomp:
+    enable: true
+    enforcementAction: deny
 minimumDeploymentReplicas:
     enable: {{ .Values.opa.minimumDeploymentReplicas.enabled }}
     enforcementAction: {{ .Values.opa.minimumDeploymentReplicas.enforcement }}
--- a/config/common-config.yaml
+++ b/config/common-config.yaml
@@ -884,6 +884,11 @@ opa:
     enabled: true
     enforcement: warn
 
+  ## Prevent accidental deletion
+  preventAccidentalDeletion:
+    enabled: true
+    enforcement: deny
+
 trivy:
   enabled: true
   # comma separated list of namespaces (or glob patterns) to be excluded from scanning
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -3685,6 +3685,29 @@ properties:
               deny: Deny actions violating the constraint.
               warn: Warn actions violating the constraint.
               dryrun: Dryrun actions violating the constraint.
+      preventAccidentalDeletion:
+        title: Safeguard Prevent Accidental Deletion
+        description: |-
+          Configure constraint to reject deletion of sensitive resources.
+        type: object
+        additionalProperties: false
+        properties:
+          enabled:
+            title: Safeguard Prevent Accidental Deletion
+            type: boolean
+            default: false
+          enforcement:
+            title: Safeguard Prevent Accidental Deletion Enforcement
+            type: string
+            default: deny
+            enum:
+              - deny
+              - warn
+              - dryrun
+            meta:enum:
+              deny: Deny actions violating the constraint.
+              warn: Warn actions violating the constraint.
+              dryrun: Dryrun actions violating the constraint.
       resourceRequests:
         title: Safeguard Resource Requests
         description: |-
--- a/helmfile.d/charts/gatekeeper/constraints/templates/disallow-localhost-seccomp/constraint.yaml
+++ b/helmfile.d/charts/gatekeeper/constraints/templates/disallow-localhost-seccomp/constraint.yaml
@@ -1,9 +1,10 @@
+{{- if .Values.disallowLocalhostSeccomp.enable -}}
 apiVersion: constraints.gatekeeper.sh/v1beta1
 kind: K8sDisallowLocalhostSeccomp
 metadata:
   name: elastisys-disallow-localhost-seccomp-profile
 spec:
-  enforcementAction: deny
+  enforcementAction: {{ .Values.disallowLocalhostSeccomp.enforcementAction }}
   match:
     kinds:
     - apiGroups:
@@ -36,3 +37,4 @@ spec:
         operator: NotIn
         values:
         - ingress-nginx
+{{- end -}}
--- a/helmfile.d/charts/gatekeeper/constraints/templates/prevent-accidental-deletion/constraint.yaml
+++ b/helmfile.d/charts/gatekeeper/constraints/templates/prevent-accidental-deletion/constraint.yaml
@@ -1,0 +1,23 @@
+{{- if .Values.preventAccidentalDeletion.enable -}}
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: K8sPreventAccidentalDeletion
+metadata:
+  name: elastisys-prevent-accidental-deletion
+spec:
+  enforcementAction: {{ .Values.preventAccidentalDeletion.enforcementAction }}
+  match:
+    kinds:
+      - apiGroups: ["cluster.x-k8s.io"]
+        kinds:
+          - Cluster
+      - apiGroups: ["infrastructure.cluster.x-k8s.io"]
+        kinds:
+          - OpenStackCluster
+          - AzureCluster
+  parameters:
+    annotation: elastisys.io/ok-to-delete
+    kinds:
+      - Cluster
+      - OpenstackCluster
+      - AzureCluster
+{{- end }}
--- a/helmfile.d/charts/gatekeeper/constraints/values.yaml
+++ b/helmfile.d/charts/gatekeeper/constraints/values.yaml
@@ -17,8 +17,14 @@ rejectLoadBalancerService:
     enable: true
     enforcementAction: dryrun
 minimumDeploymentReplicas:
-    enable: true
+    enable: false
     enforcementAction: warn
+preventAccidentalDeletion:
+    enable: true
+    enforcementAction: deny
+disallowLocalhostSeccomp:
+    enable: false
+    enforcementAction: deny
 
 imageRegistryURL: registry.example.com
 
--- a/helmfile.d/charts/gatekeeper/templates/templates/prevent-accidental-deletion/template.yaml
+++ b/helmfile.d/charts/gatekeeper/templates/templates/prevent-accidental-deletion/template.yaml
@@ -1,0 +1,26 @@
+apiVersion: templates.gatekeeper.sh/v1
+kind: ConstraintTemplate
+metadata:
+  name: k8spreventaccidentaldeletion
+spec:
+  crd:
+    spec:
+      names:
+        kind: K8sPreventAccidentalDeletion
+      validation:
+        # Schema for the `parameters` field
+        openAPIV3Schema:
+            type: object
+            properties:
+              annotation:
+                description: This annotation allows removal of sensitive resources protected from accidental deletion.
+                type: string
+              kinds:
+                description: A list of resource kinds that will be affected by the constraint.
+                type: array
+                items:
+                  type: string
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+{{ .Files.Get "policies/prevent-accidental-deletion.rego"  | indent 8 }}
--- a/helmfile.d/stacks/gatekeeper.yaml.gotmpl
+++ b/helmfile.d/stacks/gatekeeper.yaml.gotmpl
@@ -49,7 +49,6 @@ templates:
     chart: charts/gatekeeper/constraints
     version: 0.1.0
     name: gatekeeper-constraints
-    condition: ck8sWorkloadCluster.enabled
     needs:
       - gatekeeper-system/gatekeeper-templates # creates gatekeeper/constraints
     values:
--- a/helmfile.d/values/gatekeeper/gatekeeper.yaml.gotmpl
+++ b/helmfile.d/values/gatekeeper/gatekeeper.yaml.gotmpl
@@ -57,7 +57,10 @@ validatingWebhookCustomRules:
       - UPDATE
       - DELETE
     resources:
+      - azureclusters
       - customresourcedefinitions
+      - clusters
+      - openstackclusters
       - secrets
 mutatingWebhookCustomRules:
   - apiGroups:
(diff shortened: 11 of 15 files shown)


helmfile.d/charts/gatekeeper/constraints/templates/no-cluster-deletion/constraint.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.noClusterDeletion.enable -}}
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: K8sNoClusterDeletion
+metadata:
+  name: elastisys-no-cluster-deletion
+spec:
+  enforcementAction: {{ .Values.noClusterDeletion.enforcementAction }}
+  match:
+    kinds:
+      - apiGroups: ["*"]
```
davidumea, line 10: This can be limited right?
```suggestion
      - apiGroups: ["cluster.x-k8s.io"]
```

helmfile.d/charts/gatekeeper/constraints/templates/no-cluster-deletion/constraint.yaml
```diff
@@ -0,0 +1,15 @@
...
+        kinds:
+          - Cluster
+          - OpenStackCluster
+          - Secret
```
davidumea, line 14:
```suggestion
```

config/sc-config.yaml
```diff
@@ -1034,6 +1034,12 @@ opa:
   disallowedTags:
     enabled: false
 
+  disallowLocalhostSeccomp:
+    enabled: false
+
+  minimumDeploymentReplicas:
+    enabled: false
```
davidumea, line 1041: Any reason this was added to sc-config?

Eliastisys: Since gatekeeper-constraints wasn't previously installed in sc, these additions makes this pr only add the noclusterdeletion constraint to sc when installing the chart there.

simonklb (commented):
Before digging into the nitty-gritty:

While this is fine and do help prevent mistakenly deleting a cluster what we really _should_ do is to promote the habit of using credentials which doesn't have delete permissions in the first place. Using accounts with cluster-admin rights should be **last resort**.

Not a big fan of "no cluster deletion" as a policy name. Why not "prevent accidental cluster deletion", too long? At least that would be more descriptive. Maybe just "prevent accidental deletion" which allows us to add more sensitive resources in the future.

Add unit tests please!

config/sc-config.yaml
davidumea: Okay, but do they have to be config options? I don't think one would want to enable this for SC

helmfile.d/charts/gatekeeper/templates/policies/no-cluster-deletion.rego
```diff
@@ -0,0 +1,14 @@
...
+    kinds := {
+        "Cluster",
+        "OpenStackCluster"
+    }
+
+    input.review.object.kind == kinds[_]
+    input.review.operation == "DELETE"
+    not (input.review.object.metadata.labels["ok-to-delete"] == "true")
+
+    msg := sprintf("Cluster deletion is not allowed.\nTo bypass the constraint, run:\nkubectl label %v -n %v %v ok-to-delete=true",[input.review.object.kind, input.review.namespace, input.review.name])
```
viktor-f, line 13: I suggest that the label is possible to modify using a parameter. Like we have done for minimum replicas ([rego](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile.d/charts/gatekeeper/templates/policies/deployment-minimum-replicas.rego), [template](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile.d/charts/gatekeeper/templates/templates/deployment-minimum-replicas/template.yaml), [constraint](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile.d/charts/gatekeeper/constraints/templates/deployment-minimum-replicas/constraint.yaml)). Partially to keep the same kind of configurability for similar policies.
Similarly I would also suggest to change this into an annotation and add a `elastisys.io/` prefix to the default annotation name.
Do you think that sounds good?

Xartos:
> @elastisys/goto-cluster-api do we want capi terminate to encounter this or should i include the labling part in that command?

I think this should be something separate so that you don't by accident run terminate on a cluster you didn't intend to

> While this is fine and do help prevent mistakenly deleting a cluster what we really should do is to promote the habit of using credentials which doesn't have delete permissions in the first place. Using accounts with cluster-admin rights should be last resort.

While I do agree with you here, I think that we'd need to do some bigger initiative for this. Since if we were to implement this, we'd need to keep it very carefully up to date since it might hinder a troubleshooting of a cluster (especially on capi since bastion hosts might not be running for fetching admin kubeconfig) if something was missed such that we lost access to things.

helmfile.d/charts/gatekeeper/constraints/templates/no-cluster-deletion/constraint.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: elastisys-no-cluster-deletion
+spec:
+  enforcementAction: {{ .Values.noClusterDeletion.enforcementAction }}
+  match:
+    kinds:
+      - apiGroups: ["*"]
+        kinds:
+          - Cluster
+          - OpenStackCluster
```
viktor-f, line 13: We also support azure for cluster api. So I suggest we add this.

```yaml
      - apiGroups: ["infrastructure.cluster.x-k8s.io"]
        kinds:
          - azurecluster
```

helmfile.d/charts/gatekeeper/constraints/templates/no-cluster-deletion/constraint.yaml
viktor-f: It looks like `openstackcluster` is part of `infrastructure.cluster.x-k8s.io`. Does that then have to be in it's own list entry or would `cluster.x-k8s.io` capture that as well?

helmfile.d/charts/gatekeeper/constraints/templates/no-cluster-deletion/constraint.yaml
Xartos, line 13: Suggestion: I guess this also should be here?
```suggestion
          - OpenStackCluster
          - AzureCluster
```

If so, there's some more places it needs to be added

helmfile.d/charts/gatekeeper/templates/policies/no-cluster-deletion.rego
```diff
@@ -0,0 +1,14 @@
+package k8snoclusterdeletion
+
+violation[{"msg": msg}] {
+    kinds := {
+        "Cluster",
+        "OpenStackCluster"
+    }
+
+    input.review.object.kind == kinds[_]
```
Xartos, line 9: Question: My opa knowledge is lacking so bear with me. Do we need to validate the kind? Isn't this code called only on relevant objects?

migration/v0.43/apply/30-gatekeeper-crds.sh
```diff
@@ -0,0 +1,30 @@
...
+  execute)
+
+    if [[ "${CK8S_CLUSTER}" =~ ^(sc|both)$ ]]; then
+      log_info "installing gatekeeper crds in service cluster"
+      helmfile_do sc -l name=gatekeeper-templates apply --include-transitive-needs
+    fi
+    if [[ "${CK8S_CLUSTER}" =~ ^(wc|both)$ ]]; then
+      log_info "installing gatekeeper crds in workload cluster"
+      helmfile_do wc -l name=gatekeeper-templates apply --include-transitive-needs
+    fi
```
viktor-f, line 19: Do we need to have this in it's own migration script? Does the helmfile not make sure that we install templates before constraints? (I assume that the problem this is trying to avoid is installing constraints before the templates).
But if we need this, then I wonder if skipping `app!=gatekeeper` is correct in `80-apply.sh`. Since that would also skip installing any changes to the gatekeeper controller and the constraints. Could we either add a full install of gatekeeper (including the app and constraints) in this migration script or have the `80-apply.sh` only skip the templates?

simonklb:
> > While this is fine and do help prevent mistakenly deleting a cluster what we really should do is to promote the habit of using credentials which doesn't have delete permissions in the first place. Using accounts with cluster-admin rights should be last resort.
> 
> While I do agree with you here, I think that we'd need to do some bigger initiative for this. Since if we were to implement this, we'd need to keep it very carefully up to date since it might hinder a troubleshooting of a cluster (especially on capi since bastion hosts might not be running for fetching admin kubeconfig) if something was missed such that we lost access to things.

Yea absolutely, a "simple" initial step would be to have a two tier path where you first only use cluster-wide read-only credentials and then can escalate to cluster-admin.

@viktor-f @OlleLarsson worthy to get this ball rolling?

helmfile.d/charts/gatekeeper/constraints/templates/no-cluster-deletion/constraint.yaml
Eliastisys: I was thinking of going with what [Simon suggested](https://github.com/elastisys/compliantkubernetes-apps/pull/2405#pullrequestreview-2565043728) and rename the safeguard "prevent accidental deletion" which would allow adding more resources to it later. So I think keeping this as is would then make sense. What do you think?

helmfile.d/charts/gatekeeper/constraints/templates/no-cluster-deletion/constraint.yaml
Eliastisys: leftover from wc testing, thanks for spotting

helmfile.d/charts/gatekeeper/templates/policies/no-cluster-deletion.rego
Eliastisys: True! thanks

helmfile.d/charts/gatekeeper/constraints/templates/no-cluster-deletion/constraint.yaml
simonklb: I think it might still be worth narrowing it down otherwise it might trigger the webhook unnecessarily. Not sure if that is how the implementation actually works under the hood though.

viktor-f: I like the rewrite to make this more generic. But I still think it looks better to have several blocks of resource types with specific groups in this file (and as Simon says it might have some performance impact).

viktor-f:
> Yea absolutely, a "simple" initial step would be to have a two tier path where you first only use cluster-wide read-only credentials and then can escalate to cluster-admin.
> 
> @viktor-f @OlleLarsson worthy to get this ball rolling?

Yeah I think you can propose that as a roadmap item. @OlleLarsson do you agree?

config/sc-config.yaml
Eliastisys: removed them as sc config options: https://github.com/elastisys/compliantkubernetes-apps/pull/2405/commits/3ebe153d94529bca772990121e685135e591b05c

helmfile.d/charts/gatekeeper/templates/policies/no-cluster-deletion.rego
Xartos: This seems to still be here. Changed your mind or missed it?

helmfile.d/charts/gatekeeper/templates/policies/tests/prevent-accidental-deletion.rego
```diff
@@ -0,0 +1,207 @@
...
+                "apiVersion": "*",
+                "kind": "Pod",
+                "metadata": {
+                    "name": "test-pod",
+                    "namespace": "default",
+                }
+            }
+        },
+        "parameters": {
+            "annotation": "elastisys.io/ok-to-delete"
```
Xartos, line 204: NIT: To make sure that this actually is setting what annotation is used and that it's not hardcoded to `elastisys.io/ok-to-delete`. You can, for each test, have a different annotation. Alternatively create some other tests were it has another annotation

migration/v0.43/apply/30-gatekeeper-crds.sh
```diff
@@ -0,0 +1,30 @@
...
+# shellcheck source=scripts/migration/lib.sh
+source "${ROOT}/scripts/migration/lib.sh"
+
+run() {
+  case "${1:-}" in
+  execute)
+
+    if [[ "${CK8S_CLUSTER}" =~ ^(sc|both)$ ]]; then
+      log_info "installing gatekeeper crds in service cluster"
+      helmfile_do sc -l name=gatekeeper-templates apply --include-transitive-needs
```
Xartos, line 14: NIT: You can skip this one in the apply step. I.e. add `name!=gatekeeper-templates`. But since it's so small I'm fine with leaving it like this

Xartos (commented):
Some small stuff, but looks good in general

helmfile.d/values/gatekeeper/constraints.yaml.gotmpl
```diff
@@ -13,14 +13,21 @@ disallowedTags:
...
     enable: {{ .Values.gatekeeper.allowUserCRDs.enabled }}
     enforcementAction: {{ .Values.gatekeeper.allowUserCRDs.enforcement }}
+disallowLocalhostSeccomp:
+    enable: true
+    enforcementAction: deny
 minimumDeploymentReplicas:
-    enable: {{ .Values.opa.minimumDeploymentReplicas.enabled }}
-    enforcementAction: {{ .Values.opa.minimumDeploymentReplicas.enforcement }}
+    enable: true
+    enforcementAction: warn
```
viktor-f, line 28: Why was this changed?

helmfile.d/charts/gatekeeper/templates/policies/prevent-accidental-deletion.rego
```diff
@@ -0,0 +1,30 @@
...
+            input.review.object.metadata.namespace,
+            input.review.object.metadata.name,
+            input.parameters.annotation
+        ]
+    )
+}
+
+correct_annotation {
+    annotations := input.review.object.metadata.annotations
+    annotations[input.parameters.annotation] == "true"
```
viktor-f, line 29: Question, nit: in the minimum replicas policy we only check that the annotation is present and do not care about the value of the annotation. Here we also check that the value is "true". Is there any particular reason why this is different? For consistency I think it would look nice if these worked the same way.

helmfile.d/values/gatekeeper/constraints.yaml.gotmpl
Eliastisys: https://github.com/elastisys/compliantkubernetes-apps/pull/2405#discussion_r1923976415

helmfile.d/charts/gatekeeper/templates/policies/no-cluster-deletion.rego
Eliastisys: Yeah i changed my mind since the rego unit tests doesn't work without this.

helmfile.d/charts/gatekeeper/templates/policies/tests/prevent-accidental-deletion.rego
simonklb: Or actually hard code it until we find the need to have different annotations.

helmfile.d/charts/gatekeeper/templates/policies/prevent-accidental-deletion.rego
Eliastisys: https://github.com/elastisys/compliantkubernetes-apps/pull/2405/commits/c41aaeb7fffbca359b29ab5057dde15fb4b23fc1

helmfile.d/charts/gatekeeper/templates/policies/tests/prevent-accidental-deletion.rego
Eliastisys: https://github.com/elastisys/compliantkubernetes-apps/pull/2405/commits/562f8046b1185466fa55b40494820cf08f7d3ffc

migration/v0.43/apply/30-gatekeeper-crds.sh
Eliastisys: https://github.com/elastisys/compliantkubernetes-apps/pull/2405/commits/a7c859d1789773976605e54e2aca8ac9c40b0b94

helmfile.d/charts/gatekeeper/templates/policies/no-cluster-deletion.rego
viktor-f: I think that it would be ok if this is not in the rego. Since then the default policy is just about preventing deletion of any resource. Alternatively you take in the kinds as a parameter.

helmfile.d/values/gatekeeper/constraints.yaml.gotmpl
Eliastisys: https://github.com/elastisys/compliantkubernetes-apps/pull/2405/files/562f8046b1185466fa55b40494820cf08f7d3ffc..8783a3c84054cc64cdd533f3745892f984735e56

Eliastisys: I think I'm just confused by the constraints being "some always enabled in wc but never in sc, and some configurable in config"

helmfile.d/charts/gatekeeper/templates/policies/no-cluster-deletion.rego
viktor-f: If you go with the option of just not having it in the rego, then yes.
If you instead go with the option of passing the correct kinds via parameter, then test that parameter.
Either is ok with me :slightly_smiling_face:

Eliastisys: https://github.com/elastisys/compliantkubernetes-apps/pull/2405/commits/d36280fc9f01d1baabcf52931254d904d870010a

migration/v0.43/apply/30-gatekeeper-crds.sh
Xartos: Is `app!=gatekeeper-templates` the same as `name!=gatekeeper-templates`? Because you are applying using the `name` variant, so to keep it consistent I'd like to have the same for both

Eliastisys: it's not the same, was a typo, thanks for noticing! https://github.com/elastisys/compliantkubernetes-apps/pull/2405/commits/1c5dc93b7df82b5f335941593ba768bbf43d10d8

Xartos (approved):

helmfile.d/charts/gatekeeper/templates/policies/tests/prevent-accidental-deletion.rego
```diff
@@ -0,0 +1,240 @@
...
+            }
+        },
+        "parameters": {
+            "annotation": "elastisys.io/ok-to-delete",
+            "kinds": ["Cluster", "OpenStackCluster", "AzureCluster"]
+        }
+    }
+}
+
+test_without_annotations_field {
```
simonklb, line 156: Is this test not the same as `test_delete_cluster_without_annotation_is_denied`?

helmfile.d/values/gatekeeper/constraints.yaml.gotmpl
```diff
@@ -13,14 +13,21 @@ disallowedTags:
...
     enforcementAction: {{ .Values.gatekeeper.allowUserCRDs.enforcement }}
+disallowLocalhostSeccomp:
+    enable: true
+    enforcementAction: deny
 minimumDeploymentReplicas:
     enable: {{ .Values.opa.minimumDeploymentReplicas.enabled }}
     enforcementAction: {{ .Values.opa.minimumDeploymentReplicas.enforcement }}
 {{- end }}
 
+
```
simonklb, line 31:
```suggestion
```

migration/v0.43/apply/30-gatekeeper-crds.sh
simonklb, line 14: I assume we don't want to run with `--include-transitive-needs` here since that could mess up the `skipped` filter in the apply step, right? CC @aarnq

simonklb (changes requested):
Just a question on the migration step, otherwise this LGTM! Nice work :)

helmfile.d/charts/gatekeeper/templates/policies/tests/prevent-accidental-deletion.rego
Eliastisys: empty annotation field vs missing annotation field

migration/v0.43/apply/30-gatekeeper-crds.sh
Eliastisys: true, tested migration again and it works without the needs flag

Eliastisys: https://github.com/elastisys/compliantkubernetes-apps/pull/2405/commits/11ae11ce716211982beb6231c3a94c590f27e38d

migration/v0.43/apply/30-gatekeeper-crds.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
aarnq, line 1: This isn't targeting v0.43 any more.

Also is there really a need for this? Shouldn't gatekeeper-templates be upgraded before gatekeeper-constraints anyway?

helmfile.d/charts/gatekeeper/templates/policies/prevent-accidental-deletion.rego
```diff
@@ -0,0 +1,39 @@
...
+import future.keywords.in
+
+violation[{"msg": msg}] {
+
+    input.review.operation == "DELETE"
+    not correct_annotation
+    correct_kind
+
+    msg := sprintf(
+        "%v deletion is not allowed.\nTo bypass the constraint, run:\nkubectl annotate %v -n %v %v %v=anything",
```
aarnq, line 12: To me it feels strange that we want to have a `key=anything` but if that is what we have elsewhere then sure, but we should consider changing it.

migration/v0.43/apply/30-gatekeeper-crds.sh
Eliastisys: `app=gatekeeper apply` gives me:
```
  Error: Failed to render chart: exit status 1: Error: unable to build kubernetes objects from release manifest: resource mapping not found for name: "elastisys-prevent-accidental-deletion" namespace: "" from "": no matches for kind "K8sPreventAccidentalDeletion" in version "constraints.gatekeeper.sh/v1beta1"
  ensure CRDs are installed first
```

helmfile.d/charts/gatekeeper/templates/policies/prevent-accidental-deletion.rego
Eliastisys: Agree. Minimum replicas only looks at the key. It's included in the [public docs](https://elastisys.io/welkin/user-guide/safeguards/enforce-minimum-replicas/). That policy is also missing unit tests.

migration/v0.43/apply/30-gatekeeper-crds.sh
aarnq: And with the regular apply snippet?

helmfile.d/charts/gatekeeper/templates/policies/prevent-accidental-deletion.rego
```diff
@@ -0,0 +1,39 @@
...
+    key == value
+}
+
+correct_kind {
+
+    kind := input.review.object.kind
+    kinds := input.parameters.kinds
+
+    some i
+    kinds[i] == kind
```
viktor-f, line 38: Nit: I think this can be simplified to this:
```suggestion
    kinds[_] == kind
```

migration/v0.43/apply/30-gatekeeper-crds.sh
Eliastisys: @aarnq Same error with regular apply snippet

Eliastisys: moved migration to v0.44 https://github.com/elastisys/compliantkubernetes-apps/pull/2405/commits/9f3b438b07d84eda84146e75f6f4352adb4b1b89

viktor-f (approved):
Nice work!
Will you also look into adding the integration needed in capi so that we can easily add the correct labels when we actually want to delete a cluster?

Eliastisys:
> Nice work! Will you also look into adding the integration needed in capi so that we can easily add the correct labels when we actually want to delete a cluster?

Sure! Something like a force flag for terminate? @viktor-f  @elastisys/goto-cluster-api

simonklb:
What happens if you run `helm uninstall` btw? Do you get a nice error message telling you why the uninstall failed?

viktor-f:
> > Nice work! Will you also look into adding the integration needed in capi so that we can easily add the correct labels when we actually want to delete a cluster?
> 
> Sure! Something like a force flag for terminate? @viktor-f @elastisys/goto-cluster-api

As @Xartos wrote earlier it could be good if this was a separate command. To avoid accidentally writing the wrong cluster name.

Eliastisys:
> What happens if you run `helm uninstall` btw? Do you get a nice error message telling you why the uninstall failed?

Only a `Error: failed to delete release: example`. Running with `--debug` shows the nice error message.

migration/v0.43/apply/30-gatekeeper-crds.sh
aarnq: That is strange, the `helmfile_upgrade` should be take them in order to prevent that. I guess I'll have to review our helmfile setup.

viktor-f (commented):
Code looks good, nice work!

Sine this is on by default and actively changes things for the administrator I think it would make sense to add an admin notice for it. Could you add that in the PR description?

viktor-f (approved):

PR #2408, opened 2025-01-22
scripts: Start of new requirements installer

aarnq:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

This aim to develop a lightweight requirements installer that can be used in a container as well as on workstations, using the current PURL format.

The goal is to be able to serve the tests/Dockerfile with the following types: 
- deb packages
- helm plugins
- npm packages
- direct bin/tar/zip downloads with checksum verification

As much as possible is to be done in bash alone, debs are always installed first as that may install other package managers, but if needed they are expected to be installed before running the script.

The last type is provided by mapping, which isn't the greatest method, but the one that we kind of already use in the current ansible based approach.

Note that the purl parser does not attempt to be compliant, but tries to be enough for our purposes.

#### Information to reviewers

Currently I've tested to just run the script against the base requirement file when building a container.

We probably need to have a discussion around the purl types as we are using strange ways to refer to `golang` packages.
Logically it does not make sense for us to refer to `golang` packages, as we are not fetching `golang` packages. It should be `github` packages for a lot of these as we fetch them from their github releases.

I'll wait with implementing more mappings until we have discussed this.

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/scripts/requirements.sh
+++ b/scripts/requirements.sh
@@ -1,0 +1,365 @@
+#!/usr/bin/env bash
+
+# Requirements installer for containers
+# The goal is to make it usable as _the_ way to install requirements but we need to start somewhere.
+
+set -euo pipefail
+shopt -s extglob
+
+# purl variables
+declare -a packages
+declare -A types
+declare -A namespaces
+declare -A names
+declare -A versions
+declare -A qualifiers
+declare -A subpaths
+
+# TODO: Remove when actually used
+export qualifiers
+export subpaths
+
+# others mappings
+declare -A targets
+declare -A sources
+declare -A intermediates
+
+# <package> <target> <source> [intermediate]
+parse-mappings() {
+  local package="${1}" target="${2}" source="${3}" intermediate="${4:-}"
+
+  targets["${package}"]="${target}"
+  sources["${package}"]="${source}"
+  intermediates["${package}"]="${intermediate}"
+}
+
+# parse-mappings /usr/local/bin/gomplate "https://github.com/hairyhenderson/gomplate/releases/download/v\${version}/gomplate_linux-amd64"
+parse-mappings golang/helm.sh/helm/v3 /usr/local/bin/helm "https://get.helm.sh/helm-v\${version}-linux-amd64.tar.gz" linux-amd64/helm
+# helm plugin install https://github.com/databus23/helm-diff --version "v${HELM_DIFF_VERSION}" >/dev/null
+# helm plugin install https://github.com/jkroepke/helm-secrets --version "v${HELM_SECRETS_VERSION}" >/dev/null
+parse-mappings golang/github.com/helmfile/helmfile /usr/local/bin/helmfile "https://github.com/helmfile/helmfile/releases/download/v\${version}/helmfile_\${version}_linux_amd64.tar.gz" helmfile
+parse-mappings generic//kind /usr/local/bin/kind "https://github.com/kubernetes-sigs/kind/releases/download/v\${version}/kind-linux-amd64"
+parse-mappings generic//kubectl /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/v\${version}/bin/linux/amd64/kubectl"
+# parse-mappings /usr/local/bin/kubeconform "https://github.com/yannh/kubeconform/releases/download/v\${version}/kubeconform-linux-amd64.tar.gz" kubeconform
+# parse-mappings /usr/local/bin/kubectl-oidc_login "https://github.com/int128/kubelogin/releases/download/v\${version}/kubelogin_linux_amd64.zip" kubelogin
+# parse-mappings /usr/local/bin/opa "https://github.com/open-policy-agent/opa/releases/download/\${version}/opa_linux_amd64"
+parse-mappings golang/getsops/sops/v3 /usr/local/bin/sops "https://github.com/getsops/sops/releases/download/v\${version}/sops-v\${version}.linux.amd64"
+parse-mappings golang/github.com/neilpa/yajsv /usr/local/bin/yajsv "https://github.com/neilpa/yajsv/releases/download/v\${version}/yajsv.linux.amd64"
+parse-mappings golang/github.com/vmware-tanzu/velero /usr/local/bin/velero "https://github.com/vmware-tanzu/velero/releases/download/v\${version}/velero-v\${version}-linux-amd64.tar.gz" "velero-v\${version}-linux-amd64/velero"
+parse-mappings golang/github.com/mikefarah/yq/v4 /usr/local/bin/yq4 "https://github.com/mikefarah/yq/releases/download/v\${version}/yq_linux_amd64"
+
+# <file>
+parse-requirements() {
+  local file="${1}"
+
+  if ! [[ -f "${file}" ]]; then
+    echo "failure, not a file"
+    return 1
+  fi
+
+  local -a rows
+  readarray rows <"${1}"
+
+  for row in "${rows[@]}"; do
+    parse-package "${row}"
+  done
+}
+
+# <purl>
+parse-package() {
+  local purl="${1}" remainder component
+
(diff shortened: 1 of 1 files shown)


simonklb:
The reason why I picked the `golang` PURL type was because vulnerabilities are usually reported for the go modules.

scripts/requirements.sh
```diff
@@ -0,0 +1,365 @@
...
+
+# <target> <source> <checksum-bin>
+install-from-bin() {
+  local target="${1}" source="${2}" checksum_bin="${4:-}" checksum
+
+  echo "installing ${target}"
+
+  if [[ -e "${target}" ]] && [[ -n "${checksum_bin}" ]]; then
+    checksum="$(sha256sum "${target}")"
+    if [[ "${checksum_bin}" == "${checksum%%" "*}" ]]; then
```
Zash, line 151:
```suggestion
    if sha256sum --check - <<< "${checksum_bin}  ${target}" >/dev/null 2>/dev/null; then
```

Could save a tiny bit of string wrangling, but should otherwise be equivalent.

aarnq:
> The reason why I picked the `golang` PURL type was because vulnerabilities are usually reported for the go modules.

Sure but if that is going to be a consideration then we should have a at least a plan to take advantage of that, so please create an issue for it else I don't see why we should have roundabout ways to manage it.
And if that is going to be a requirement then many purls are mismatched against OSV.
Would there be any other sources we want to query purls against?

As an alternative, what we have talked about is for the requirement script to be able to download newer versions and at release create a pinned requirements listing. Would it make sense in a such requirements list to write `golang` purls as an additional alias for that type of use?
Then we could allow the "floating" requirements to have more potential _non standard shenanigans_ to make the fetching easier for us, and the pinned one to be as compliant as we can make it.

Zash:
Looks promising!

aarnq:
> > > The reason why I picked the `golang` PURL type was because vulnerabilities are usually reported for the go modules.
> > 
> > 
> > Sure but if that is going to be a consideration then we should have a at least a plan to take advantage of that, so please create an issue for it else I don't see why we should have roundabout ways to manage it.
> 
> My understanding was that [this issue](https://github.com/elastisys/ck8s-issue-tracker/issues/410) would cover that.

Well, there was nothing there even suggesting that it should be implemented using queries on purls rather than our standard method of subscribing to security notices.
So I want added it. 

> > And if that is going to be a requirement then many purls are mismatched against OSV. Would there be any other sources we want to query purls against?
> 
> All the PURLs ideally. But matching CVEs against PURLs isn't a straight forward job, talking from experience from previous engagements... 😄

Note, I'm asking for _sources to query against_ so I can evaluate which purls to use before I create specific mappings that are going to replaced, not which purls to query as I assume we want that for everything. And as you have previous experience, I thought you might have some suggestions. :wink:

> > As an alternative, what we have talked about is for the requirement script to be able to download newer versions and at release create a pinned requirements listing. Would it make sense in a such requirements list to write `golang` purls as an additional alias for that type of use? Then we could allow the "floating" requirements to have more potential _non standard shenanigans_ to make the fetching easier for us, and the pinned one to be as compliant as we can make it.
> 
> Yea I'm not against having a list of "installable" PURLs and then a sublist of aliases/additional resources which can be used to, for example, discover source code for `deb` packages and vulnerabilities.

Aight, then I'll move towards this because it make more logical sense to me for the requirements installer, and then we can create a framework for increasing our query coverage with mappings later on. :+1:

simonklb:
> > > And if that is going to be a requirement then many purls are mismatched against OSV. Would there be any other sources we want to query purls against?
> > 
> > 
> > All the PURLs ideally. But matching CVEs against PURLs isn't a straight forward job, talking from experience from previous engagements... 😄
> 
> Note, I'm asking for _sources to query against_ so I can evaluate which purls to use before I create specific mappings that are going to replaced, not which purls to query as I assume we want that for everything. And as you have previous experience, I thought you might have some suggestions. 😉

Sorry I misread it. I am not aware of any other sources that can query PURLs directly so we'll likely have to parse and translate.

> Aight, then I'll move towards this because it make more logical sense to me for the requirements installer, and then we can create a framework for increasing our query coverage with mappings later on. 👍

Yea that sounds good. Let's focus on getting installations properly implemented and figure out security scanning later. Having extras/metadata part of the requirements structure doesn't hurt though.

simonklb (commented):
What is the reason for the DIY PURL parser over using one of the official ones? Other than that it's a fun exercise of course! :smile:

aarnq:
> What is the reason for the DIY PURL parser over using one of the official ones? Other than that it's a fun exercise of course! 😄

Zero dependencies.
I want to be able to use this in a container, so I don't want to have to add any runtime or external tool on top of it, and there are no official script version.

simonklb:
> > What is the reason for the DIY PURL parser over using one of the official ones? Other than that it's a fun exercise of course! 😄
> 
> Zero dependencies. I want to be able to use this in a container, so I don't want to have to add any runtime or external tool on top of it, and there are no official script version.

If you truly want to achieve zero dependencies it would make sense to me to build this in a compiled language. Otherwise you will have to bundle this with bash. The [go library](https://github.com/package-url/packageurl-go) is really straight forward if you want to give it a go (hehe)!

aarnq:
> > > What is the reason for the DIY PURL parser over using one of the official ones? Other than that it's a fun exercise of course! 😄
> > 
> > 
> > Zero dependencies. I want to be able to use this in a container, so I don't want to have to add any runtime or external tool on top of it, and there are no official script version.
> 
> If you truly want to achieve zero dependencies it would make sense to me to build this in a compiled language. Otherwise you will have to bundle this with bash. The [go library](https://github.com/package-url/packageurl-go) is really straight forward if you want to give it a go (hehe)!

Definitely in the future, but given how we have it today this is the simplest solution, and it doesn't require anything but a file to be included.

aarnq:
I have been working on this on and off for a while and now it is only missing a handler for helm-plugins, then it can install everything the current requirements installer can.
_And_ it can find the latest versions and checksums of them automagically.

simonklb:
This looks great! The only thing that is unfortunate is that it will be all done in the same layer in the Dockerfile. Would it be possible to perform them as separate layers so that you don't have to re-download everything if you just change one dependency?

aarnq:
> This looks great! The only thing that is unfortunate is that it will be all done in the same layer in the Dockerfile. Would it be possible to perform them as separate layers so that you don't have to re-download everything if you just change one dependency?

Not in a Docker/Containerfile directly, if we were to use something like Buildah and script the build process it would be easy, dunno if Docker has something similar :thinking: 

Perhaps if we had it generate/update a Docker/Containerfile with the steps to install it instead :thinking: 

My thinking though is that it will be a separate PR to make this ready for containers, and focus on the requirements first.

simonklb:
> > This looks great! The only thing that is unfortunate is that it will be all done in the same layer in the Dockerfile. Would it be possible to perform them as separate layers so that you don't have to re-download everything if you just change one dependency?
> 
> Not in a Docker/Containerfile directly, if we were to use something like Buildah and script the build process it would be easy, dunno if Docker has something similar 🤔
> 
> Perhaps if we had it generate/update a Docker/Containerfile with the steps to install it instead 🤔
> 
> My thinking though is that it will be a separate PR to make this ready for containers, and focus on the requirements first.

Yep, that makes sense!

aarnq:
helm-plugin support!

PR #2409, opened 2025-01-22
Add Crossplane module PoC

simonklb:
:warning: THIS SHOULD NOT BE MERGED :warning: 

This is a proof of concept of making a simple part of Apps managed by [Crossplane](https://www.crossplane.io/). It migrates the management of the metrics-server Helm chart to a [Crossplane Helm Release resource](https://github.com/crossplane/crossplane/blob/main/design/one-pager-helm-provider.md).

There are no diffs between the Helm release that we currently install and the one installed by the [Crossplane Helm Provider](https://github.com/crossplane-contrib/provider-helm) which means that an existing Helm release should be adopted without any changes to the metrics-server resources.

#### Information to reviewers

New charts:

* upstream/crossplane - This installs [Crossplane](https://docs.crossplane.io/v1.18/software/install/) itself.
* charts/crossplane-packages - This installs the [Providers](https://docs.crossplane.io/v1.18/concepts/providers/), [Functions](https://docs.crossplane.io/v1.18/guides/function-patch-and-transform/) and [Configurations](https://docs.crossplane.io/v1.18/concepts/packages/).
* charts/metrics-server - This installs the MetricsServer [Composite Resource](https://docs.crossplane.io/v1.18/concepts/composite-resources/).

To test it out for yourself, here's what I did.

On main:

```shell
helmfile -e service_cluster apply --selector app=metrics-server --include-transitive-needs
```

On this branch:

```shell
helmfile -e service_cluster apply --selector app=crossplane --include-transitive-needs

# This is a workaround. In a real world scenario we would have to implement this wait properly.
kubectl wait --for condition=Healthy=true providers.pkg.crossplane.io/provider-helm --timeout 120s
kubectl wait --for condition=Healthy=true configurations.pkg.crossplane.io/metrics-server --timeout 120s

helmfile -e service_cluster apply --selector app=metrics-server --include-transitive-needs
```


simonklb:
Closing after demo but keeping branch on remote so that you can continue looking at it and to allow for future proof of concept PRs. :)

Issue #2014, opened 2024-02-26
[2] Add an alert for missing Node Group labels


lunkan93:
### Proposed feature

The [reworked capacity management](https://github.com/elastisys/compliantkubernetes-apps/commit/5f1e11f3fb012cf33f8ad7bb73c75b78a70fea82) relies on the `elastisys.io/node-group=<node-group>` label on the nodes to function. We should add some optional lower priority alert that can alert the Platform Administrator if a node is missing this label.

### Proposed alternatives

The alternative is silently not getting capacity management metrics/alerts for any node missing this label, which can be difficult to notice.

### Definition of done

- [ ] We have an optional alert that triggers if the `elastisys.io/node-group=<node-group>` is missing from a node

PR #2411, opened 2025-01-22
apps: change query expression for LessKubeletsThanNodes alerts

HaoruiPeng:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

This change should mitigate the chances of false firing the **LessKubeletsThanNodes** alerts during autoscaling process.
The old query checks if there's any kubelet job down.
The new query compare the number of running kubelet jobs and the total number of nodes in the cluster, and triggers the alerts when the former is less than the latter.
The **LessKubeletsThanNodes** alerts is split into two based on the label `node-restriction.kubernetes.io/autoscaled-node-type`, which indicates if a node belongs to an autoscaling group or not.
1. **LessKubeletsThanNonautoscalableNodes**: Checks nodes without the mentioned label, P2 alerts has 5m pending time, P1 alert has 15m pending time.
2.  **LessKubeletsThanAutoscalableNodes**: Checks nodes with the mentioned label, P2 alerts has 15m pending time, P1 alert has 30m pending time.

NOTE: Currently the alerts does not check the values of the label, only checks if the label exits or not.

- Fixes https://github.com/elastisys/ck8s-cluster-api/issues/276

#### Information to reviewers

See the comment section of the issue for more details. 
#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
+++ b/helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
@@ -176,28 +176,85 @@ spec:
 {{- if .Values.defaultRules.additionalRuleLabels }}
 {{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
 {{- end }}
-    - alert: LessKubeletsThanNodes
+{{- if .Values.autoscaledNodeGroupAlerts.enabled }}
+    - alert: KubletDownForAutoscaledNodeFor15m
       annotations:
-        description: There are less kubelets than nodes for the last 5m.
-        runbook_url: {{ .Values.defaultRules.runbookUrl }}kubernetes/LessKubeletsThanNodes
-        summary: There are less kubelets than nodes for the last 5m.
-      expr: up{job="kubelet", metrics_path="/metrics"} == 0
+        description: The kubelet job of an existing autoscaled node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 15m.
+        summary: The kubelet job of an existing autoscaled cluster {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.node }}`}} is down for the last 15m.
+      expr: |-
+          sum by (node, cluster) (kube_node_info and on (node) up{job="kubelet",metrics_path="/metrics"} == 0)
+          and on (node)
+          kube_node_labels{ {{ .Values.autoscaledNodeGroupAlerts.groupLabel }}{{ .Values.autoscaledNodeGroupAlerts.groupLabelValue.regex }} }
+      for: 15m
+      labels:
+        severity: warning
+{{- if .Values.defaultRules.additionalRuleLabels }}
+{{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
+{{- end }}
+    - alert: KubletDownForAutoscaledNodeFor30m
+      annotations:
+        description: The kubelet job of an existing autoscaled node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 30m.
+        summary: The kubelet job of an existing autoscaled node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 30m.
+      expr: |-
+          sum by (node, cluster) (kube_node_info and on (node) up{job="kubelet",metrics_path="/metrics"} == 0)
+          and on (node)
+          kube_node_labels{ {{ .Values.autoscaledNodeGroupAlerts.groupLabel }}{{ .Values.autoscaledNodeGroupAlerts.groupLabelValue.regex }} }
+      for: 30m
+      labels:
+        severity: critical
+{{- if .Values.defaultRules.additionalRuleLabels }}
+{{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
+{{- end }}
+    - alert: KubletDownForNonAutoscaledNodeFor5m
+      annotations:
+        description: The kubelet job of an existing non-autoscaled node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 5m.
+        summary: The kubelet job of an existing non-autoscaled node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 5m.
+      expr: |-
+          sum by (node, cluster) (kube_node_info and on (node) up{job="kubelet",metrics_path="/metrics"} == 0)
+          unless on (node)
+          kube_node_labels{ {{ .Values.autoscaledNodeGroupAlerts.groupLabel }}{{ .Values.autoscaledNodeGroupAlerts.groupLabelValue.regex }} }
       for: 5m
       labels:
         severity: warning
 {{- if .Values.defaultRules.additionalRuleLabels }}
 {{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
 {{- end }}
-    - alert: LessKubeletsThanNodes
+    - alert: KubletDownForNonAutoscaledNodeFor15m
       annotations:
-        description: There are less kubelets than nodes for the last 15m.
-        runbook_url: {{ .Values.defaultRules.runbookUrl }}kubernetes/LessKubeletsThanNodes
-        summary: There are less kubelets than nodes for the last 15m.
-      expr: up{job="kubelet", metrics_path="/metrics"} == 0
+        description: The kubelet job of an existing non-autoscaled node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 15m.
+        summary: The kubelet job of an existing non-autoscaled node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 15m.
+      expr: |-
+          sum by (node, cluster) (kube_node_info and on (node) up{job="kubelet",metrics_path="/metrics"} == 0)
+          unless on (node)
+          kube_node_labels{ {{ .Values.autoscaledNodeGroupAlerts.groupLabel }}{{ .Values.autoscaledNodeGroupAlerts.groupLabelValue.regex }} }
       for: 15m
       labels:
         severity: critical
 {{- if .Values.defaultRules.additionalRuleLabels }}
 {{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
+{{- end }}
+{{- else }}
+    - alert: KubeletDownFor5m
+      annotations:
+        description: The kubelet job of an existing node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 5m.
+        summary: The kubelet job of an existing node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 5m.
+      expr: sum by (node, cluster) (kube_node_info and on (node) up{job="kubelet",metrics_path="/metrics"} == 0)
+      for: 5m
+      labels:
+        severity: warning
+{{- if .Values.defaultRules.additionalRuleLabels }}
+{{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
+{{- end }}
+    - alert: KubeletDownFor15m
+      annotations:
+        description: The kubelet job of an existing node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 15m.
+        summary: The kubelet job of an existing node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 15m.
+      expr: sum by (node, cluster) (kube_node_info and on (node) up{job="kubelet",metrics_path="/metrics"} == 0)
+      for: 15m
+      labels:
+        severity: critical
+{{- if .Values.defaultRules.additionalRuleLabels }}
+{{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
+{{- end }}
 {{- end }}
 {{- end }}
--- a/config/common-config.yaml
+++ b/config/common-config.yaml
@@ -399,6 +399,12 @@ prometheus:
   nodeSelector: {}
   webhookAlerts:
     enabled: true
+  # Split alerts for autoscaled node from normal node
+  # TODO: .autoscaledNodeGroupAlerts.groupLabel must have value if autoscaledNodeGroupAlerts is enabled, can be configured with a different label in the future
+  autoscaledNodeGroupAlerts:
+    enabled: true
+    groupLabel: "node-restriction.kubernetes.io/autoscaled-node-type"
+    groupLabelValues: []
   ## Predictive alert if the resource usage will hit the set percentage in 3 days
   capacityManagementAlerts:
     enabled: true
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -4440,6 +4440,23 @@ properties:
             items:
               title: S3 Bucket Name
               type: string
+      autoscaledNodeGroupAlerts:
+        title: Autoscaled NodeGroup Alerts
+        description: Configure whether to split KubeletDownForXm alerts into autoscaled and non-autoscaled nodes groups.
+        type: object
+        additionalProperties: false
+        properties:
+          groupLabel:
+            title: Autoscaled node group label
+            description: The label to identity whether a node belongs to an autoscaled node group.
+            type: string
+            default: "node-restriction.kubernetes.io/autoscaled-node-type"
+          groupLabelValues:
+            title: Autoscaled node group label values
+            description: The label values to a autoscaled node group if their are multiple autoscaled node groups.
+            type: array
+            items:
+              type: string
     additionalProperties: false
     properties:
       replicas:
@@ -4615,6 +4632,19 @@ properties:
             title: Webhook Alerts Enabled
             type: boolean
             default: true
+      autoscaledNodeGroupAlerts:
+        title: Autoscaled NodeGroup Alerts
+        description: Configure whether to split KubeletDownForXm alerts into autoscaled and non-autoscaled nodes groups.
+        type: object
+        properties:
+          enabled:
+            title: Enabled aleter splitting for autoscaled nodes.
+            type: boolean
+            default: true
+          groupLabel:
+            $ref: "#/properties/prometheus/$defs/autoscaledNodeGroupAlerts/properties/groupLabel"
+          groupLabelValues:
+            $ref: "#/properties/prometheus/$defs/autoscaledNodeGroupAlerts/properties/groupLabelValues"
       resources:
         $ref: '#/$defs/kubernetesResourceRequirements'
       tolerations:
--- a/helmfile.d/charts/prometheus-alerts/values.yaml
+++ b/helmfile.d/charts/prometheus-alerts/values.yaml
@@ -19,6 +19,12 @@ s3BucketAlerts:
     count: 1638400
   exclude: []
 
+autoscaledNodeGroupAlerts:
+  enabled: true
+  groupLabel: set-me
+  groupLabelValue:
+    regex: set-me
+
 diskAlerts:
   storage:
     #  - hours: 24 # How many hours to predict over
--- a/helmfile.d/values/kube-prometheus-stack-sc.yaml.gotmpl
+++ b/helmfile.d/values/kube-prometheus-stack-sc.yaml.gotmpl
@@ -47,8 +47,14 @@ kube-state-metrics:
   extraArgs:
     - "--custom-resource-state-config-file=/etc/config/clusterapi-metrics.yaml"
   {{- end }}
+  {{- if .Values.prometheus.autoscaledNodeGroupAlerts.enabled }}
+  metricLabelsAllowlist:
+  - nodes=[topology.kubernetes.io/zone,elastisys.io/node-group,{{ .Values.prometheus.autoscaledNodeGroupAlerts.groupLabel }}]
+  {{- else }}
   metricLabelsAllowlist:
   - nodes=[topology.kubernetes.io/zone,elastisys.io/node-group]
+  {{- end }}
+
 kubeApiServer:
   serviceMonitor:
     metricRelabelings:
--- a/helmfile.d/values/kube-prometheus-stack-wc.yaml.gotmpl
+++ b/helmfile.d/values/kube-prometheus-stack-wc.yaml.gotmpl
@@ -25,8 +25,13 @@ kube-state-metrics:
     enabled: true
   resources: {{- toYaml .Values.kubeStateMetrics.resources | nindent 4 }}
 
+  {{- if .Values.prometheus.autoscaledNodeGroupAlerts.enabled }}
+  metricLabelsAllowlist:
+  - nodes=[topology.kubernetes.io/zone,elastisys.io/node-group,node.kubernetes.io/instance-type,{{ .Values.prometheus.autoscaledNodeGroupAlerts.groupLabel }}]
+  {{- else }}
   metricLabelsAllowlist:
   - nodes=[topology.kubernetes.io/zone,elastisys.io/node-group,node.kubernetes.io/instance-type]
+  {{- end }}
 
 prometheus-node-exporter:
   resources: {{- toYaml .Values.prometheusNodeExporter.resources | nindent 4 }}
--- a/helmfile.d/values/prometheus-alerts-sc.yaml.gotmpl
+++ b/helmfile.d/values/prometheus-alerts-sc.yaml.gotmpl
@@ -4,6 +4,14 @@
 
 # TODO: We should update the rules for things with set namespace and job name to be able to monitor instances in the workload cluster.
 
+# Get and transform autoscaledNodeGroupAlerts label and values
+{{- $groupLabel := .Values.prometheus.autoscaledNodeGroupAlerts.groupLabel -}}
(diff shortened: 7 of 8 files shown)


helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
```diff
@@ -181,7 +181,7 @@ spec:
         description: There are less kubelets than nodes for the last 5m.
         runbook_url: {{ .Values.defaultRules.runbookUrl }}kubernetes/LessKubeletsThanNodes
         summary: There are less kubelets than nodes for the last 5m.
-      expr: up{job="kubelet", metrics_path="/metrics"} == 0
+      expr: count(up{job="kubelet", metrics_path="/metrics"} == 1) < count(kube_node_status_condition{condition="Ready", status="true"} == 1)
```
Xartos, line 184: I'm not sure if this captures what we want, testing it on an affected time period definitely works. But I'm unsure if this alert would ever trigger?
If the kubelet is down I'd argue that the node is always "NotReady" by definition.
However with that said, I can't seem to figure out any way to capture this in a good way.

If the node would've been labeled by capi during scaledown we could "filter" them out from the previous expression. Buut they don't get any...
We could've tried to split the alert up into "autoscaled nodes" and "non-autoscaled nodes" using our naming convention of autoscaled nodes and have the timeout longer for "autoscaled nodes". Buut that seems like an ugly solution. 

So I think that I'd like us to rather try to get to the source of the slow scale down on GPU nodes, is there something that CAPI/CAPZ does with the nodes from the time that the node gets its kubelet shutdown/removed until the node object is removed from the cluster that we can speed up. Buuut we can't get any more quota to test this 🤦 And I'm not sure if the scope of this issue should become greater than it already is. 

So I think I'm honestly at a loss. Any one else that has any ideas?

HaoruiPeng: > I'm not sure if this captures what we want, testing it on an affected time period definitely works. But I'm unsure if this alert would ever trigger?
> If the kubelet is down I'd argue that the node is always "NotReady" by definition.
> However with that said, I can't seem to figure out any way to capture this in a good way.

I was also thinking about this. Another idea I had is to compare the number of running kubelet to number of total nodes, regardless if they are ready or not.  In this way, the alert will be fired in the GPU nodes autoscaling case (according to the metrics data we have), but should have less probability to fire in other cases compared to the old `up{job="kubelet", metrics_path="/metrics"} == 0` query.

But I'm also not sure about the other scenarios in ops that we want to capture with this alert (Since I didn't encounter this in my own dev clusters), so yes the changes I made may filter out some events we are supposed to capture.  Also had similar ideas that to filter out the autoscaled nodes, but as you said, it seems not a neat solution :/

Xartos: > but should have less probability to fire in other cases compared to the old `up{job="kubelet", metrics_path="/metrics"} == 0` query

Right, so basically this query would be

```
# Not a real query, just illustrative
count(up{job="kubelet", metrics_path="/metrics"} == 1) < count(nodes)
```

And that would capture the case where the node exists but the kubelet scrape target doesn't _AND_ when it exists and it's down instead of only capturing the case where it exists and is down, right?
That would certainly match the name of the alert better at least, and it feels like a good approach. But, as you said, it wouldn't fix the issue of autoscaling

anders-elastisys: Would splitting this alert into separate alerts per node group be a possible solution? E.g. `LessKubeletsThanNodesControlPlane` and `LessKubeletsThanNodesGPU` or something. Then you could have separate duration configured for a GPU node-group if we know they take longer time to scale down. But that requires that we have specific labels for the GPU nodes.

HaoruiPeng: Wanna write here a summary on possible approaches we have brought up and I have thought (will add if we have came up more), please leave some thoughts.
1. Compare with `count(up{kubelet})` with `count(nodes)`, 
    - Pro: Matche better the alert name, and mitigate the problem of firing alerts due to the weird behavior of `up{job="kubelet", metrics_path="/metrics"} == 0` metrics.
    - Con: Does not solve the autoscaling issue with gpu nodes.
3. Split alerts for 'non-autoscaled nodes' and 'autoscale node', and use longer timeout for the latter.  The node label `node-restriction.kubernetes.io/autoscaled-node-type` can be used for splitting.
    - Pro: Kinda solves the con of solution 1.
    - Con: Non-GPU but autoscaled nodes willl be affected.
5. Split alerts based on `elastisys.io/node-group` label and use longer timeout for GPU nodes.
    - Pro: Seems a better way of splitting, as we basically filter out GPU nodes from this alert.
    - Con: I don't think we have enforced adding this label to our clusters yet . The lables were added by default if clusters were deployed by devbox (https://github.com/elastisys/ck8s-devbox/issues/374),  but we didn't have gpu nodes labled ``elastisys.io/node-group=gpu`` by that fix. This solution could work as expected after we make sure all our clusters have been enforced to add the labels by these two issues https://github.com/elastisys/compliantkubernetes-apps/issues/2014, https://github.com/elastisys/ck8s-issue-tracker/issues/429 (include GPU nodes)
1. Split alerts based on `elastisys.io/node-type` label and use longer timeout for GPU nodes.
   - Pro: Compared to solution 3, this label already exists on our gpu nodes. Only prior fix is to add it to our Prometheus `metricLabelsAllowlist`.
   - Con: This feels more like a workaround than a solution? Also a bit ugly to goto-monitoring people?

anders-elastisys: I think for option 2 you can configure it per key value pair, so it only needs to have increased duration for e.g. `node-restriction.kubernetes.io/autoscaled-node-type=gpu` or similar and let duration be the same as for non-autoscale nodes, but it is maybe not as clean as option 3. I am not entirely sure how if we have a default set up for GPU node groups in CAPI to ensure which label names are used.
Personally I like option 3 if we implement validation that ensures that nodes have the proper labels in this repo, but since that is not in place atm we can maybe wait with implementing this. A short-term solution could be to go with option 1 for now to get this merged and then implement 3 once https://github.com/elastisys/ck8s-issue-tracker/issues/429 is implemented, but I am interested to hear what others think of this @Xartos  @elastisys/ops-lead @davidumea

viktor-f: I think it would make sense to go for option 2 as the default configuration but make the label configurable. So that by default it just checks if there label `node-restriction.kubernetes.io/autoscaled-node-type` is there, but we can then change it to a specific label that catches just gpu nodes if we want (including both label key and value). That way we do not have to ensure that any specific label is set.

HaoruiPeng:
PR updated with [solution 2](https://github.com/elastisys/compliantkubernetes-apps/pull/2411#discussion_r1936813889) suggested in the comment section.

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
```diff
@@ -176,24 +176,44 @@ spec:
 {{- if .Values.defaultRules.additionalRuleLabels }}
 {{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
 {{- end }}
-    - alert: LessKubeletsThanNodes
+    - alert: LessKubeletsThanAutoscalableNodes
```
davidumea, line 179: What do you think of using "Autoscaled" instead of "Autoscalable"?

```suggestion
    - alert: LessKubeletsThanAutoscaledNodes
```
...

HaoruiPeng: Was expecting some suggestions on the alert names :) 
I can change that for both alerts.

davidumea:
I think you need to rebase or drop this commit https://github.com/elastisys/compliantkubernetes-apps/pull/2411/commits/1d72f2941b6b768f451eaacff1972f0fc967e359

HaoruiPeng:
> I think you need to rebase or drop this commit [1d72f29](https://github.com/elastisys/compliantkubernetes-apps/commit/1d72f2941b6b768f451eaacff1972f0fc967e359)

Ouch I cherry-picked it for another quick test and forgot to drop it. Thanks for reminding!

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
```diff
@@ -176,24 +176,44 @@ spec:
...
+      annotations:
+        description: There are less running kubelet jobs than existing nodes in a non-autoscaled node group for the last 5m.
+        summary: There are less running kubelet jobs than existing nodes in a non-autoscaled node group for the last 5m.
+      expr: count((up{job="kubelet", metrics_path="/metrics"} == 1) unless on(node) kube_node_labels{label_node_restriction_kubernetes_io_autoscaled_node_type!=""}) < count(kube_node_info unless on(node) kube_node_labels{label_node_restriction_kubernetes_io_autoscaled_node_type!=""})
       for: 5m
       labels:
         severity: warning
 {{- if .Values.defaultRules.additionalRuleLabels }}
 {{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
 {{- end }}
```
viktor-f, line 211: I might be missing something. But it seems like both of the alerts have duplicate entries.
```suggestion
```

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
```diff
@@ -176,24 +176,44 @@ spec:
...
-    - alert: LessKubeletsThanNodes
+    - alert: LessKubeletsThanAutoscaledNodes
       annotations:
-        description: There are less kubelets than nodes for the last 5m.
-        runbook_url: {{ .Values.defaultRules.runbookUrl }}kubernetes/LessKubeletsThanNodes
-        summary: There are less kubelets than nodes for the last 5m.
-      expr: up{job="kubelet", metrics_path="/metrics"} == 0
+        description: There are less running kubelet jobs than existing nodes in an autoscaling enabled node group for the last 15m.
+        summary: There are less running kubelet jobs than existing nodes in an autoscaling enabled node group for the last 15m.
+      expr: count((up{job="kubelet", metrics_path="/metrics"} == 1) and on(node) kube_node_labels{label_node_restriction_kubernetes_io_autoscaled_node_type!=""}) < count(kube_node_info and on(node) kube_node_labels{label_node_restriction_kubernetes_io_autoscaled_node_type!=""})
```
viktor-f, line 183: As I suggested in the other comment thread I would like it if was possible to change the label (including value) this looks at via config.
Perhaps even making it possible to look at several labels, but that be more complicated than it is worth.
What do you think of that?

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
HaoruiPeng: Yes, they are duplicated but with different pending time, 5m triggers P2 alert, 15m triggers P1 alert.

viktor-f: Ah, I totally missed that.  Thanks!

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
HaoruiPeng: Do you mean something like `kube_node_labels{label_node_restriction_kubernetes_io_autoscaled_node_type=$value}`, and we read the `$value` from the value file? But we need to pre-config it in the value file anyway right?

I was trying to get labels out from the query results, so it tells which node triggers the alerts. But I didn't manage to do that. To get the instance label information we need to compare the kulebet and kube_node_info for individual node instead of summing them up. But in that case the query will drop the nodes returning nil values, which is exactly what we want to capture with the alert. Couldn't think of any idea for it. What do you think? @anders-elastisys

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
```diff
@@ -176,24 +176,44 @@ spec:
...
+        severity: warning
+{{- if .Values.defaultRules.additionalRuleLabels }}
+{{ toYaml .Values.defaultRules.additionalRuleLabels | indent 8 }}
+{{- end }}
+    - alert: LessKubeletsThanAutoscaledNodes
+      annotations:
+        description: There are less running kubelet jobs than existing nodes in an autoscaling enabled group for the last 15m.
+        summary: There are less running kubelet jobs than existing nodes in an autoscaling enabled group for the last 15m.
+      expr: count((up{job="kubelet", metrics_path="/metrics"} == 1) and on(node) kube_node_labels{label_node_restriction_kubernetes_io_autoscaled_node_type!=""}) < count(kube_node_info and on(node) kube_node_labels{label_node_restriction_kubernetes_io_autoscaled_node_type!=""})
+      for: 30m
```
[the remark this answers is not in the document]
HaoruiPeng: Yes that's a typo from bad copy-paste :)

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
viktor-f: Yes, exactly. As you say we do need a default value that we put in [config/common-config.yaml](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/common-config.yaml).

Hmm, yeah I think that it might not be possible to get the missing nodes with the way we are currently calculating this.
The current approach of `calculate_up_nodes - calculate_all_registered_nodes` would not give us enough context. If we want to know what nodes are missing, then we would need to make some individual comparison for individual nodes between the two metrics. It's probably possible, but would be a bit different. It would then also create one alert per missing node instead of one alert regardless of how many nodes are missing.

HaoruiPeng: > Yes, exactly. As you say we do need a default value that we put in [config/common-config.yaml](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/common-config.yaml).

Ah ok, I will update it to make it configurable in the common-config.yaml

> The current approach of calculate_up_nodes - calculate_all_registered_nodes would not give us enough context. If we want to know what nodes are missing, then we would need to make some individual comparison for individual nodes between the two metrics. It's probably possible, but would be a bit different. It would then also create one alert per missing node instead of one alert regardless of how many nodes are missing.

Yea I was trying to get label information for missing node but didn't manage to do that, the query logic would be different if we want this, maybe we'd better change the alerts names 😂, since the old name suggests the compare between the sums. 
I will look into that more and get back with an alternative for that.

HaoruiPeng: Did some more digs into this alert, I think we need to first clarify what we exactly want this alert to capture. 

1. The old LessKubeletThanNodes alert query only calculates `count(up{job="kubelet"}==0)`, which triggers if any kubelet job is not running properly.  However, when a node was already taken down from the cluster after autoscaling, the `up{job="kubelet"}` query for some reason can still return 0 value on that non-existing node for a while and eventually notifies us the kubelet is not running. We don't want our alert to capture it because the node is already taken down, and we don't care if the metric for that node is still there, whether returning 0 or 1.

2. If we change the query to `count(up{job="kubelet"}==1) < count{kube_node_info}` avoid the false trigger, and it capture two things: 
    1. kubelet job of an existing node is not running, aka. `up{job="kubelet"}` returns 0 value on that node.
    2. kubelet is absent on an existing node, this will also be captured by **KubeletDown** alert.
The cons of using this query or the old query is that we don't get any label info on which node is wrong, since we only calculated the `count` metric.

3. If we compare it node by node, the query will be like `sum by (node)(kube_node_info) - sum by (node)(up{job="kubelet"})`, with this we can get information on which existing nodes' kubelet stopped running, but we won't capture anything when `kube_node_info` or `up{job="kubelet"}` is absent for a node, since the query does not calculate on  nil value. I think this query is equivalent to `kube_node_info and on(node) up{job="kubelet"}`, which only checks `up{job="kubelet"}` on existing nodes, return 1 if kubelet is running, 0 if kubelet stops, returns no data (nil) if this metric is absent. 

I'm inclined towards `kube_node_info and on(node) up{job="kubelet"}`, which ignores the absent node or absent kubelet job, but also won't trigger at all during a normal autoscaling case (not even in pending status). 

Any thoughts? @viktor-f @anders-elastisys

HaoruiPeng:
PR updated, changed to query to `sum by (node, cluster) (kube_node_info and on (node) up{job="kubelet",metrics_path="/metrics"} == 0)`. Now the alert will be triggered by individual node with configurable label and label values, the message will be like.
Screenshot from 2025-02-05 14-03-54

I also added back LessKubeletThanNode alert with the new metric, this alert will only be enabled if alert splitting for autoscaled node is disabled in the config.

Still need some thought on the alert names and the query, since the logic does not match any more if we choose to go with this way.

anders-elastisys (approved):
I like this change, the alerts clearly shows which nodes are affected. I think the name of the alert should maybe be changed tho, but I am not sure what to name it without confusing it with the `KubeletDown` alert :thinking: 
Maybe the `KubeletDown` alert can be renamed to `KubeletAbsent`, I see upstream `absent()` alerts that have a `absent` suffix or similar, but the upstream for this alert is still `KubeletDown` :thinking:
But if we changed it. then these alerts can instead be e.g. `KubletDownForAutoscaledNodeFor15m`, also including the deadline time to also separate between the warning and critical alerts. Naming things is difficult :face_with_spiral_eyes:

davidumea:
> these alerts can instead be e.g. `KubletDownForAutoscaledNodeFor15m`, also including the deadline time to also separate between the warning and critical alerts.

I like this suggestion, I personally don't see the point in mimicking the "upstream" name if we can have a better one :slightly_smiling_face:

HaoruiPeng:
Names changed, I will start fixing json schema checks for the newly added `.prometheus.autoscaledNodeGroupAlerts` config if everyone agrees with this alert change.

HaoruiPeng:
This PR has passed all check for merge, but I don't think the latest change has received enough approval to be merged. Please approve the PR if you agree with all the changes, otherwise leave your comment/review. @davidumea @Xartos @viktor-f @raviranjanelastisys

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
```diff
@@ -176,28 +176,85 @@ spec:
...
 {{- end }}
-    - alert: LessKubeletsThanNodes
+{{- if .Values.autoscaledNodeGroupAlerts.enabled }}
+    - alert: KubletDownForAutoscaledNodeFor15m
       annotations:
-        description: There are less kubelets than nodes for the last 5m.
-        runbook_url: {{ .Values.defaultRules.runbookUrl }}kubernetes/LessKubeletsThanNodes
-        summary: There are less kubelets than nodes for the last 5m.
-      expr: up{job="kubelet", metrics_path="/metrics"} == 0
+        description: The kubelet job of an existing auntoscaled node {{`{{`}} $labels.node {{`}}`}} in cluster {{`{{`}} $labels.cluster {{`}}`}} is down for the last 15m.
```
davidumea, line 182: Can't we do this? Also noticed a typo
```suggestion
        description: The kubelet job of an existing autoscaled node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 15m.
```

HaoruiPeng: Typo fixed. I checked and both expression work. Changed to a simpler one.

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
```diff
@@ -176,28 +176,85 @@ spec:
...
-        description: There are less kubelets than nodes for the last 5m.
-        runbook_url: {{ .Values.defaultRules.runbookUrl }}kubernetes/LessKubeletsThanNodes
-        summary: There are less kubelets than nodes for the last 5m.
-      expr: up{job="kubelet", metrics_path="/metrics"} == 0
+        description: The kubelet job of an existing autoscaled node {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.cluster }}`}} is down for the last 15m.
+        summary: The kubelet job of an existing autoscaled cluster {{`{{ $labels.node }}`}} in cluster {{`{{ $labels.node }}`}} is down for the last 15m.
+      expr: |-
+          sum by (node, cluster) (kube_node_info and on (node) up{job="kubelet",metrics_path="/metrics"} == 0)
+          and on (node)
+          kube_node_labels{ {{ .Values.autoscaledNodeGroupAlerts.groupLabel }}{{ .Values.autoscaledNodeGroupAlerts.groupLabelValue.regex }} }
```
Xartos, line 187: Suggestion: I feel like it's easier to understand how this works if you move the `=~` here

```suggestion
          kube_node_labels{ {{ .Values.autoscaledNodeGroupAlerts.groupLabel }}=~{{ .Values.autoscaledNodeGroupAlerts.groupLabelValue.regex }} }
```

Since we know it's a regex given the name of the variable.
But I could also accept leaving it like this if you feel differently

Xartos (commented):
Very nice work! I only have a minor nit

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
HaoruiPeng: The problem is that we have do use `!=` when the groupLableValue is an empty array, that would add another if statement in the alert template, makes it a bit messy. So I choose to put the if statement in the value file.

Xartos: Hmm, right. Yea that make sense, yea then keep it :)
Side note, isn't it `!~` when you're dealing with "not equals regex"?

> It is also possible to negatively match a label value, or to match label values against regular expressions. The following label matching operators exist:
> * `=`: Select labels that are exactly equal to the provided string.
> * `!=`: Select labels that are not equal to the provided string.
> * `=~`: Select labels that regex-match the provided string.
> * `!~`: Select labels that do not regex-match the provided string.

HaoruiPeng: Yes, not equal is only used when the label value is empty, in this case we filter out all nodes with autoscaledNode label regardless the label value. It's always an empty string at right of this equation, not a regex expression.

viktor-f (approved):
Nice work, I like that this now triggers per node and that the label is configurable :slightly_smiling_face:

helmfile.d/charts/prometheus-alerts/templates/alerts/kubernetes-system-kubelet.yaml
HaoruiPeng: @Xartos  Thought through about it, my previous explanation is correct. `!=` is only used when label value array is empty, so there won't be any regex expression on the right side of the expression. So there's no point using `!~`.
Here in the value file is how I decide when to use `=~` and when to use `!=`. 
```
autoscaledNodeGroupAlerts:
  enabled: {{ .Values.prometheus.autoscaledNodeGroupAlerts.enabled }}
  groupLabel: {{ $groupLabel }}
  {{- if $groupLabelValueArrayRegex }}
  groupLabelValue:
    regex: '=~"{{ $groupLabelValueArrayRegex }}"'
  {{- else }}
  groupLabelValue:
    regex: '!="{{ $groupLabelValueArrayRegex }}"'
  {{- end }}

```
The reason why `!=` is used here for filtering out nodes is that when the value array is empty, `kube_node_labels{groupLable=~""}` couldn't split nodes with and without `groupLable` label,  it returns the metrics for all nodes, so I had to use `kube_node_labels{groupLable!=""}` to separate them. When the value array is not empty, `kube_node_labels{groupLable=~"value1|value2"}` can split the nodes based on this label without any issue.

Xartos: Riiight! Then maybe you could change it into

```
  {{- if $groupLabelValueArrayRegex }}
  groupLabelValue:
    regex: '=~"{{ $groupLabelValueArrayRegex }}"'
  {{- else }}
  groupLabelValue:
    regex: '!=""'
  {{- end }}
```

Just to make this a bit clearer? What do you think?

HaoruiPeng: Yes that is more clear, will change that

Xartos (approved):
Looks good to me now :) Great work

davidumea (approved):

davidumea (commented):
Squash reminder :slightly_smiling_face:

PR #2440, opened 2025-02-18
Allow subnet in update-ips

simonklb:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Running `ck8s update-ips` on existing OpenStack CAPI clusters will now allow the entire subnet instead of individual node IPs.

For OpenStack CAPI clusters allow the entire subnet instead of individual node IPs.

This should help with autoscaling where nodes are added on demand. Without doing this the administrator would have to run `update-ips` to make sure the new nodes are allowed.

It also has the benefit of allowing non-node addresses in the subnet such as loadbalancers.

#### Information to reviewers

It looks like it's already common practice to set the entire subnet rather than allowing individual IPs, it just has been done manually. However, if anyone has a good reason why we should **not** allow the entire subnet please speak up!

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/bin/update-ips.bash
+++ b/bin/update-ips.bash
@@ -117,13 +117,13 @@ rclone_enabled() {
   return 1
 }
 
-# Fetch the InternalIP, Calico tunnel IP and Wireguard IP of Kubernetes
+# Fetch the Calico tunnel IP and Wireguard IP of Kubernetes
 # nodes using a label selector.
 #
 # If the label selector isn't specified, all nodes will be returned.
 #
-# Usage: get_kubectl_ips <cluster> <label>
-get_kubectl_ips() {
+# Usage: get_tunnel_ips <cluster> <label>
+get_tunnel_ips() {
   local cluster="${1}"
   local label="${2}"
 
@@ -132,17 +132,40 @@ get_kubectl_ips() {
     label_argument="-l ${label}"
   fi
 
-  local -a ips_internal
   local -a ips_calico_ipip
   local -a ips_calico_vxlan
   local -a ips_wireguard
-  mapfile -t ips_internal < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].status.addresses[?(@.type=="InternalIP")].address}')
   mapfile -t ips_calico_vxlan < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv4VXLANTunnelAddr}')
   mapfile -t ips_calico_ipip < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv4IPIPTunnelAddr}')
   mapfile -t ips_wireguard < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv4WireguardInterfaceAddr}')
 
   local -a ips
-  read -r -a ips <<<"${ips_internal[*]} ${ips_calico_vxlan[*]} ${ips_calico_ipip[*]} ${ips_wireguard[*]}"
+  read -r -a ips <<<"${ips_calico_vxlan[*]} ${ips_calico_ipip[*]} ${ips_wireguard[*]}"
+
+  if [ ${#ips[@]} -eq 0 ]; then
+    log_error "No IPs for ${cluster} nodes with label ${label} was found"
+    exit 1
+  fi
+
+  echo "${ips[@]}"
+}
+
+# Fetch the InternalIP of Kubernetes nodes using a label selector.
+#
+# If the label selector isn't specified, all nodes will be returned.
+#
+# Usage: get_internal_ips <cluster> <label>
+get_internal_ips() {
+  local cluster="${1}"
+  local label="${2}"
+
+  local label_argument="--ignore-not-found"
+  if [[ "${label}" != "" ]]; then
+    label_argument="-l ${label}"
+  fi
+
+  local -a ips
+  mapfile -t ips < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].status.addresses[?(@.type=="InternalIP")].address}')
 
   if [ ${#ips[@]} -eq 0 ]; then
     log_error "No IPs for ${cluster} nodes with label ${label} was found"
@@ -238,9 +261,9 @@ get_swift_url() {
 
 # Sort IP addresses.
 #
-# For example, [1.0.0.10, 1.0.0.2] would be reordered to [1.0.0.2, 1.0.0.10].
-sort_ips() {
-  python3 -c "import ipaddress; import sys; ips = [ipaddress.ip_address(ip) for ip in sys.argv[1:]]; ips.sort(); [print(ip) for ip in ips]" "${@}"
+# For example, [1.0.0.10/32, 1.0.0.2/32] would be reordered to [1.0.0.2/32, 1.0.0.10/32].
+sort_cidrs() {
+  python3 -c "import ipaddress; import sys; cidrs = [ipaddress.ip_network(cidr) for cidr in sys.argv[1:]]; cidrs.sort(); [print(cidr) for cidr in cidrs]" "${@}"
 }
 
 # Check if an IP address is part of a CIDR address block.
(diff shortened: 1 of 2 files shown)


bin/update-ips.bash
```diff
@@ -304,26 +327,37 @@ parse_url_port() {
...
+# Usage: allow_ips <config_file> <config_option> <ip> [<ip> ...]
+allow_ips() {
+  local config_file="${1}"
+  local config_option="${2}"
+  shift 2
+
+  local -a cidrs
+  readarray -t cidrs <<<"$(process_ips_to_cidrs "${config_file}" "${config_option}" "${@}")"
+
+  allow_cidrs "${config_file}" "${config_option}" "${cidrs[@]}"
```
Xartos, line 358: Suggestion: Do we want to sort this before sending it? I can't see that the `process_ips_to_cidrs` does this, but I might be missing something.

```suggestion
  mapfile -t sortedCidrs < <(sort_cidrs "${cidrs[@]}")
  allow_cidrs "${config_file}" "${config_option}" "${sortedCidrs[@]}"
```

simonklb: It's sorted in `allow_cidrs`: https://github.com/elastisys/compliantkubernetes-apps/blob/e7e05d51347bc3cd407fc53542fbdc1cb5058bac/bin/update-ips.bash#L333-L339

simonklb: We also have this test that verifies that they are sorted: https://github.com/elastisys/compliantkubernetes-apps/blob/e7e05d51347bc3cd407fc53542fbdc1cb5058bac/tests/unit/general/bin-update-ips.bats#L115-L128

Xartos: Right, sorry. Yea I was looking at that function and noticed that it wasn't sorted here. But you're correct I didn't connect the dots there :sweat_smile:

Xartos (approved):

bin/update-ips.bash
```diff
@@ -402,13 +436,56 @@ allow_nodes() {
...
+    allow_nodes "${cluster}" "${config_option}" ""
+    return
+  fi
+
+  local config_file="${config["override_${cluster}"]}"
+
+  local environment_name
+  environment_name="$(yq_read "${cluster}" '.global.ck8sEnvironmentName' "")"
+
+  # TODO: We should not have to do this in the future...
```
Zash, line 466: Why and how to identify the when would be good to clarify here.

bin/update-ips.bash
```diff
@@ -402,13 +436,56 @@ allow_nodes() {
...
+  capi_cluster_name="${environment_name}-${cluster}"
+  if [ "${cluster}" = "sc" ]; then
+    # If no cluster named <environment>-sc is found, try <environment>-mc
+    if ! "${here}/ops.bash" kubectl sc -n capi-cluster get openstackcluster "${capi_cluster_name}" >/dev/null 2>&1; then
+      capi_cluster_name="${environment_name}-mc"
+    fi
+  fi
+
+  local subnet_cidr
+  subnet_cidr=$("${here}/ops.bash" kubectl sc -n capi-cluster get openstackcluster "${capi_cluster_name}" -o jsonpath='{.status.network.subnets[0].cidr}')
```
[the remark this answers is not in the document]
Zash: Worth pointing out in a comment, so someone adding Azure (or other provider) support finds it easily?

tests/unit/general/bin-update-ips-subnet.bats
```diff
@@ -0,0 +1,94 @@
...
+setup_file() {
+  load "../../bats.lib.bash"
+  load_common "env.bash"
+  load_common "gpg.bash"
+  load_common "yq.bash"
+
+  gpg.setup
+  env.setup
+
+  env.init openstack capi dev
```
Zash, line 14: It would be nice to test the fallback in a non-openstack and/or non-capi environment as well.

Zash (changes requested):
Overall looks good. I would like to see more varied test coverage.

bin/update-ips.bash
```diff
@@ -238,9 +261,9 @@ get_swift_url() {
 
 # Sort IP addresses.
 #
-# For example, [1.0.0.10, 1.0.0.2] would be reordered to [1.0.0.2, 1.0.0.10].
-sort_ips() {
-  python3 -c "import ipaddress; import sys; ips = [ipaddress.ip_address(ip) for ip in sys.argv[1:]]; ips.sort(); [print(ip) for ip in ips]" "${@}"
+# For example, [1.0.0.10/32, 1.0.0.2/32] would be reordered to [1.0.0.2/32, 1.0.0.10/32].
+sort_cidrs() {
+  python3 -c "import ipaddress; import sys; cidrs = [ipaddress.ip_network(cidr) for cidr in sys.argv[1:]]; cidrs.sort(); [print(cidr) for cidr in cidrs]" "${@}"
```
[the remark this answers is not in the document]
simonklb: If we could get rid of all Python call outs I would be fine with replacing it with this but since we would then have the IPv6 issue and we use Python in other parts of the script I'll leave it as it is for now.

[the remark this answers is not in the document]
simonklb: The `printf` thing was neat though!

tests/unit/general/bin-update-ips-subnet.bats
simonklb: That should be what all the existing tests in https://github.com/elastisys/compliantkubernetes-apps/blob/e7e05d51347bc3cd407fc53542fbdc1cb5058bac/tests/unit/general/bin-update-ips.bats does. No?

bin/update-ips.bash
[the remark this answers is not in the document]
simonklb: https://github.com/elastisys/compliantkubernetes-apps/pull/2440/commits/24efda54eb6c3243851fcd7d6f1b5cc3e1b97228

tests/unit/general/bin-update-ips-subnet.bats
Zash: Ah, never mind then.

Zash (approved):

Eliastisys (approved):

tests/unit/general/bin-update-ips-subnet.bats
aarnq, line 1: As this is openstack only I would have preferred it not being under general but that is not a blocker.

bin/update-ips.bash
```diff
@@ -402,13 +436,64 @@ allow_nodes() {
...
+
+  # TODO: Support for other CAPI providers could be added here by supporting
+  # other resource types than "openstackcluster".
+
+  # TODO: Currently Apps requires two clusters named "sc" and "wc" exactly.
+  # However, in cluster API it's common to refer to the cluster
+  # controlling CAPI resources the "management cluster" and therefore it's not
+  # uncommon that we name the service cluster "mc" instead of "sc" there. This
+  # code should be improved once we have better alignment between the
+  # repositories.
```
aarnq, line 474: We could ensure the global.ck8sClusterName would propagate properly, then this would be the perfect thing to look up.
Because the current solution won't work for multi cluster, and it will probably break.

Could you make it fallback to the old method if it cannot find a matching openstackcluster resource?

tests/unit/general/bin-update-ips-subnet.bats
simonklb: Ah that didn't cross my mind. Maybe we should make an effort into generating the tests for each provider and installer since the normal tests are only running for one specific as well (kubespray baremetal)?

aarnq: Potentially, but currently update-ips only have variants for "generic" and openstack capi, so I'm unsure if there are more than those two we need to focus on?
Or am I forgetting something?

bin/update-ips.bash
simonklb: > We could ensure the global.ck8sClusterName would propagate properly, then this would be the perfect thing to look up. Because the current solution won't work for multi cluster, and it will probably break.

Yea that's true, multiple workload clusters will fail here. I'm not aware of a `global.ck8sClusterName`, are you proposing adding it or am I just unaware of it?

> Could you make it fallback to the old method if it cannot find a matching openstackcluster resource?

Yep! https://github.com/elastisys/compliantkubernetes-apps/pull/2440/commits/d599ada604f85d01b44f36da764c6e7fa8192d5b

tests/unit/general/bin-update-ips-subnet.bats
simonklb: I don't _think_ there should be any difference, but that's what the tests would prove I guess. :smile: On the other hand the tests and the script is already complicated enough as it is so maybe an effort into trying to make this whole allow listing less complicated would be better time spent.

bin/update-ips.bash
aarnq: > > We could ensure the global.ck8sClusterName would propagate properly, then this would be the perfect thing to look up. Because the current solution won't work for multi cluster, and it will probably break.
> 
> Yea that's true, multiple workload clusters will fail here. I'm not aware of a `global.ck8sClusterName`, are you proposing adding it or am I just unaware of it?

It exists:
```yaml
# apps/config/sc-config.yaml
global:
```

But it remains with the `sc` suffix even if we are using `mc` for cluster api, so it unfortunately doesn't serve us right now.
We would have to update that in our setup process, and migrate current environments (which I don't even know if it is possible).

> > Could you make it fallback to the old method if it cannot find a matching openstackcluster resource?
> 
> Yep! [d599ada](https://github.com/elastisys/compliantkubernetes-apps/commit/d599ada604f85d01b44f36da764c6e7fa8192d5b)

:+1: Then it should continue to work as is for multi-cluster.

PR #1188, opened 2022-09-22
apps sc: Added network policies for the Logging stack

Xartos:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(use the format `fixes #<issue number>(, fixes #<issue_number>, ...)` to automatically close the issue when PR gets merged)*: fixes #1127 

**Public facing documentation PR** *(if applicable)*

**Special notes for reviewer**:

**Add a screenshot or an example to illustrate the proposed solution:**

**Checklist:**

- [ ] Added relevant notes to [WIP-CHANGELOG.md](https://github.com/elastisys/compliantkubernetes-apps/blob/main/WIP-CHANGELOG.md)
- [ ] Proper commit message prefix on all commits
- [ ] Updated the [public facing documentation](https://github.com/elastisys/compliantkubernetes)
- Is this changeset backwards compatible for existing clusters? Applying:
    - [ ] is completely transparent, will not impact the workload in any way.
    - [ ] requires running a migration script.
    - [ ] will create noticeable cluster degradation.
          E.g. logs or metrics are not being collected or Kubernetes API server
          will not be responding while upgrading.
    - [ ] requires draining and/or replacing nodes.
    - [ ] will change any APIs.
          E.g. removes or changes any CK8S config options or Kubernetes APIs.
    - [ ] will break the cluster.
          I.e. full cluster migration is required.
- Chart checklist (pick exactly one):
    - [ ] I upgraded no Chart.
    - [ ] I upgraded a Chart and determined that no migration steps are needed.
    - [ ] I upgraded a Chart and added [migration steps](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration).

**Pipeline config** *(if applicable)*
If you change some config options (e.g. add/rename variable or change the default value) you may need to update the config used by the pipeline in `pipeline/config`.

PR #2448, opened 2025-02-21
Make update-ips apply exit 0 even if it has applied diffs

simonklb:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Currently if you run `ck8s update-ips both apply` and a diff is applied the command exits with a non-zero status. That is incorrect because the command did what it was supposed to do successfully.

#### Information to reviewers

@Xartos added you as a reviewer well since this has been the behavior since the introduction of the command and just want to confirm that it wasn't added with some intention:
https://github.com/elastisys/compliantkubernetes-apps/pull/1188

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/bin/update-ips.bash
+++ b/bin/update-ips.bash
@@ -719,4 +719,6 @@ if rclone_enabled; then
   sync_rclone '.objectStorage.sync.secondaryUrl' '.networkPolicies.rclone.sync.secondaryUrl'
 fi
 
-exit ${has_diff}
+if ${dry_run}; then
+  exit ${has_diff}
+fi


bin/update-ips.bash
```diff
@@ -92,11 +92,10 @@ yq_eval() {
 
   if ${dry_run}; then
     log_warning "Diff found for ${config_option} in ${config_filename} (diff shows actions needed to be up to date)"
+    has_diff=$((has_diff + 1))
   else
     yq4 -i "${expression}" "${config_file}"
   fi
-
-  has_diff=$((has_diff + 1))
```
aarnq, line 99: Is there a reason why that is conditional here, rather than the final exit statement?

simonklb: I'm not sure I follow. Do you mean why it's not exiting right then and there? If so then I think it's because there can be multiple diffs.

aarnq: No, I'm wondering why we make the counting optional, rather than the exit with the count?
It make more sense to me to have the condition on the final exit statement to not propagate the count rather than make counting optional and still exit on the count.

Xartos (commented):
I think this was only to be able to detect if something was changed when doing a dry-run (like [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/apply.bash#L21))
But I'm not sure if this behavior is used during apply as well somewhere else after the fact. So while I agree to the change I think we should take a look and see if we have somewhere where it looks at the exit code.

bin/update-ips.bash
```diff
@@ -628,4 +628,4 @@ if rclone_enabled; then
   sync_rclone '.objectStorage.sync.secondaryUrl' '.networkPolicies.rclone.sync.secondaryUrl'
 fi
 
-exit ${has_diff}
+${dry_run} && exit ${has_diff}
```
aarnq, line 631: Can we avoid this `check && conditional` construct?
As it is an exception in behaviour compared to how we otherwise script given that we run with exit on err and pipefail.

And I know I use `check || conditional`, but its logic feels more sane as the final value is still handled under exit on err, while `false && nothing` is an exception to the rule.

simonklb:
> I think this was only to be able to detect if something was changed when doing a dry-run (like [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/apply.bash#L21)) But I'm not sure if this behavior is used during apply as well somewhere else after the fact. So while I agree to the change I think we should take a look and see if we have somewhere where it looks at the exit code.

Did some grepping in relevant repositories and wasn't able to find any other uses than the one you point out.

bin/update-ips.bash
simonklb: Yep! https://github.com/elastisys/compliantkubernetes-apps/pull/2448/commits/5cb2572d3e5a584dfaf6105331af780de8625fdb

aarnq (approved):

Xartos (approved):

PR #2450, opened 2025-02-25
add ipv6 support to update-ips

### LINKED ISSUE #2342: [3] Add IPv6 support for `update-ips` script
### Proposed feature

To support dual-stack, and potentially IPv6 only clusters, the [update-ips](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/update-ips.bash) should support adding IPv6 addresses.

### Proposed alternatives

Currently, I see that at least the following should be added:
- Getting IPv6 addresses from AAAA records (i.e. running `dig +short AAAA "${domain}"`)
- Getting host IPv6 addresses (`projectcalico.org/IPv6Address` annotation)
- Getting IPv6 tunnel addresses (`projectcalico.org/IPv6VXLANTunnelAddr` annotation)

Also, adding IPv6 netpols should only be done when needed, I can see three options for this:
- Either a global config option that tells that the environment is dual-stack
- The script checks the cluster if it supports IPv6 (dual-stack enabled or IPv6 only)
- Through a flag or argument for the `update-ips` script

### Additional context

Follow up task from https://github.com/elastisys/mse-internal-docs/issues/220

### Definition of done

- [ ] `update-ips` script supports IPv6

vomba:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Adds support for IPv6 to update-ips script.

- Fixes #2342 

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/bin/ck8s
+++ b/bin/ck8s
@@ -43,6 +43,7 @@ SYNC=""
 KUBECTL=""
 GEN_NEW_SECRETS=""
 CONCURRENCY="--concurrency=8"
+CK8S_IPV6_ENABLED=""
 
 for arg in "${@}"; do
   case "${arg}" in
@@ -50,6 +51,7 @@ for arg in "${@}"; do
   "--kubectl") KUBECTL="kubectl" ;;
   "--generate-new-secrets") GEN_NEW_SECRETS="--generate-new-secrets" ;;
   "--concurrency="*) CONCURRENCY="${arg}" ;;
+  "--ipv6") export CK8S_IPV6_ENABLED="true" ;;
   esac
 done
 
--- a/bin/ck8s
+++ b/bin/ck8s
@@ -43,7 +43,7 @@ SYNC=""
 KUBECTL=""
 GEN_NEW_SECRETS=""
 CONCURRENCY="--concurrency=8"
-CK8S_IPV6_ENABLED=""
+CK8S_IPV6_ENABLED="false"
 
 for arg in "${@}"; do
   case "${arg}" in
--- a/bin/ck8s
+++ b/bin/ck8s
@@ -51,7 +51,7 @@ for arg in "${@}"; do
   "--kubectl") KUBECTL="kubectl" ;;
   "--generate-new-secrets") GEN_NEW_SECRETS="--generate-new-secrets" ;;
   "--concurrency="*) CONCURRENCY="${arg}" ;;
-  "--ipv6") export CK8S_IPV6_ENABLED="true" ;;
+  "--ipv6") CK8S_IPV6_ENABLED="true" ;;
   esac
 done
 
@@ -139,7 +139,7 @@ explain)
 update-ips)
   [[ "${2}" =~ ^(wc|sc|both)$ ]] || usage
   [[ "${3}" =~ ^(apply|dry-run)$ ]] || usage
-  "${here}/update-ips.bash" "${2}" "${3}"
+  "${here}/update-ips.bash" "${2}" "${3}" "${CK8S_IPV6_ENABLED}"
   ;;
 fix-psp-violations)
   [[ "${2}" =~ ^(wc|sc)$ ]] || usage
--- a/bin/update-ips.bash
+++ b/bin/update-ips.bash
@@ -188,8 +188,9 @@ get_dns_ips() {
   local -a ips4
   local -a ips6
   mapfile -t ips4 < <(dig A +short "${domain}" | grep '^[.0-9]*$')
-  mapfile -t ips6 < <(dig AAAA +short "${domain}" | grep -E '^(\:\:)?[0-9a-fA-F]{1,4}(\:\:?[0-9a-fA-F]{1,4}){0,7}(\:\:)?$')
-
+  if [ -n "${CK8S_IPV6_ENABLED+x}" ]; then
+    mapfile -t ips6 < <(dig AAAA +short "${domain}" | grep -E '^(\:\:)?[0-9a-fA-F]{1,4}(\:\:?[0-9a-fA-F]{1,4}){0,7}(\:\:)?$')
+  fi
   local -a ips
   read -r -a ips <<<"${ips4[*]} ${ips6[*]}"
 
@@ -418,7 +419,6 @@ allow_domain() {
   local -a ips dns_ips
   dns_ips=$(get_dns_ips "${dns_record}")
   readarray -t ips <<<"$(echo "${dns_ips}" | tr ' ' '\n')"
-
   allow_ips "${config_file}" "${config_option}" "${ips[@]}"
 }
 
--- a/bin/update-ips.bash
+++ b/bin/update-ips.bash
@@ -23,7 +23,7 @@ has_diff=0
 #TODO: To be changed when decision made on networkpolicies for azure storage
 storage_service=$(yq4 '.objectStorage.type' "${CK8S_CONFIG_PATH}/defaults/common-config.yaml")
 
-CK8S_IPV6_ENABLED="${4}"
+CK8S_IPV6_ENABLED="${3:-"false"}"
 
 # Get the value of the config option or the provided default value if the
 # config option is unset.
@@ -143,7 +143,7 @@ get_tunnel_ips() {
   mapfile -t ips_calico_ipip < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv4IPIPTunnelAddr}')
   mapfile -t ips_wireguard < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv4WireguardInterfaceAddr}')
 
-  if [ -n "${CK8S_IPV6_ENABLED+x}" ]; then
+  if [ "${CK8S_IPV6_ENABLED}" = "true" ]; then
     mapfile -t ips6_calico_vxlan < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv6VXLANTunnelAddr}')
     mapfile -t ips6_calico_ipip < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv6IPIPTunnelAddr}')
   fi
@@ -192,7 +192,7 @@ get_dns_ips() {
   local -a ips4
   local -a ips6
   mapfile -t ips4 < <(dig A +short "${domain}" | grep '^[.0-9]*$')
-  if [ -n "${CK8S_IPV6_ENABLED+x}" ]; then
+  if [ "${CK8S_IPV6_ENABLED}" = "true" ]; then
     mapfile -t ips6 < <(dig AAAA +short "${domain}" | grep -E '^(\:\:)?[0-9a-fA-F]{1,4}(\:\:?[0-9a-fA-F]{1,4}){0,7}(\:\:)?$')
   fi
   local -a ips
--- a/bin/update-ips.bash
+++ b/bin/update-ips.bash
@@ -324,7 +324,7 @@ process_ips_to_cidrs() {
 
   for ip in "${@}"; do
     for cidr in "${old_cidrs[@]}"; do
-      if [[ "${cidr}" != "" ]] && [[ "${cidr}" != "set-me" ]] && ! [[ "${cidr}" =~ .*/32 ]] && check_ip_in_cidr "${ip}" "${cidr}"; then
+      if [[ "${cidr}" != "" ]] && [[ "${cidr}" != "set-me" ]] && ! [[ "${cidr}" =~ .*/32 ]] && ! [[ "${cidr}" =~ .*/128 ]] && check_ip_in_cidr "${ip}" "${cidr}"; then
         new_cidrs+=("${cidr}")
         continue 2
       fi
--- a/bin/update-ips.bash
+++ b/bin/update-ips.bash
@@ -134,13 +134,17 @@ get_tunnel_ips() {
 
   local -a ips_calico_ipip
   local -a ips_calico_vxlan
+  local -a ips6_calico_ipip
+  local -a ips6_calico_vxlan
   local -a ips_wireguard
   mapfile -t ips_calico_vxlan < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv4VXLANTunnelAddr}')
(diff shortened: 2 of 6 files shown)


simonklb (changes requested):
Nice work! The changes look good but since the number of `kubectl` and `dig` calls have changed you will have to update the mocked `kubectl` and `dig` calls in the tests.

I would like to see tests for both with and without ipv6 results as well.

bin/update-ips.bash
```diff
@@ -294,13 +324,17 @@ process_ips_to_cidrs() {
 
   for ip in "${@}"; do
     for cidr in "${old_cidrs[@]}"; do
-      if [[ "${cidr}" != "" ]] && [[ "${cidr}" != "set-me" ]] && ! [[ "${cidr}" =~ .*/32 ]] && check_ip_in_cidr "${ip}" "${cidr}"; then
+      if [[ "${cidr}" != "" ]] && [[ "${cidr}" != "set-me" ]] && ! [[ "${cidr}" =~ .*/32 ]] && ! [[ "${cidr}" =~ .*/128 ]] && check_ip_in_cidr "${ip}" "${cidr}"; then
```
Zash, line 327: Is this going to do the right thing with an IPv6 address like `2001:db8::/32` ?

vomba: It should stay the way it is.

Zash: I mean, when `cidr=2001:db8::/32` then `! [[ "${cidr}" =~ .*/32 ]]` → `! [[ /32 == /32 ]]` → `! true` → `false` and then the condition stops there. I assume that's only meant to happen for IPv4 `/32`

vomba: Hmm... I might be missing the issue here, but wouldn't that mean that the cidr just doesn't get processed, and therefore it gets added the way it is, which i think the behavior we want at that point ?

bin/update-ips.bash
```diff
@@ -263,7 +274,26 @@ get_swift_url() {
 #
 # For example, [1.0.0.10/32, 1.0.0.2/32] would be reordered to [1.0.0.2/32, 1.0.0.10/32].
 sort_cidrs() {
-  python3 -c "import ipaddress; import sys; cidrs = [ipaddress.ip_network(cidr) for cidr in sys.argv[1:]]; cidrs.sort(); [print(cidr) for cidr in cidrs]" "${@}"
+  python3 -c "
```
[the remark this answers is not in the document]
simonklb: AFAIK we only officially support Ubuntu 22.04 atm and Python should be included in the base. @aarnq Please correct me if I'm wrong. At least that is what the `REQUIREMENTS` are currently aiming at.

bin/update-ips.bash
simonklb: Yes, this is wrong. This should catch single IP CIDRs only, with the introduction of IPv6 `.*/32` is not doing that anymore.

bin/ck8s
```diff
@@ -43,13 +43,15 @@ SYNC=""
...
 CONCURRENCY="--concurrency=8"
+CK8S_IPV6_ENABLED=""
 
 for arg in "${@}"; do
   case "${arg}" in
   "--sync") SYNC="sync" ;;
   "--kubectl") KUBECTL="kubectl" ;;
   "--generate-new-secrets") GEN_NEW_SECRETS="--generate-new-secrets" ;;
   "--concurrency="*) CONCURRENCY="${arg}" ;;
+  "--ipv6") export CK8S_IPV6_ENABLED="true" ;;
```
simonklb, line 54: Instead of exporting it here please pass it as an argument to `update-ips.bash` for now to align with the other flags.

bin/update-ips.bash
```diff
@@ -134,19 +134,25 @@ get_tunnel_ips() {
...
+
   local -a ips
-  read -r -a ips <<<"${ips_calico_vxlan[*]} ${ips_calico_ipip[*]} ${ips_wireguard[*]}"
+  read -r -a ips <<<"${ips_calico_vxlan[*]} ${ips_calico_ipip[*]} ${ips6_calico_vxlan[*]} ${ips6_calico_ipip[*]} ${ips_wireguard[*]}"
 
   if [ ${#ips[@]} -eq 0 ]; then
     log_error "No IPs for ${cluster} nodes with label ${label} was found"
     exit 1
   fi
-
```
simonklb, line 149: Keep this linebreak

bin/update-ips.bash
```diff
@@ -171,7 +177,6 @@ get_internal_ips() {
...
```
simonklb, line 174: Keep this linebreak

bin/update-ips.bash
```diff
@@ -263,7 +274,26 @@ get_swift_url() {
...
+
+for cidr in cidrs:
+    if cidr.version == 6: cidrs6.append(cidr)
+
+cidrs4.sort()
+cidrs6.sort()
+
+cidrSorted = cidrs4 + cidrs6
+[print(cidr) for cidr in cidrSorted]
+" "${@}"
```
simonklb, line 296: Just out of curiosity, why do you have to split the lists? Does Python not support sorting a list that contains both IPv4 and IPv6 subnets?

bin/update-ips.bash
```diff
@@ -383,7 +417,6 @@ allow_domain() {
   local -a ips dns_ips
   dns_ips=$(get_dns_ips "${dns_record}")
   readarray -t ips <<<"$(echo "${dns_ips}" | tr ' ' '\n')"
-
```
simonklb, line 386: Keep this linebreak

bin/update-ips.bash
```diff
@@ -440,7 +473,6 @@ allow_nodes() {
   internal_ips=$(get_internal_ips "${cluster}" "${label}")
   tunnel_ips=$(get_tunnel_ips "${cluster}" "${label}")
   readarray -t ips <<<"$(echo "${internal_ips}" "${tunnel_ips}" | tr ' ' '\n')"
-
```
simonklb, line 443: Keep this linebreak

tests/common/bats/update-ips.bash
```diff
@@ -188,6 +210,20 @@ update_ips.populate_minimal() {
...
+
+#   yq.set wc .networkPolicies.global.wcApiserver.ips '["127.0.1.4/32", "127.0.1.5/32", "127.0.1.6/32", "127.0.1.51/32", "127.0.2.4/32", "127.0.2.5/32", "127.0.2.6/32", "127.0.2.51/32", "127.0.3.4/32", "127.0.3.5/32", "127.0.3.6/32", "127.0.3.51/32", "fd3e:fab4:5eda:b233::18/128", "fd3e:fab4:5eda:b233::19/128", "fd3e:fab4:5eda:b233::20/128", "fd3e:fab4:5eda:b233::21/128", "fd3e:fab4:5eda:b233::22/128", "fd3e:fab4:5eda:b233::23/128"]'
+#   yq.set wc .networkPolicies.global.wcNodes.ips '["127.0.1.10/32", "127.0.1.11/32", "127.0.1.12/32", "127.0.1.111/32", "127.0.2.10/32", "127.0.2.11/32", "127.0.2.12/32", "127.0.2.111/32", "127.0.3.10/32", "127.0.3.11/32", "127.0.3.12/32", "127.0.3.111/32", "fd3e:fab4:5eda:b233::24/128", "fd3e:fab4:5eda:b233::25/128", "fd3e:fab4:5eda:b233::26/128", "fd3e:fab4:5eda:b233::27/128", "fd3e:fab4:5eda:b233::28/128", "fd3e:fab4:5eda:b233::29/128"]'
+# }
+
```
simonklb, line 226: Remove?

simonklb (changes requested):
Just minor things left to fix up, nice work!

I would have liked to see a test of `maximal run dualStack` as well but if you feel that you are about to claw your eyes out seeing another bats run then I'm fine with you skipping it for now. :smile:  There is some work needed to be done here to make it easier to test.

bin/update-ips.bash
vomba: Yep, exactly, sorting them both is not supported.

Zash: Might simplify some things by using [IPv6-mapped IPv4 addresses](https://en.wikipedia.org/wiki/IPv6_address#Representation) (like `::ffff:xxxx:xxxx/xx`), thereby having a single canonical form to deal with, then undo it for output.

bin/update-ips.bash
Zash: Maybe like this

```suggestion
      if [[ "${cidr}" != "" ]] && [[ "${cidr}" != "set-me" ]] && ! [[ "${cidr}" =~ [0-9.]*/32 ]] && ! [[ "${cidr}" =~ [0-9a-f:]*/128 ]] && check_ip_in_cidr "${ip}" "${cidr}"; then
```

Alternatively, but probably more work, only dealing with IPv6-mapped IPv4 addresses would need only a single `/128` comparison.

bin/update-ips.bash
simonklb: > Might simplify some things by using [IPv6-mapped IPv4 addresses](https://en.wikipedia.org/wiki/IPv6_address#Representation) (like `::ffff:xxxx:xxxx/xx`), thereby having a single canonical form to deal with, then undo it for output.

Feel free to send a patch if you have a way to simplify this or anything else in the script! Let's not introduce more work for this specific PR task though.

bin/update-ips.bash
vomba: That looks good, will add it.

bin/update-ips.bash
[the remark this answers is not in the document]
simonklb: Yea but Python is also included in 24.04. What I'm saying is that we should only add packages that need to be installed on top of a base Ubuntu 24.04 installation, not try to list every package required in an environment like Linux From Scratch. :smile:

bin/update-ips.bash
```diff
@@ -23,6 +23,8 @@ has_diff=0
 #TODO: To be changed when decision made on networkpolicies for azure storage
 storage_service=$(yq4 '.objectStorage.type' "${CK8S_CONFIG_PATH}/defaults/common-config.yaml")
 
+CK8S_IPV6_ENABLED="${4}"
```
simonklb, line 26: To make it backwards compatible you can do this:
```suggestion
CK8S_IPV6_ENABLED="${4:-"false"}"
```

bin/ck8s
```diff
@@ -43,13 +43,15 @@ SYNC=""
 KUBECTL=""
 GEN_NEW_SECRETS=""
 CONCURRENCY="--concurrency=8"
+CK8S_IPV6_ENABLED=""
```
simonklb, line 46: Then you need to to this:
```suggestion
CK8S_IPV6_ENABLED="false"
```

bin/update-ips.bash
```diff
@@ -180,13 +189,20 @@ get_internal_ips() {
 # Usage: get_dns_ips <domain>
 get_dns_ips() {
   local domain="${1}"
-
+  local -a ips4
+  local -a ips6
+  mapfile -t ips4 < <(dig A +short "${domain}" | grep '^[.0-9]*$')
+  if [ -n "${CK8S_IPV6_ENABLED+x}" ]; then
```
simonklb, line 195: And this:
```suggestion
  if [ "${CK8S_IPV6_ENABLED}" = "true" ]; then
```

bin/update-ips.bash
```diff
@@ -134,13 +136,20 @@ get_tunnel_ips() {
...
   local -a ips_calico_ipip
   local -a ips_calico_vxlan
+  local -a ips6_calico_ipip
+  local -a ips6_calico_vxlan
   local -a ips_wireguard
   mapfile -t ips_calico_vxlan < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv4VXLANTunnelAddr}')
   mapfile -t ips_calico_ipip < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv4IPIPTunnelAddr}')
   mapfile -t ips_wireguard < <("${here}/ops.bash" kubectl "${cluster}" get node "${label_argument}" -o jsonpath='{.items[*].metadata.annotations.projectcalico\.org/IPv4WireguardInterfaceAddr}')
 
+  if [ -n "${CK8S_IPV6_ENABLED+x}" ]; then
```
simonklb, line 146:
```suggestion
  if [ "${CK8S_IPV6_ENABLED}" = "true" ]; then
```

tests/unit/general/bin-update-ips.bats
```diff
@@ -93,6 +93,14 @@ _apply_normalise() {
   update_ips.assert_minimal
 }
 
+@test "ck8s update-ips performs minimal run dualStack" {
+  update_ips.mock_minimal_v6
+
+  CK8S_IPV6_ENABLED=true run ck8s update-ips both apply
```
simonklb, line 99:
```suggestion
  ck8s update-ips both apply true
```

bin/update-ips.bash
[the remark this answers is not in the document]
aarnq: I missed that part, then yes for now Python doesn't need to be included as the only target is a regular Ubuntu installation.
But this might change in the future if we change the main target to their base container to simplify building images.

bin/update-ips.bash
```diff
@@ -294,13 +329,17 @@ process_ips_to_cidrs() {
 
   for ip in "${@}"; do
     for cidr in "${old_cidrs[@]}"; do
-      if [[ "${cidr}" != "" ]] && [[ "${cidr}" != "set-me" ]] && ! [[ "${cidr}" =~ .*/32 ]] && check_ip_in_cidr "${ip}" "${cidr}"; then
+      if [[ "${cidr}" != "" ]] && [[ "${cidr}" != "set-me" ]] && ! [[ "${cidr}" =~ [0-9.].*/32 ]] && ! [[ "${cidr}" =~ [0-9a-f:].*/128 ]] && check_ip_in_cidr "${ip}" "${cidr}"; then
```
Zash, line 332: `[0-9.].*` only ensures the first thing is a number or `.`, while `[0-9.]*` only matches numbers and literal `.`. 

```suggestion
      if [[ "${cidr}" != "" ]] && [[ "${cidr}" != "set-me" ]] && ! [[ "${cidr}" =~ [0-9.]*/32 ]] && ! [[ "${cidr}" =~ [0-9a-f:]*/128 ]] && check_ip_in_cidr "${ip}" "${cidr}"; then
```

Zash (changes requested):

simonklb (approved):
After you have fixed the thing pointed out by @Zash this LGTM, great work!

Zash (approved):

Issue #2318, opened 2024-10-18
[3] Add Safeguard for pods that prevent CAPI cluster-autoscaler


anders-elastisys:
### Proposed feature

[The cluster-autoscaler documentation contains a list](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node) of what types of pods can prevent the autoscaler from scaling down nodes. Would like some OPA Gatkeeper constraints that warns or restricts (depending on the constraint) Application Developers when they try to deploy such pods in environments with cluster-autoscaler (e.g. CAPI). We currently already have a lot of safeguards that would prevent Application Developers from deploying such pods, e.g. constraint for `hostPath`, but at least non-memory `emptyDir` are still allowed which can prevent the autoscaler. If this gets added, the [public documentation should be updated with the new safeguard(s)](https://elastisys.io/compliantkubernetes/user-guide/safeguards/) and also [this notice](https://github.com/elastisys/compliantkubernetes/pull/976/files) should probably be updated as well.

### Proposed alternatives

- Don't add safeguards for this
- Add some operator that labels _preventing_ pods with `safe-to-evict` labels automatically in Application Developer namespaces

### Additional context

https://github.com/elastisys/ck8s-issue-tracker/issues/313

### Definition of done

- [ ] Added safeguards for Application Developer pods as to not prevent cluster-autoscaler from scaling down clusters
- [ ] The public documentation is updated with the Safeguard(s)

PR #2459, opened 2025-03-11
Add opa policy to restrict PDBs, always allow at least 1 disruption

viktor-f:
> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change   
- [x] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

A new gatekeeper policy has been added that will deny any PodDisruptionBudget and connected Pod controller if the PodDisruptionBudget does not allow at least 1 Pod disruption. Note that this will apply in both sc and wc, it will also apply to namespaces even if they have the label owner=operator.

### Application Developer notice

A new gatekeeper policy has been added that will deny any PodDisruptionBudget and connected Pod controller if the PodDisruptionBudget does not allow at least 1 Pod disruption.

This adds a new gatekeeper policy that will deny any PodDisruptionBudget and connected Pod controller if the PodDisruptionBudget does not allow at least 1 Pod disruption. Note that this will apply in both sc and wc, it will also apply to namespaces even if they have the label owner=operator.

Pod controllers only includes: Deployment, ReplicaSet, StatefulSet, ReplicaController.

The general logic for this is:
```
If creating or modifying PDB
  If maxUnavailable == 0 or "0%"
    deny request
  If minAvailable
    If any matching pod controller
      If minAvailable is >= number of replicas
        deny request
If creating or modifying pod controller
  If any matching PDB
    Use the logic above under "IF PDB"
```
In sentences this means that we want to stop either PDBs or pod controllers if they together do not allow for any disruption. If the PDB is using maxUnavailable, then this happens if it is set to 0 or 0%, regardless of the replicas in the pod controller. If the PDB is using minAvailable, then this happens if it is equal or higher than the replicas in the pod controller. The policy includes logic to stop requests that would create or edit both PDBs or pod controllers.

There is also an extra check to allow replicasets to violate this policy if they are controlled by a deployment. This is because any matching PDB will look at all of the pods for the deployment, not any single replicaset.

The logic has been based on an [upstream policy](https://open-policy-agent.github.io/gatekeeper-library/website/validation/poddisruptionbudget/), but it has almost entirely been reworked. The logic of matching selectors has been based on the logic in our [networkpolicy gatekeeper policy](https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile.d/charts/gatekeeper/templates/policies/network-policies.rego).

I was planning on adding more checks to this. But this took longer to implement than planned, so I'm now planning on turning the rest into a separate task. The things that were not implemented are:
- Denying pods without pod controllers that are matching a PDB. This is not a pattern we want to allow.
- Denying PDBs and matching daemonsets/jobs/cronjobs if the PDB is matching a daemonsets/jobs/cronjobs. This is not a pattern we want to allow.
- Denying PDBs and pods/podcontrollers if the PDB is matching multiple pod controllers or pods without controllers. This is not something we want to allow because "The eviction API will disallow eviction of any pod covered by multiple PDBs, so most users will want to avoid overlapping selectors." ([ref](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors))

This PR adds more resources that will be synced (cached) by gatekeeper. This is needed so that we can compare PDBs to pod controllers (otherwise you just have access to the object that is being validated). 
But adding resources here will increase the resource usage (primarily memory) for gatekeeper. Syncing PDBs will likely not increase the usage significantly, since there are relatively few PDBs in a cluster. However pod controllers are a lot more common and will noticeably increase the memory usage, but as I show below I think that the usage has not increased so much more that it should prevent us from adding this feature. I did some testing to see how much the resource usage increased. Note that some of the testing included pods, because that would be needed for one of the extra features mentioned above that I was planning on adding but am now skipping.

> Number of resources in the cluster:
> Pods: 240
> Deployments: 55
> StatefulSets: 10
> ReplicaSets: 73
> ReplicationController: 0
> 
> Note: many small pod manifests in this test, actual usage with "real pods" is likely higher (tested in scenario 4 and 5 below)
> 
> Without pods or controllers:
>   Gatekeeper controllers 140 MB
>   Gatekeeper audit 160 MB, 215 MB spikes
> With controllers:
>   Gatekeeper controllers 160 MB
>   Gatekeeper audit 175 MB idle, 250 MB spikes
> With pods and controllers:
>   Gatekeeper controllers 180 MB
>   Gatekeeper audit 200 MB idle, 300 MB spikes
> With pods and controllers and extra annotations (to get larger pod manifests):
>   Gatekeeper controllers 190MB
>   Gatekeeper audit 210 MB idle, 320 MB spikes
> With pods and controllers and extra annotations on a larger cluster (1000 pods, 140 deployments, 160 replicasets)
>   Gatekeeper controllers 290MB idle, 370 MB spikes
>   Gatekeeper audit 310 MB idle, 520 MB spikes

- Part of #2318 

Even though this is part of an autoscaling task, this policy is good to have for any type of cluster. Since PDBs can disrupt any type of scenario where we want to drain nodes or similar.

#### Information to reviewers

As of creating the PR I have not yet added tests or public documentation. I will work on that next, but the code and config should be ok to review for now.

This PR also included some whitespace fixes that I stumbled upon. I hope it is ok that is is fixed in this PR. Otherwise let me know and I will move that to a separate PR.

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:


helmfile.d/charts/gatekeeper/constraints/templates/restrict-pod-disruption-budgets/constraint.yaml
```diff
@@ -0,0 +1,22 @@
...
+        kinds: ["Pod", "ReplicationController"]
+      - apiGroups: ["apps"]
+        kinds: ["Deployment", "StatefulSet", "DaemonSet", "ReplicaSet"]
+      - apiGroups: ["extensions"]
+        kinds: ["Deployment", "StatefulSet", "DaemonSet", "ReplicaSet"]
+      - apiGroups: ["batch"]
+        kinds: ["Job", "CronJob"]
+      - apiGroups: ["policy"]
+        kinds: ["PodDisruptionBudget"]
+    excludedNamespaces: ["kube-system", "kube-public", "kube-node-lease", "calico-system"]
```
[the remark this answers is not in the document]
viktor-f: Yes, this makes sense. I will write another task for this, since it is a bit out of scope for this PR.

helmfile.d/charts/gatekeeper/templates/policies/restrict-pod-disruption-budgets.rego
```diff
@@ -0,0 +1,174 @@
...
+
+	replicas <= min_available
+}
+
+not_valid_pdb_max_unavailable(pdb) {
+    pdb.spec.maxUnavailable == 0
+}
+
+not_valid_pdb_max_unavailable(pdb) {
+    pdb.spec.maxUnavailable == "0%"
```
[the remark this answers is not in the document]
viktor-f: Yes, that is at least what the [upstream documentation](https://kubernetes.io/docs/tasks/run-application/configure-pdb) says.


> When you specify the value as a percentage, it may not map to an exact number of Pods. For example, if you have 7 Pods and you set minAvailable to "50%", it's not immediately obvious whether that means 3 Pods or 4 Pods must be available. Kubernetes rounds up to the nearest integer, so in this case, 4 Pods must be available. When you specify the value maxUnavailable as a percentage, Kubernetes rounds up the number of Pods that may be disrupted. Thereby a disruption can exceed your defined maxUnavailable percentage. You can examine the [code](https://github.com/kubernetes/kubernetes/blob/23be9587a0f8677eb8091464098881df939c44a9/pkg/controller/disruption/disruption.go#L539) that controls this behavior.

> If you set maxUnavailable to 0% or 0, or you set minAvailable to 100% or the number of replicas, you are requiring zero voluntary evictions.

davidumea:
What happens if a pdb that violates this already exists? Will it have to be manually removed?

davidumea:
> Denying PDBs and matching daemonsets/jobs/cronjobs if the PDB is matching a daemonsets/jobs/cronjobs. This is not a pattern we want to allow.

This was one of the patterns you mentioned that was not included in this PR, was it written correctly? If yes, could you expand?

---

I like that the PR description is very extensive and informative, thanks for that! I think it would be nice if you wrote in text format as well which patterns to not allow were added in PR.

> This adds a new gatekeeper policy that will deny any PodDisruptionBudget and connected Pod controller if the PodDisruptionBudget does not allow at least 1 Pod disruption.

I think this is good information but it doesn't explain what is actually happening under the hood :slightly_smiling_face:

viktor-f:
> > Denying PDBs and matching daemonsets/jobs/cronjobs if the PDB is matching a daemonsets/jobs/cronjobs. This is not a pattern we want to allow.
> 
> This was one of the patterns you mentioned that was not included in this PR, was it written correctly? If yes, could you expand?

I think the text is accurate, but maybe not very easy to read. Regardless, the idea is that I think PDBs should not be allowed for daemonsets, cronjobs, and jobs.
The part of the reasoning for this is that the PDBs does not have full functionality on these resources ([on any resource that does not have the scale api](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors)).
For daemonsets it is also odd to have a PDB since the number of pods will change with the number of nodes and the number of nodes might not be controlled by the users that specify the PDB.
For Jobs and cronjobs the purpose is to at some point have the jobs complete, then the number of pods would naturally reduce and you will then eventually violate the PDB.

Does that make sense?

> 
> I like that the PR description is very extensive and informative, thanks for that! I think it would be nice if you wrote in text format as well which patterns to not allow were added in PR.
> 
> > This adds a new gatekeeper policy that will deny any PodDisruptionBudget and connected Pod controller if the PodDisruptionBudget does not allow at least 1 Pod disruption.
> 
> I think this is good information but it doesn't explain what is actually happening under the hood 🙂

Thanks, I will try to clarify this.

viktor-f:
> Thanks, I will try to clarify this.

I have now updated the PR description with some more details. Please let me know if this explains it well or if I should clarify further.

helmfile.d/charts/gatekeeper/constraints/templates/restrict-pod-disruption-budgets/constraint.yaml
```diff
@@ -0,0 +1,22 @@
...
+  match:
+    kinds:
+      - apiGroups: [""]
+        kinds: ["Pod", "ReplicationController"]
+      - apiGroups: ["apps"]
+        kinds: ["Deployment", "StatefulSet", "DaemonSet", "ReplicaSet"]
+      - apiGroups: ["extensions"]
+        kinds: ["Deployment", "StatefulSet", "DaemonSet", "ReplicaSet"]
+      - apiGroups: ["batch"]
+        kinds: ["Job", "CronJob"]
```
davidumea, line 17: These do nothing at the moment right? Same with daemonset

helmfile.d/charts/gatekeeper/templates/templates/config.yaml
```diff
@@ -9,3 +9,18 @@ spec:
...
+        kind: "ReplicaSet"
+      - group: "apps"
+        version: "v1"
+        kind: "StatefulSet"
+      - group: ""
+        version: "v1"
+        kind: "ReplicationController"
+      - group: "policy"
+        version: "v1"
+        kind: "PodDisruptionBudget"
```
davidumea, line 26: If you want to keep the kinds mentioned in my previous comment for future work, then please add them here as well :slightly_smiling_face: might save a future headache.

davidumea (commented):
> This PR also included some whitespace fixes that I stumbled upon. I hope it is ok that is is fixed in this PR. Otherwise let me know and I will move that to a separate PR.

I think this is fine, thanks for the cleanup!

> I have now updated the PR description with some more details. Please let me know if this explains it well or if I should clarify further.

Thanks it's crystal clear now!

I plan on continuing my review tomorrow

helmfile.d/charts/gatekeeper/constraints/templates/restrict-pod-disruption-budgets/constraint.yaml
viktor-f: Yeah, this was just something I forgot to remove when I reduced the scope of the task.
Removed now.

helmfile.d/charts/gatekeeper/templates/templates/config.yaml
viktor-f: I removed the kinds from the constraint.
As extra info: it would actually not be good to add the extra kinds here before they are needed, because that would cause gatekeeper to cache those kinds and increase the resource of the gatekeeper pods.

viktor-f:
Added a bunch of tests now. I have probably missed some tests, but this should cover most things. I at least have full code coverage:
```console
❯ opa test restrict-pod-disruption-budgets.rego tests/restrict-pod-disruption-budgets.rego -v -c
{
  "files": {
    "restrict-pod-disruption-budgets.rego": {
      ...
      "covered_lines": 98,                                                                                                                                  
      "coverage": 100                                                                                                                                       
    },                                                                                                                                                      
    "tests/restrict-pod-disruption-budgets.rego": {
      ...
      "covered_lines": 257,
      "coverage": 100
    }
  },
  "covered_lines": 355,
  "not_covered_lines": 0,
  "coverage": 100
}
```

helmfile.d/charts/gatekeeper/templates/policies/tests/restrict-pod-disruption-budgets.rego
```diff
@@ -0,0 +1,742 @@
...
+                    "labels": labels
+                }
+            }
+        }
+    }
+}
+
+input_wrap(obj) = input {
+    input := {"review": {"object": obj}}
+}
```
davidumea, line 73: Nice detail :slightly_smiling_face:

helmfile.d/charts/gatekeeper/templates/policies/restrict-pod-disruption-budgets.rego
```diff
@@ -0,0 +1,181 @@
...
+    )
+}
+
+# The type of pod controller to validate
+pod_controller_groups_kinds := [
+    {"group": "apps/v1", "kind": "Deployment"},
+    {"group": "apps/v1", "kind": "StatefulSet"},
+    {"group": "apps/v1", "kind": "ReplicaSet"},
+    {"group": "v1", "kind": "ReplicationController"}
+]
```
davidumea, line 84: You only use the group here in 1 place (below), might want to split these into separate functions, one for group and one for kind.

https://github.com/elastisys/compliantkubernetes-apps/blob/8d6c09e4021c9233498844e31b1d633cbd1e4e13/helmfile.d/charts/gatekeeper/templates/policies/restrict-pod-disruption-budgets.rego#L26

Or do they need to be together to make sure it's the same object? If I understand it correctly it should take one object at a time so even if the information is fetched from two functions that shouldn't be a problem.

helmfile.d/charts/gatekeeper/templates/policies/restrict-pod-disruption-budgets.rego
```diff
@@ -0,0 +1,181 @@
...
+
+# Do not reject replicasets that are controlled by deployment, instead reject the deploymentd
```
davidumea, line 86:
```suggestion
# Do not reject replicasets that are controlled by deployment, instead reject the deployment
```

helmfile.d/charts/gatekeeper/templates/policies/tests/restrict-pod-disruption-budgets.rego
```diff
@@ -0,0 +1,742 @@
...
+
+test_matching_match_labels {
+	namespace := "default"
+    selector := {"matchLabels": {"key": "val"}}
+    labels := {"key": "val"}
+
+    count(k8srestrictpoddisruptionbudgets.violation) == 1
+        with input as input_wrap(generate_pod_controller(
+            "Deployment",
+			3,
```
davidumea, line 474: Indentation is a bit off here, right?

Same with the replica integer for most tests below

helmfile.d/charts/gatekeeper/templates/policies/restrict-pod-disruption-budgets.rego
viktor-f: The idea was to ensure that it just uses these pairs of group and kind. This was the easiest way that I could think of to do this, but there are probably other ways.
I could also just have two separate lists of groups and kinds and then let it go through all combinations. I assume that is slightly less efficient, but probably not significantly.

IMO the current version feels nice to read and understand which groups and kinds belong to each other. But I think we should use the version that most people find easiest to read and understand. So I'm ok with changing it if that is what you and others want.

helmfile.d/charts/gatekeeper/templates/policies/tests/restrict-pod-disruption-budgets.rego
viktor-f: Thanks, the issues was that I had tabs instead of spaces here.

For some reason my vscode sometimes wants to use tabs instead of spaces for the rego files. I have tried to fix it, but not quite managed yet.

helmfile.d/charts/gatekeeper/templates/policies/restrict-pod-disruption-budgets.rego
davidumea: Okay, I would be fine with leaving it but I don't think this reads super well when referencing this `pod_controller_group_kind.group`

davidumea (commented):
I think the code looks good. Really nice work :slightly_smiling_face: 

I assume the plan is to add public docs and update the links here before you want to merge this?

viktor-f:
Public docs PR is now up as well https://github.com/elastisys/welkin/pull/1073
The code in this PR have been updated with links to the new public docs page (will not work until the public docs page is merged).
So I think that should be the last thing needed. Except for any additional comments you reviewers might have.

helmfile.d/charts/gatekeeper/templates/policies/restrict-pod-disruption-budgets.rego
```diff
@@ -0,0 +1,181 @@
...
+
+# Reject PDB if maxUnavailable does not allow at least 1 pod disruption
+violation[{"msg": msg}] {
+    input.review.object.kind == "PodDisruptionBudget"
+    pdb := input.review.object
+
+    pdb.spec.maxUnavailable
+
+    not_valid_pdb_max_unavailable(pdb)
+    # TODO update messages with link to public docs
```
davidumea, line 11:
```suggestion
```

helmfile.d/charts/gatekeeper/templates/policies/tests/restrict-pod-disruption-budgets.rego
```diff
@@ -0,0 +1,742 @@
...
+
+    res = [test | test = count(k8srestrictpoddisruptionbudgets.violation) == 0
+        with input as input_wrap(generate_pdb(
+            "minAvailable",
+            1,
+            namespace,
+            selector
+        )) with data.inventory as inventories[_]
+    ]
+    # TODO consider adding a count so that it is certain how many test cases that are run
```
davidumea, line 162: Was this TODO for future improvements or was it something you wrote for yourself here?

helmfile.d/charts/gatekeeper/templates/policies/tests/restrict-pod-disruption-budgets.rego
```diff
@@ -0,0 +1,742 @@
...
+        ))
+        with data.inventory as {"namespace": {namespace: {"policy/v1": {"PodDisruptionBudget": [generate_pdb(
+            "minAvailable",
+            3,
+            namespace,
+            selector
+    )]}}}}
+}
+
+
```
davidumea, line 704: nit
```suggestion
```

helmfile.d/charts/gatekeeper/templates/policies/tests/restrict-pod-disruption-budgets.rego
viktor-f: Right, forgot about that one.
I was considering doing that now. But I'm not sure if it is needed or not, or even if it makes it easier to read.
Is this something that you would appreciate? It should be rather easy to add.

davidumea: I don't think it's necessary, it should be clear if a test fails either way

Do as you wish :slightly_smiling_face: but please remove the comment

viktor-f: Added now

helmfile.d/charts/gatekeeper/templates/policies/restrict-pod-disruption-budgets.rego
viktor-f: I will keep it like this for now then.

viktor-f:
Task for the scenarios that were not covered in this PR: https://github.com/elastisys/compliantkubernetes-apps/issues/2476

PR #2449, opened 2025-02-21
Add Crossplane multi-module PoC

simonklb:
:warning: THIS SHOULD NOT BE MERGED :warning: 

### What does this PR do / why do we need this PR?

This is a proof of concept of modularization OpenSearch and the Fluentd Forwarder. See the [previous PR](https://github.com/elastisys/compliantkubernetes-apps/pull/2409) for more basic information about how we use Crossplane in these PoCs.

This primarily focuses on showcasing the capability idea where one module is able to depend on another and retrieve necessary configuration. Specifically here the FluentdForwarder module depends on the `log-write` capability which in this case references the connection details necessary to write logs to OpenSearch.

#### Information to reviewers

This chart shows the primary flow of the connection details from the OpenSearch module to the FluentdFowarder module:

```mermaid
flowchart TD
 subgraph s1["MC"]
        n1["Composite: OpenSearch"]
        n3["<b>Secret: FluentdCredentials<br></b>* Username<br>* Password"]
        n4["<b>Ingress: opensearch-client<br></b>* Hostname"]
        n6["<b>Connection Details<br></b>* Hostname<br>* Username<br>* Password"]
        n9["Helm Release: opensearch-client"]
        n7["Helm Chart: module-opensearch"]
  end
 subgraph s2["WC"]
        n2["Composite: fluentd-forwarder"]
        n11["Helm Release: fluentd-forwarder"]
        n12["EnvironmentConfig: capability-log-write"]
        n13["Helm Chart: module-fluentd-forwarder"]
        n14["<b>Connection Details<br></b>* Hostname<br>* Username<br>* Password"]
  end
    n7 -- Create --> n1 & n3
    n1 -- Create --> n12 & n9
    n1 -- Generate --> n6
    n3 -- Read --> n1
    n9 -- Create --> n4
    n4 -- Read --> n1
    n2 -- Create --> n11
    n12 -- Read --> n2
    n13 -- Create --> n2
    n1 -- Forward --> n14
    n12 -- Reference --> n14
    n3@{ shape: doc}
    n4@{ shape: doc}
    n6@{ shape: doc}
    n9@{ shape: rect}
    n7@{ shape: rect}
    n2@{ shape: rect}
    n11@{ shape: rect}
    n12@{ shape: rect}
    n13@{ shape: rect}
    n14@{ shape: doc}
     n3:::Sky
     n4:::Aqua
     n4:::Ash
     n4:::Sky
     n6:::Sky
     n14:::Sky
[...]

PR #2463, opened 2025-03-11
Add alternative Crossplane multi-module PoC

simonklb:
:warning: THIS SHOULD NOT BE MERGED :warning: 

This is a proof of concept of an alternative log-write capability implementation. See the [previous PR](https://github.com/elastisys/compliantkubernetes-apps/pull/2449) for more details on how the modules and the connection details are created.

The biggest difference between the previous PoC and this one is the log-write capability model. In previous PoC the capability was baked in to the modules while in this PoC the capability is a standalone resource. The capability resource detects _producers_  and _consumers_, where a producer is a module that provides the capability of writing logs (a log sink) and the consumers are secrets that should be patched with the credentials and connection details necessary to perform the log writing.

#### Information to reviewers

I hope I added everyone that was part of the previous demo. Feel free to assign more reviewers.

This chart shows the main flow of the log-write capability mechanism:

```mermaid
flowchart TD
 subgraph s1["MC"]
        n1["Composite: LogWrite"]
        n6["Composite: OpenSearch"]
        n16["Connection Details"]
  end
 subgraph s2["WC"]
        n2["Composite: fluentd-forwarder"]
        n3["Composite: LogWrite"]
        n12["EnvironmentConfig: capability-log-write-consumers"]
        n14["Secret: opensearch-credentials"]
  end
    n6 -- Generate --> n16
    n2 -- Create --> n14
    n3 -- Detect consumer --> n14
    n12 -- Reference --> n14
    n3 -- Create --> n12
    n1 -- Get remote consumers --> n12
    n1 -- Patch consumer --> n14
    n1 -- Read --> n16
    n1 -- Detect producer --> n6
```

Worth noting is that there are some obvious security concerns with this solution, the primary one being that anyone can create a consumer secret and gain access to OpenSearch. In a real implementation of this model this would have to be mitigated. RBAC and policies for who has access to what and who is allowed to be a consumer are likely candidates.

Another security concern is that the implementation relies heavily on the [Kubernetes Crossplane Provider](https://github.com/crossplane-contrib/provider-kubernetes) resource `Object`, which when referencing a `Secret` will leak the secret data. This is also something that would have to be mitigated before promoting this further.


modules/capability-log-write/apis/composition.yaml
```diff
@@ -0,0 +1,188 @@
...
+                "secrets" list
+              }}
+              {{- range $_, $secret := $remoteConsumers }}
+                  {{- $cluster = set $cluster "secrets" (append $cluster.secrets (dict "name" $secret.name "namespace" $secret.namespace)) }}
+              {{- end }}
+              {{- $clusters = append $clusters $cluster }}
+            {{- end }}
+            {{- end }}
+            {{- end }}
+            {{- end }}
```
simonklb, line 137: This is a lot of logic that, in a proper implementation, should be implemented in a function rather than in Go templating. :smile: Also worth considering if Module Capabilities might require enough functionality to warrant it's own Crossplane Provider / Kubernetes controller.

simonklb:
Demoed

PR #2490, opened 2025-04-03
Prepare for Helmfile v1 / Introduce default disabled environment

aarnq:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

This ensures that we are ready for Helmfile v1 when it drops, currently tested with latest RC 12).

Additionally it cleans up conditions and installed checks so that all warnings about needs not being met should be removed (with Helmfile v1).
Essentially everything that is disabled via conditions should also not be installed.

Finally it adds a `default` environment in which all releases are disabled, this will be used in tests which needs to know about all releases.

One thing I discovered was that Felix Metrics for Calico probably never worked for Cluster API as it was installed in the wrong namespace.

...

- Fixes #

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:


simonklb (approved):
Great work!

What could have caught the failed Felix metrics deployment? Can we add a regression test to prevent it from happening again?

helmfile.d/stacks/calico.yaml.gotmpl
vomba, line 1: I think this might need a migration step. I tried running this on a CAPI-cluster and faced this error.
```sh
Error: Unable to continue with install: Service "calico-felix-metrics-svc" in namespace "kube-system" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "calico-system": current value is "kube-system"
```

aarnq: Yes, and it seems to still get into the wrong namespace :thinking:

aarnq:
> What could have caught the failed Felix metrics deployment? Can we add a regression test to prevent it from happening again?

The best test would be to check that it generated expected metrics. But I feel like that is out of scope here, as it would take some time to come up with a good method to deal with it.
But perhaps the dealings with template tests will create something reusable, so the only thing we need is to deploy and fetch the metrics.

Although, it would need to be an end to end test as we do not include installers at other levels.

Zash (approved):
great work

really nice that helmfile v1 enforces that templates be clearly named as such

I note that there exists some `.gotmpl` files that contain no templating, does it make sense to drop the extension there?

aarnq:
> really nice that helmfile v1 enforces that templates be clearly named as such
> 
> I note that there exists some `.gotmpl` files that contain no templating, does it make sense to drop the extension there?

I would rather keep all as `.gotmpl` as to not create confusion about what should be where, unless we have things introduced into the template process that should act as some sort of source of truth that needs to be machine readable/writeable.

PR #2491, opened 2025-04-03
Aarnq/template tests

aarnq:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Work in progress to add template tests.

...

- Fixes #

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:


aarnq:
I got to stop here and ask if this is sane to do.

The idea is that we template each release and track changes, for the different configs variants we have.
It adds considerable amount of manifest that all have the potential for conflicts.

Though my plan is to add make commands to remove old resources and regenerate.
The issue is that there isn't the possibility to better target things, as the dynamic test registration only supports one tag per test.

Secondly I want to attempt to use common templates where possible so we can also test that they are consistent across clusters, installers, or providers, as needed.

simonklb (commented):
I can see _what_ this is testing but I think it's important to also have a _why_ the test should exist in the first place.

A lot of the tests here are testing upstream Helm charts. Do we not trust that upstream charts will produce the expected results? If we do this should/cloud be reduced to only making sure that our configs and Helm values are correct.

If the tests are becoming too big there is a risk of just copy-pasting diffs to fix them up rather than actually making use of the test results and catching bugs. This type of test might have been more appropriate if each component had it's own repository and tests.

Might be a dumb idea but what about only keeping the oldest release manifests and then incremental diffs? That would both reduce the amount of test resources as well as highlight what is actually changing between releases which I think could be very useful.

aarnq:
> I can see _what_ this is testing but I think it's important to also have a _why_ the test should exist in the first place.

Yeah, the goal here is to track the effect our changes have on the resulting templates, and as a bonus ensure that they are consistent in their output if the input is.

> A lot of the tests here are testing upstream Helm charts. Do we not trust that upstream charts will produce the expected results? If we do this should/cloud be reduced to only making sure that our configs and Helm values are correct.

These tests checks everything that is active, it doesn't separate local and upstream charts as it isn't a test of some input generates some output on a chart level. It is a full render of the pipeline from config through templates and charts with more templates to manifests.

> If the tests are becoming too big there is a risk of just copy-pasting diffs to fix them up rather than actually making use of the test results and catching bugs. This type of test might have been more appropriate if each component had it's own repository and tests.

I don't understand this risk of "just copy-pasting diffs", as this shouldn't include any copy pasting of diffs?
There would be support for updating the checked manifests, but that would still generate a diff, it would just be moved to git to be reviewed.
If someone mindlessly regenerated them, then those changes would be rather visible.

Like this is not intended to be a "this is how the manifest always should look like, forever and after".

And we don't have that split now, that would require that we split apps, like now.
Once we have modules in separate repos then it would, but then we would already need to have something to compare to.

> Might be a dumb idea but what about only keeping the oldest release manifests and then incremental diffs? That would both reduce the amount of test resources as well as highlight what is actually changing between releases which I think could be very useful.

I'm not sure how that would work, or what that would provide?
Like the intention here is to have a snapshot of how it looks right now.
If there is changes, then the tests will fail, which should tell the contributor to either accept or further work on it to not break something.
And let git take care of providing us with the changes done to them, so we can review them.

simonklb:
My bad I didn't get that this was the Module test framework. This makes a lot more sense now. I expected the snapshots for the Module tests to live in each Module's repository so I thought this was to snapshot test our current setup. Ignore my previous response. :smile: 

Just one initial thought then:

The amount of manifests that would have to be checked in for every configuration variant is going to explode exponentially. What about doing it at runtime during the tests and just have different supported configs as input and compare Apps output and Module output? Are you worried that both renders could be wrong and we wouldn't notice or why would you want to check everything in?

I envision a "Module conformance test" that iterates over every type of configuration variant that affects the Helm charts it produces and compares it to the manifests that is outputted by the current Apps setup.

aarnq:
> My bad I didn't get that this was the Module test framework. This makes a lot more sense now. I expected the snapshots for the Module tests to live in each Module's repository so I thought this was to snapshot test our current setup. Ignore my previous response. 😄

This is meant to test our current setup, and be the basis for the module test framework, and once we have modules the idea is for them to migrate to applicable modules.

As it is today we have a gap that it is difficult for a reviewer to see the impact of changes without having to verify it manually.
And for the modules we will have the same problem, potentially even worse as we will introduce a different template pipeline with Crossplane, as it will be a completely different beast.

So my vision is to provide this as something we can start working with today to get confidence into our current setup, that will translate into a security later on when we build modules.

And I don't want to build a framework for a one off conformance test for modules, as we will keep on doing upgrades and changes and what not, so the goal post will constantly move.

simonklb:
> > My bad I didn't get that this was the Module test framework. This makes a lot more sense now. I expected the snapshots for the Module tests to live in each Module's repository so I thought this was to snapshot test our current setup. Ignore my previous response. 😄
> 
> This is meant to test our current setup, and be the basis for the module test framework, and once we have modules the idea is for them to migrate to applicable modules.
> 
> As it is today we have a gap that it is difficult for a reviewer to see the impact of changes without having to verify it manually. And for the modules we will have the same problem, potentially even worse as we will introduce a different template pipeline with Crossplane, as it will be a completely different beast.
> 
> So my vision is to provide this as something we can start working with today to get confidence into our current setup, that will translate into a security later on when we build modules.
> 
> And I don't want to build a framework for a one off conformance test for modules, as we will keep on doing upgrades and changes and what not, so the goal post will constantly move.

Sounds good!

I'm still worried that it might give a false sense of security if it only covers the default configuration and not all the different variants. For example the rendering is different between `fluentd.audit = true` vs `fluentd.audit = false`. This will only test `fluentd.enabled = false`. And even if we added a specific test case for `fluentd.enabled = false` that won't cover `fluentd.audit.compaction.enabled = false` and so on.

Just to be clear, this is not to stop you from proceeding with this approach but I think we need to keep this in mind.

Zash:
Definitely promising work!

Seemed to have some trouble tho, or am I missing some prerequisite? 

```
ok 20 releases with custom resources have validation on install disabled - aws - workload cluster
bats: unknown test name `--description "template is consistent - aws capi ${cluster} ${release}" --tags resources -- template_test "${cluster}" "${release}"'
bats: unknown test name `--description "template is consistent - aws kubespray ${cluster} ${release}" --tags resources -- template_test "${cluster}" "${release}"'
ok 23 configuration is valid - aws:capi:air-gapped - service cluster
...
# bats warning: Executed 56 instead of expected 58 tests
make: *** [Makefile:83: container-unit/aws] Error 1
make: *** [Makefile:66: run-unit] Error 2
```

As for testing every combination of every value of every config option, some kind of fuzzing-like procedure would be an interesting **future** project. 😄

simonklb:
> As for testing every combination of every value of every config option, some kind of fuzzing-like procedure would be an interesting **future** project. 😄

Indeed. Or, which I would love even more, reduce the number of config options and possible values to a manageable number. :smile:

PR #2493, opened 2025-04-07
apps: require either letsencrypt or a seperate issuer

kristiangronas:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [x] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

There needs to be a clusterissuer, but some users may have their own issuer like sectigo or harica

#### Information to reviewers

You can test the fetching manually with `yajsv`

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -1354,9 +1354,6 @@ properties:
           Normally one of `letsencrypt-staging` or `letsencrypt-prod`.
         type: string
         default: letsencrypt-staging
-        enum:
-          - letsencrypt-staging
-          - letsencrypt-prod
       verifyTls:
         description: Verify ingress certificates
         type: boolean
@@ -6582,6 +6579,17 @@ properties:
     title: Issuers Config
     description: Configure issuers for cert-manager.
     type: object
+
+    # Need to either define your own issuers or enable letsencrypt
+    anyOf:
+      - properties:
+          letsencrypt:
+            properties:
+              enabled:
+                type: boolean
+                const: true
+      - required:
+          - extraIssuers
     additionalProperties: false
     properties:
       letsencrypt:
@@ -6634,6 +6642,18 @@ properties:
       extraIssuers:
         title: Extra Issuers
         type: array
+        minItems: 1
+        items:
+          $ref: "https://github.com/datreeio/CRDs-catalog/raw/refs/heads/main/cert-manager.io/clusterissuer_v1.json"
+          type: object
+          required:
+            - metadata
+          properties:
+            metadata:
+              type: object
+              required:
+                - name
+
   networkPolicies:
     title: Network Policies Config
     description: |-


config/schemas/config.yaml
```diff
@@ -6582,6 +6579,17 @@ properties:
...
+    # Need to either define your own issuers or enable letsencrypt
+    anyOf:
+      - properties:
+          letsencrypt:
+            properties:
+              enabled:
+                type: boolean
+                const: true
+      - required:
+          - extraIssuers
```
anders-elastisys, line 6592: We can enforce that minimum one extraIssuer is defined by adding the property in the `anyOf` block instead:
```suggestion
    anyOf:
      - properties:
          letsencrypt:
            properties:
              enabled:
                type: boolean
                const: true
      - properties:
          letsencrypt:
            properties:
              enabled:
                type: boolean
                const: false
          extraIssuers:
            type: array
            minItems: 1
        required:
          - extraIssuers
```

config/schemas/config.yaml
```diff
@@ -6634,6 +6642,18 @@ properties:
       extraIssuers:
         title: Extra Issuers
         type: array
+        minItems: 1
```
anders-elastisys, line 6645: With the other suggestion we can remove `minItems` from here (this would address the failed CI job):
```suggestion
```

anders-elastisys (commented):
Thanks for the contribution!
I am not an expert on configuring the schema, but I did some testing to ensure that it gets validated when letsencrypt is enabled, which is the default in Welkin. I added some comments with suggestions that seemed to achieve what you wanted.

kristiangronas:
Ah thanks, i think i assumed that extraIssuers would be unset completely instead of an empty array

config/schemas/config.yaml
```diff
@@ -6634,6 +6651,17 @@ properties:
       extraIssuers:
         title: Extra Issuers
         type: array
+        items:
```
anders-elastisys, line 6654: The unit tests complains about missing `title` and `description`, could you add this as well? :)

config/schemas/config.yaml
```diff
@@ -6634,6 +6651,17 @@ properties:
       extraIssuers:
```
anders-elastisys, line 6651: Seems to also be complaining about missing description for this

config/schemas/config.yaml
```diff
@@ -6634,6 +6651,17 @@ properties:
...
         title: Extra Issuers
         type: array
+        items:
+          $ref: "https://github.com/datreeio/CRDs-catalog/raw/refs/heads/main/cert-manager.io/clusterissuer_v1.json"
```
Zash, line 6656: I am a little bit worried about the security and performance of fetching these over the Internet, which is why we haven't done it like this before. Not sure how to proceed with that.

Zash: One way I looked at before was to pass referenced schemas via `yajsv -r path/to/schemas`, but at the time decided against because it made local references exceedingly verbose as they needed to be absolute URLs (matching the `$id` of the referenced schemas)

PR #2507, opened 2025-04-16
Purge yq3

rarescosma:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [x] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

> [!note]
> From this point onward the requirements installation step will install yq version 4 as simply `yq`.
>
> Most repositories have been updated to work with this change in mind, but if you've got any other workflows depending on `yq` being version 3 and that are running in the context of this repository's docker image or wrapper scripts, please inform someone from @elastisys/goto-scripts so we can update them too. This situation will be easy to notice if you're running through a wrapper script as it will simply abort with the notice: `expecting the yq binary to be at least version v4`.

Changes all `yq4` calls to use `yq` in an effort to break the dependency on the old version 3 `yq`. 

The main change here is that we'll install yq version 4 as both the `yq` and `yq4` binaries for a transition period.

- Part of elastisys/ck8s-issue-tracker#199

#### Information to reviewers

Unit tests should be passing. And there should not be any mention of `yq4` left in the repository, except for the requirements installation recipe.

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
    - [x] The [public documentation](https://github.com/elastisys/welkin) required an update - [link to Welkin PR](https://github.com/elastisys/welkin/pull/1099)
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:


rarescosma:
Guess the most interesting challenge here was switching from the yq3 merge command:

```console
yq m \
  ${CK8S_CONFIG_PATH}/defaults/common-config.yaml \
  ${CK8S_CONFIG_PATH}/defaults/sc-config.yaml \
  ${CK8S_CONFIG_PATH}/common-config.yaml \
  ${CK8S_CONFIG_PATH}/sc-config.yaml \
  -a overwrite -x > old_yq.yaml
```

to merging via `eval-all` using yq4:

```console
yq4 --no-colors eval-all '. as $item ireduce ({}; . * $item )' \
  ${CK8S_CONFIG_PATH}/defaults/common-config.yaml \
  ${CK8S_CONFIG_PATH}/defaults/sc-config.yaml \
  ${CK8S_CONFIG_PATH}/common-config.yaml \
  ${CK8S_CONFIG_PATH}/sc-config.yaml > new_yq.yaml
```

Diffing `old_yq.yaml` with `new_yq.yaml` produces an astonishing number of extra comments on the yq4 side.

But when we remove them: 

```console
yq4 -M '... comments=""' old_yq.yaml > old_yq_no_comments.yaml
yq4 -M '... comments=""' new_yq.yaml > new_yq_no_comments.yaml
```

and re-take the diff:

```console
$ diff new_yq_no_comments.yaml old_yq_no_comments.yaml

1231c1231
<       clusterIP: "10.96.0.20"
---
>       clusterIP: 10.96.0.20
```

helmfile.d/charts/harbor/harbor-backup/scripts/harbor-backup.sh
```diff
@@ -92,7 +92,7 @@ gcs_get_records() {
   # * Select only entries older than given date (epoch time)
   # * Save name of path to key "Key"
   gsutil -o "Credentials:gs_service_key_file=${GCS_KEYFILE}" ls -L "${PATH_TO_BACKUPS}" |
-    yq r - -j |
+    yq eval -j |
```
aarnq, line 95:
```suggestion
    yq eval -oj |
```

scripts/local-cluster.sh
```diff
@@ -55,14 +55,6 @@ log.continue() {
...
 }
 
-yq() {
-  if command -v yq4 >/dev/null; then
-    command yq4 "${@}"
-  else
-    command yq "${@}"
-  fi
-}
-
```
aarnq, line 65: Files using this should not really be touched as they only support yq v4.
The point with this function is to allow the script to use it as `yq` as we want to move away from installing it as `yq4` eventually.

rarescosma: Good to know, as I've just nuked that function from 4 other repositories 😅

rarescosma: @aarnq wait, does this mean that all the calls to `yq` that I've replaced with `yq4` should actually be just `yq`? For futureproofing?

helmfile.d/charts/harbor/harbor-backup/scripts/harbor-backup.sh
Zash: or even `--output-format json` for better readability? and maybe `--input-format yaml` instead of guessing?

scripts/local-cluster.sh
rarescosma: Here's an idea: why don't we change all of these blocks to define `yq4()` as a function instead of `yq` and then make sure we only have `yq4` invocations through-out the source code. 

That'll make it explicit that we expect the newer syntax, and also easier to track what has been ported and what not.

aarnq: I would rather do the opposite, so we actually start handling `yq` as `yq4`, and after this PR is done everything here will support that.

And regardless if the function turns into `yq4` or `yq` it _must_ prefer the `yq4` binary if it exists over `yq` for backwards compatibility until we have releases for everything that uses it.

roles/get-requirements.yaml
```diff
@@ -35,10 +34,6 @@
         version: "{{ sops_version }}"
         url: https://github.com/getsops/sops/releases/download/v{{ sops_version }}/sops-v{{ sops_version }}.linux.amd64
         dest: "{{ install_path }}/sops"
-      - command: yq
-        version: "{{ yq_version }}"
-        url: https://github.com/mikefarah/yq/releases/download/{{ yq_version }}/yq_linux_amd64
-        dest: "{{ install_path }}/yq"
```
Zash, line 41: So do we want to, in addition to removing this, also install `yq` version 4.x as `yq` and have some `state: absent` check on the old `yq4` binary? @aarnq

rarescosma:
The explosion radius of this PR is about to get huge. Pushing an update where all calls to `yq4` have been replaced to calls to `yq`, _except_ for migrations.

roles/get-requirements.yaml
rarescosma: @Zash First part is addressed here: https://github.com/elastisys/compliantkubernetes-apps/pull/2507/commits/cac93a578b30656dabc732359b747e6a5eb70075#diff-eb5d30c7072ceb41fb0540c8bea9f59f77eaa87b1e8c5156b2564bfdaca9ecc2L37-R40

Not sure how & where to add the `state: absent` check.

helmfile.d/charts/harbor/harbor-backup/scripts/harbor-backup.sh
rarescosma: Done!

scripts/local-cluster.sh
rarescosma: Ok, I updated this (and all the other PRs) to call `yq` wherever there were calls to `yq4` before. So as long as the base images install the binary under the correct name and/or everything is run through wrapper scripts that have the the `yq()` Bash dispatch function we should be fine.

roles/get-requirements.yaml
aarnq: Not sure, we have other currently supported releases that might rely on it exiting as `yq4` so I would rather not break that unnecessarily.

scripts/local-cluster.sh
aarnq: So, this wasn't really how I meant it, as my intention was to keep `yq()` instead of `yq4()`, and ensure that everything is called that way for a transition period until no supported version relies on yq v3.
Though this might be a better way forward, though then we must have an admin notice about this break.

The bin scripts lack the `yq()` dispatch, though it will check the version so it should be evident if the wrong one is used.

pipeline/README.md
```diff
@@ -6,6 +6,7 @@ If you want to run Bats tests locally (instead of in the CI pipeline), execute t
 git submodule update --recursive --init # init and update bats helpers
 
 docker build -t ck8s-apps ./pipeline
+# FIXME - there is no ./pipeline/sops-pgp-setup.bash anymore (?)
```
Zash, line 9: Good find. Looks like that script was removed in #1847 so maybe this file can be removed too? Running `bats` tests locally is covered by `tests/README.md` right? @elastisys/goto-qa 
Maybe a bit out of scope for this PR though?

roles/get-requirements.yaml
aarnq: I would say just leave it as is, and add it into the admin notice that it may be removed  unless they use it in other contexts.

scripts/local-cluster.sh
aarnq: Since not all scripts are covered by dispatchers, and since all documentation have been updated so we cannot keep v3 as `yq`, then I don't see the point of the dispatchers and they should be changed to a version check.

pipeline/README.md
aarnq: Yeah, this readme can be removed.

roles/get-requirements.yaml
rarescosma: Alright, we're now installing both `yq4` and `yq` binaries, but both at version 4.

scripts/local-cluster.sh
rarescosma: I think it's good to have them during the transitional period to avoid any impedance mismatches or corner cases in the release cycle. 

How about we remove them as part of the [follow-up task](https://github.com/elastisys/ck8s-issue-tracker/issues/485)?

pipeline/README.md
rarescosma: Removing.

Very out of scope, but just wandering if any code under `/pipeline` is still used *at all*?

scripts/local-cluster.sh
rarescosma: Ok, so basically change all dispatchers to just say the following?

```bash
    if ! command yq -V | grep --extended-regexp "v4\." >/dev/null 2>&1; then
      echo -e -n "[\e[31mck8s\e[0m] expecting the yq binary to be at least version v4" 1>&2
      exit 1
    else
      command yq "${@}"
    fi
```

aarnq: Except bin scripts, they have a version check already.

pipeline/README.md
aarnq: Yes, for `./bin/ck8s tests` only. The rest has been migrated into tests/.

scripts/local-cluster.sh
rarescosma: Cool, so plan is to remove the check for `yq4`. 
Should I go ahead and update the dispatchers in the other 8 pull requests as well? 😆

aarnq (approved):
Do update the platform administrator notice to reflect that we now install yq v4 as `yq`.

rarescosma:
> Do update the platform administrator notice to reflect that we now install yq v4 as `yq`.

Done, & thank you for the approval!

Zash (approved):
Awesome work!

PR #2510, opened 2025-04-24
apps wc: expose internal ingress-nginx LB config

### LINKED ISSUE #2504: [3] Expose option to enable additional internal load balancer for ingress-nginx
### Proposed feature

We want to enable the use of an [additional internal load balancer in ingress-nginx](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/README.md#additional-internal-load-balancer) so that ingress traffic can be split between internal and external traffic without requiring the use of additional ingress controllers or ingresses for the matter.

### Proposed alternatives

We want essentially to be able to expose a second `internal` service for the ingress controller, that allows for similar configuration as the current external one.

We additionally want to be able to setup this for multiple providers in the future so the solution should be generic, but we only require specific config for Azure to be provided with this issue.

### Additional context

_No response_

### Definition of done

- [x] Expose config option to enable the use of additional internal load balancers for ingress-nginx.
- [x] Generic way to configure to allow it to be used for multiple providers.
- [x] Specific defaults to enable it for Azure.

davidumea:
> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Exposes internal ingress-nginx load balancer config
It's possible to configure it for multiple providers

- Fixes #2504

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
    - [x] Any changed Pod is covered by Kubernetes Pod Security Standards
    - [x] Any changed Pod is covered by Gatekeeper Pod Security Policies
    - [x] The change does not cause any Pods to be blocked by Pod Security Standards or Policies
- NetworkPolicy checks:
    - [x] Any changed Pod is covered by Network Policies
    - [x] The change does not cause any dropped packets in the NetworkPolicy Dashboard
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/wc-config.yaml
+++ b/config/wc-config.yaml
@@ -320,12 +320,6 @@ ingressNginx:
         ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation
         allocateLoadBalancerNodePorts: true
 
-        ## Set external traffic policy to: "Local" to preserve source IP on
-        ## providers supporting it
-        ## Ref: https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-typeloadbalancer
-        ## Required for IP whitelisting
-        externalTrafficPolicy: Cluster
-
         ## Whitelist IP address for octavia loadbalancer
         ## ref: https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/expose-applications-using-loadbalancer-type-service.md#restrict-access-for-loadbalancer-service
         loadBalancerSourceRanges: []
--- a/config/wc-config.yaml
+++ b/config/wc-config.yaml
@@ -305,7 +305,7 @@ ingressNginx:
 
         ## Type of service.
         ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
-        type: set-me-if-(.ingressNginx.controller.service.enabled)
+        type: set-me-if-(.ingressNginx.controller.service.internal.enabled)
 
         ## Annotations to add to service
         ## ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
@@ -313,7 +313,7 @@ ingressNginx:
         ## ref: https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/README.md#additional-internal-load-balancer
         ## For Openstack use `service.beta.kubernetes.io/openstack-internal-load-balancer: "true"`
         ## For Azure use `loadbalancer.openstack.org/proxy-protocol: "true"`
-        annotations: set-me-if-(.ingressNginx.controller.service.enabled)
+        annotations: set-me-if-(.ingressNginx.controller.service.internal.enabled)
 
         ## Enable node port allocation
         ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation
--- a/config/wc-config.yaml
+++ b/config/wc-config.yaml
@@ -302,7 +302,6 @@ ingressNginx:
     ## Kubernetes service configuration.
     service:
       internal:
-        enabled: set-me
 
         ## Type of service.
         ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
--- a/config/wc-config.yaml
+++ b/config/wc-config.yaml
@@ -298,6 +298,58 @@ prometheusBlackboxExporter:
 
 ingressNginx:
   subDomain: ingress-nginx
+  controller:
+    ## Kubernetes service configuration.
+    service:
+      internal:
+        enabled: set-me
+
+        ## Type of service.
+        ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
+        type: set-me-if-(.ingressNginx.controller.service.enabled)
+
+        ## Annotations to add to service
+        ## ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+        ## Set the additional internal loadbalancer annotation if `type` is "LoadBalancer"
+        ## ref: https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/README.md#additional-internal-load-balancer
+        ## For Openstack use `service.beta.kubernetes.io/openstack-internal-load-balancer: "true"`
+        ## For Azure use `loadbalancer.openstack.org/proxy-protocol: "true"`
+        annotations: set-me-if-(.ingressNginx.controller.service.enabled)
+
+        ## Enable node port allocation
+        ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-nodeport-allocation
+        allocateLoadBalancerNodePorts: true
+
+        ## Set external traffic policy to: "Local" to preserve source IP on
+        ## providers supporting it
+        ## Ref: https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-typeloadbalancer
+        ## Required for IP whitelisting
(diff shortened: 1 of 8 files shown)


davidumea:
I will update the schema tomorrow :slightly_smiling_face:

helmfile.d/values/ingress-nginx.yaml.gotmpl
```diff
@@ -131,6 +131,34 @@ controller:
...
+    internal:
+      {{- if .Values | get "ingressNginx.controller.service.internal.type" "" | eq "LoadBalancer" }}
+      allocateLoadBalancerNodePorts: {{ .Values.ingressNginx.controller.service.internal.allocateLoadBalancerNodePorts }}
+      {{- end }}
+      {{ if .Values.ingressNginx.controller.service.internal.enabled }}
+      enabled: {{ .Values.ingressNginx.controller.service.internal.enabled }}
+      type: {{ .Values.ingressNginx.controller.service.internal.type }}
+      annotations: {{- toYaml .Values.ingressNginx.controller.service.internal.annotations | nindent 8 }}
+      loadBalancerSourceRanges: {{- toYaml .Values.ingressNginx.controller.service.internal.loadBalancerSourceRanges | nindent 8 }}
+      externalTrafficPolicy: {{ .Values.ingressNginx.controller.service.internal.externalTrafficPolicy }}
```
anders-elastisys, line 143: Would we ever configure this differently from `externalTrafficPolicy.local` that we set for other services?

davidumea: If you have proxy protocol enabled and externaltrafficpolicy set to local you can reach the internal endpoints from the public internet via something like this

`curl --insecure --connect-to <internal domain>:443:<external LB ip>:443 "https://<internal domain>"`

Actual example:
`curl --insecure --connect-to 172.16.1.236.nip.io:443:172.16.200.29:443 "https://172.16.1.236.nip.io"`

I thought I could work around this by setting externalTrafficPolicy to Cluster on the internal LB, but it doesn't seem to make a difference :smile: 

So I think we either have to disable proxy protocol when enabling this feature which would mean that the IP allowlisting annotation can not be used on public ingresses or be vulnerable to this. Any ideas @aarnq?

aarnq: > If you have proxy protocol enabled and externaltrafficpolicy set to local you can reach the internal endpoints from the public internet via something like this
> 
> `curl --insecure --connect-to <internal domain>:443:<external LB ip>:443 "https://<internal domain>"`
> 
> Actual example: `curl --insecure --connect-to 172.16.1.236.nip.io:443:172.16.200.29:443 "https://172.16.1.236.nip.io"`
> 
> I thought I could work around this by setting externalTrafficPolicy to Cluster on the internal LB, but it doesn't seem to make a difference 😄

Yeah, with PROXY protocol the source IP is part of the request, the only thing `externalTrafficPolicy: Local` does is ensure that incoming requests are only routed to Pods on the Node that received them.
With `externalTrafficPolicy: Cluster` it is allowed to go between Nodes.
So it is irrelevant for allowlisting in Ingresses.
(We only need it IIRC as our templates only allows it allowlisting in that case.) 

> So I think we either have to disable proxy protocol when enabling this feature which would mean that the IP allowlisting annotation can not be used on public ingresses or be vulnerable to this. Any ideas @aarnq?

This security concern has been brought up with product and they decided that it [is out of scope for this feature at this stage](https://github.com/elastisys/roadmaps/issues/219#issuecomment-2796380113).
Also, as this is considered for Azure only at this point we don't even use PROXY protocol.

The considered use case for this feature (again at this stage) is to have an independent load balancer for internal traffic that would be unaffected by DDoS attacks against the public load balancer.
Not to provide private Ingresses for internal clients.

Also, this shouldn't be enabled by default as it is an additional cost.
I admit that it is unclear from the issue, but the intention was that we had enough defaults for Azure to just be able to enable it with `enabled: true`.
And for the rest you didn't even have provide defaults so that is above and beyond. :rocket:

As a closing remark, wouldn't it be possible for internal ingresses to only allow clients from the private network? (Optionally the entire private network?)

davidumea: Thank you for the additional information!

>  I admit that it is unclear from the issue, but the intention was that we had enough defaults for Azure to just be able to enable it with enabled: true.

Okay, I will disable it in Azure by default.

I also removed the extra config option to set externalTrafficPolicy specifically for the internal LB and it will be determined by `externalTrafficPolicy.local` instead.

> As a closing remark, wouldn't it be possible for internal ingresses to only allow clients from the private network? (Optionally the entire private network?)

Yes you could in that case put an allowlist on the internal ingresses since it can still see the source IP. But it feels like that kind of defeats the purpose in the use case to provide private Ingresses for internal clients. However as you say that is not the goal here so I agree that we can disregard that for now.

config/wc-config.yaml
```diff
@@ -298,6 +298,51 @@ prometheusBlackboxExporter:
...
 ingressNginx:
   subDomain: ingress-nginx
+  controller:
+    ## Kubernetes service configuration.
+    service:
+      internal:
+
+        ## Type of service.
+        ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
+        type: set-me-if-(.ingressNginx.controller.service.enabled)
```
anders-elastisys, line 308: I assume we want this for the internal service
```suggestion
        type: set-me-if-(.ingressNginx.controller.service.internal.enabled)
```

config/wc-config.yaml
```diff
@@ -298,6 +298,51 @@ prometheusBlackboxExporter:
...
+
+        ## Annotations to add to service
+        ## ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+        ## Set the additional internal loadbalancer annotation if `type` is "LoadBalancer"
+        ## ref: https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/README.md#additional-internal-load-balancer
+        ## For Openstack use `service.beta.kubernetes.io/openstack-internal-load-balancer: "true"`
+        ## For Azure use `loadbalancer.openstack.org/proxy-protocol: "true"`
+        annotations: set-me-if-(.ingressNginx.controller.service.enabled)
```
anders-elastisys, line 316:
```suggestion
        annotations: set-me-if-(.ingressNginx.controller.service.internal.enabled)
```

helmfile.d/values/ingress-nginx.yaml.gotmpl
```diff
@@ -131,6 +131,40 @@ controller:
...
+      annotations: {{- toYaml .Values.ingressNginx.controller.service.internal.annotations | nindent 8 }}
+      loadBalancerSourceRanges: {{- toYaml .Values.ingressNginx.controller.service.internal.loadBalancerSourceRanges | nindent 8 }}
+      {{ if .Values.externalTrafficPolicy.local }}
+      ## Set external traffic policy to: "Local" to preserve source IP on
+      ## providers supporting it
+      ## Ref: https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-typeloadbalancer
+      ## Required for IP whitelisting
+      externalTrafficPolicy: "Local"
+      {{ else }}
+      enabled: false
```
anders-elastisys, line 150: This should not be here right?
```suggestion
```

anders-elastisys (approved):
Nice!

Eliastisys (approved):

Issue #1884, opened 2023-11-23
Make it possible to configure images


aarnq:
**Is your feature request related to a problem? Please describe.**

Compliant Kubernetes is static when it comes to the configuration of which images it uses, and we should change this to address some key scenarios:

- Using a private registry instead of a public registry
- Using a pull through cache
- Patching an image version to mitigate vulnerabilites

**Describe the solution you'd like**

Make it configurable for each instance of where we are using an image so we can:
- Configure the registry of the image (falling back to the one defined by the chart)
- Configure the repository of the image (falling back to the one defined by the chart)
- Configure the tag of the image (falling back to the one defined by the chart)

Make it configurable on a global level so we can:
- Configure the registry of all images (if they do not have an specific override).

This should also include images in resources created by operators to ensure that they are captured as well.

**Definition of done:**

- [ ] Images used can be individually configured.
- [ ] Image registry can be globally configured.

cristiklein:
We solved this by configuring the container runtime, as opposed to the `apps` layer: https://elastisys.io/compliantkubernetes/operator-manual/air-gapped-environment/#migrating-images-manually

Is this issue still relevant?

simonklb:
> We solved this by configuring the container runtime, as opposed to the `apps` layer: https://elastisys.io/compliantkubernetes/operator-manual/air-gapped-environment/#migrating-images-manually
> 
> Is this issue still relevant?

@aarnq feel free to correct me if I'm wrong. I think the idea here is for us to declaratively configure each image so that we are in more control of what is running in the cluster. It looks to me that the link you provided describes how to ingest images when you don't want to pull directly from the upstream registries.

aarnq:
> > We solved this by configuring the container runtime, as opposed to the `apps` layer: https://elastisys.io/compliantkubernetes/operator-manual/air-gapped-environment/#migrating-images-manually
> > Is this issue still relevant?
> 
> @aarnq feel free to correct me if I'm wrong. I think the idea here is for us to declaratively configure each image so that we are in more control of what is running in the cluster. It looks to me that the link you provided describes how to ingest images when you don't want to pull directly from the upstream registries.

Yes, so the goal of this issue is to go further and allow us to have an inventory of images that we can approve by pinning the checksum and sign. Allowing us to secure our supply chain by fetching the images we need, and better respond to vulnerabilities by using rebuilt or patched images dynamically, rather than statically.
[...]

PR #2521, opened 2025-05-06
Centralized container image configuration

rarescosma:
> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Implements a proof of concept for specifying container image overrides as part of the configuration/schema. We introduce the top-level `images` configuration key that has a `global` and `images` subkeys. 

With `global` a global registry and/or repository can be specified and activated. Within images container URIs can be specified for various components.

We limit the scope of the PR to a single affected component (the controller image in the ingress-nginx vendored chart), so we can gather early feedback before proceeding to systematically hunt down and wire in every image in our helmfile tree.

I know the approach of "convention over configuration" is questionable, but I choose to parse container URIs from a single string in the config/schema instead of having separate keys for `registry`, `repository`, `image` and so on just to keep things terse. The important bit is that we always go through an intermediary representation: 

```
{
  "registry": ...,
  "repository": ...,
  "image": ...,
  "tag": ...,
  "digest": ...,
}
```

..and then take the wiring decisions in the appropriate `.values.gotmpl` files to match impedance with `charts` and `upstream`.

- Part of #1884

#### Checklist

- Change checks:
    - [x] The change is disruptive
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:


aarnq:
I think it is good we limit this from the beginning so we can work out the details first :+1:

I don't really know where I stand on convention over configuration in this case, on one hand it is easier for an admin to manage regular URIs on the other hand another reason why we want this is to have a machine readable/writeable image list that we can verify, and map images with for mirroring/vendoring. Though it isn't _that_ hard to parse URIs for it as we know how they look.

If we think about a scenario in which we would need to bump a version to a patch version, this would require the admin to find the identifier, unless you intend that we could include fragments as well?
Would be strange but example: `image: :tag@checksum` or something similar to only indicate that we want to override parts of it. Though it could potentially become ambiguous.
We could also make a more involved merging/templating process, then we could allow both URIs and separate options. :thinking: 

I envision that we do not store the primary list in the common config, instead I would suggest an image list in the repository that is sourced during templating, and load it early to allow overrides from the common config.

Also see the comment I made in the issue.

rarescosma:
> Though it isn't that hard to parse URIs for it as we know how they look.

If it was doable in cursed gotmpl / sprig templating language, I'm sure we'll handle it in _any_ other language. 😆

rarescosma:
> ..instead I would suggest an image list in the repository that is sourced during templating..

This would require more `helmfile` foo than I currently posses. Do you think we can hop on a pair programming session to get this sorted? 

It's just a merge process, but I kinda like the idea of having it covered by jsonschema validation too.

aarnq:
> > ..instead I would suggest an image list in the repository that is sourced during templating..
> 
> This would require more `helmfile` foo than I currently posses. Do you think we can hop on a pair programming session to get this sorted?
> 
> It's just a merge process, but I kinda like the idea of having it covered by jsonschema validation too.

Yeah, we can do that. I'll send you an invite.

helmfile.d/lists/images.yaml
```diff
@@ -0,0 +1,6 @@
+---
+images:
+  ingressNginxChart:
+    controller: registry.k8s.io/ingress-nginx/controller-chroot:v1.12.1@sha256:90155c86548e0bb95b3abf1971cd687d8f5d43f340cfca0ad3484e2b8351096e
+    defaultBackend: registry.k8s.io/defaultbackend-amd64:1.5
+    admissionWebhooks: registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.5.2@sha256:e8825994b7a2c7497375a9b945f386506ca6a3eda80b89b74ef2db743f66a5ea
```
aarnq, line 6: Might be a better description, as it generates and patches in the certificate for the webhook which is in the controller.
```suggestion
    admissionWebhookPatch: registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.5.2@sha256:e8825994b7a2c7497375a9b945f386506ca6a3eda80b89b74ef2db743f66a5ea
```

helmfile.d/lists/images.yaml
```diff
@@ -0,0 +1,6 @@
+---
+images:
+  ingressNginxChart:
```
aarnq, line 3: I would probably not include chart in the key as I think it makes sense to group them based on the stack they are part of.
Else we will need to have `opensearchChart.node` and `opensearchDashboardsChart.dashboards` instead of `opensearch.{dashboards,node}`.
```suggestion
  ingressNginx:
```

helmfile.d/README.md
```diff
@@ -7,6 +7,7 @@ The structure is as follows:
 - `bases/` - contains the environment definition and some common values
 - `charts/` - contains internally developed charts grouped per stack
 - `upstream/` - contains externally developed charts grouped per source
+- `images/images.yaml` contains container image overrides
```
aarnq, line 10:
```suggestion
- `lists/images.yaml` contains container image overrides
```

aarnq (commented):
Some minor fixes, else I think it looks good.

helmfile.d/lists/images.yaml
rarescosma: I did this to differentiate between `upstream` and our own `charts`, but you're right, the `stack` level abstraction is better suited for it.

helmfile.d/lists/images.yaml
rarescosma: The same image is used in both the `job-createSecret` and `job-patchWebhook` files, and they refer to it in the chart as `.Values.controller.admissionWebhooks.patch.image`.

helmfile.d/README.md
rarescosma: Done 56c868cc

helmfile.d/lists/images.yaml
rarescosma: Done 8aef89f6

helmfile.d/lists/images.yaml
rarescosma: Done 421b0bcf

helmfile.d/values/ingress-nginx.yaml.gotmpl
```diff
@@ -155,9 +155,31 @@ controller:
...
+    {{- if $isChroot }}
+    digestChroot: {{ $controllerImage.digest }}
+    {{- else}}
+    digest: {{ $controllerImage.digest }}
+    {{- end}}
+{{- else}}
   image:
-    chroot: true
+    chroot: {{ $isChroot }}
+{{- end }}
```
aarnq, line 180: Should the image list contain the non-chroot image too?
Because effectively I don't think the non-chroot version will work now unless one manually specifies the image of the non-chroot version when chroot is disabled.

rarescosma: Done b58d7ee4

helmfile.d/lists/images.yaml
```diff
@@ -0,0 +1,7 @@
+---
+images:
+  ingressNginx:
+    controller: registry.k8s.io/ingress-nginx/controller-chroot:v1.12.1@sha256:d2fbc4ec70d8aa2050dd91a91506e998765e86c96f32cffb56c503c9c34eed5b
```
aarnq, line 4:
```suggestion
    controller: registry.k8s.io/ingress-nginx/controller:v1.12.1@sha256:d2fbc4ec70d8aa2050dd91a91506e998765e86c96f32cffb56c503c9c34eed5b
```

aarnq (approved):
One change then I'm happy.

helmfile.d/lists/images.yaml
rarescosma: Done 8c4edb33

Pavan-Gunda (approved):
Tested this locally and works like a charm. 

Great work :tada:

PR #2538, opened 2025-05-21
tests: speed up image tests by calling helmfile once per test case

rarescosma:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

It changes the structure of the `general/images` unit tests so that we render the `helmfile` templates once per tested _property_ instead of once per _test case_. This speeds up the images unit tests tremendously (from 300s to 9s on my machine), at the cost of less verbosity. 

The loss in verbosity is mitigated by logging which `container_image` and `template_file` we're looking at when extracting the image for assertions.

- Part of #1884 

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:


tests/unit/general/images.bats
```diff
@@ -3,6 +3,7 @@
 # bats file_tags=releases,general
 
 setup_file() {
+  bats_require_minimum_version 1.5.0
```
aarnq, line 6: What is the purpose of this when the function was added in 1.7.0 and we are on 1.11 something which is required for the parametric tests?

tests/unit/general/images.bats
```diff
@@ -38,199 +39,220 @@ _test_functions=(
...
 _export_params() {
-  export _image_property="${1}"
-  export _container_name="${2}"
-  export _template_file="${_templates_output}/${3}"
+  _image_properties="$(_get_test_prop image_property)"
+  _helmfile_selectors="$(_get_test_prop helmfile_selector)"
+  _container_names="$(_get_test_prop container_name)"
+  _template_files="$(_get_test_prop template_file)"
+  _num_tests=("$(echo "$_image_properties" | wc -w)" + 1)
+  export _image_properties _helmfile_selectors _container_names _template_files _num_tests
```
aarnq, line 249: I'd prefer to see the things needed to be arrays be managed as arrays, instead of intentionally word splitting and disabling SC2086.

tests/unit/general/resources/images-parametric-tests.json
```diff
@@ -2,16 +2,19 @@
   "parameters": [
     {
       "image_property": "ingressNginx.controllerChroot",
+      "helmfile_selector": "-lapp=ingress-nginx",
```
aarnq, line 5:
```suggestion
      "helmfile_selector": "app=ingress-nginx",
```
And add the flag during invocation.

Also if you want to be more specific you have the built in labels `name`, `namespace`, and `chart` to reduce the number of templates needed to be generated.

aarnq (commented):
Nice speed up!

tests/unit/general/images.bats
rarescosma: It's for calling the `run` built-in with the `--separate-stderr` argument. Without it I get spammed with warnings: 

```
BW02: Using flags on `run` requires at least BATS_VERSION=1.5.0. Use `bats_require_minimum_version 1.5.0` to fix this message.
      (from function `bats_warn_minimum_guaranteed_version' in file /home/karelian/.local/share/mise/installs/bats/1.11.1/bats-core-1.11.1/lib/bats-core/warnings.bash, line 32,
       from function `run' in file /home/karelian/.local/share/mise/installs/bats/1.11.1/bats-core-1.11.1/lib/bats-core/test_functions.bash, line 356,
       in test file tests/unit/general/images.bats, line 97)
```

tests/unit/general/images.bats
rarescosma: I know, my first implementation did use arrays, but you cannot easily export an array type in bash. Not even `mapfile -t` worked as expected. So ended up exporting strings, and then each test function had this kind of "beauty" in it: 

```
  local _container_names_a _template_files_a
  IFS=' ' read -ra _container_names_a <<< "$_container_names"
  IFS=' ' read -ra _template_files_a <<< "$_template_files"

  for ((i=0; i<${#_container_names_a[@]}; i++)); do
    ...
  done
```

Guess the real question is how do we read these parameters in arrays in each test without repeating ourselves?

tests/unit/general/images.bats
aarnq: Wonder why they would do it like that. Would the warning go away with 1.11.1, which would be more descriptive?

tests/unit/general/images.bats
rarescosma: EDIT: nevermind, the `setup` function actually allows globalizing arrays read through `read -r -a` 🥳

aarnq: You could also declare them in the test functions and read into them in the shared function, they will be available for any subsequent function from that context.

rarescosma: Done 3c79a32e

tests/unit/general/resources/images-parametric-tests.json
rarescosma: Done 3c79a32e

tests/unit/general/images.bats
rarescosma: I am on `1.11.1`, as seen in the "stack trace". Still generates those warnings. Should I add a comment before the invocation as to why it was added?

tests/unit/general/images.bats
rarescosma: Plus we got rid of `_export_params` in the process, nice!

tests/unit/general/images.bats
aarnq: Yes but did you _require_ 1.11.1?

rarescosma: Fixed + added motivation: 32136da9

tests/unit/general/resources/images-parametric-tests.json
aarnq: Recheck this, you have split the flag in two instead of the flag from the labels.

aarnq (approved):
LGTM

simonklb (approved):
33.3x speed up is nothing to scoff at. :smile:

PR #2541, opened 2025-05-21
Fix logic for disabled Opensearch snapshots

### LINKED ISSUE #2533: Disabled Opensearch snapshots breaks helmfile templating
### Experienced behaviour

When running `./bin/ck8s apply sc` and having `.opensearch.snapshots.enabled: false`, the command should run without issue.

### Expected behaviour

The command fails with the following message:

```
COMBINED OUTPUT:
  ********************
        Release was not present in Helm.  Diff will show entire contents as new.
  ********************
  Error: Failed to render chart: exit status 1: Error: template: opensearch-configurer/templates/config.yml:15:3: executing "opensearch-configurer/templates/config.yml" at <tpl (.Files.Get "files/configurer.sh") .>: error calling tpl: error during tpl function execution for..............................
template: gotpl:18:31: executing "gotpl" at <.Values.config.snapshots.repository>: nil pointer evaluating interface {}.repository
  Use --debug flag to render out invalid YAML
```

### Steps to reproduce

1. In `sc-config.yaml`, set this:

```yaml
opensearch:
  snapshot:
    enabled: false
```

2. Run: 

```bash
./bin/ck8s apply sc
```

3. The error can also be triggered without applying like so:

```bash
/bin/ck8s ops helmfile sc template -l app=opensearch
```

### Welkin Apps Version

v0.45.1

### Kubernetes Version

v1.31.7

### Additional context

This setting removes this part when set to false: https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile.d/values/opensearch/configurer.yaml.gotmpl#L37.

There are resources in the chart that are always being created and which depend on the `.opensearch.snapshots` object being available. https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile.d/charts/opensearch/configurer/files/configurer.sh is one example but there may be others as well.

elastisys-staffan:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

...

- Fixes #2533 

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:


elastisys-staffan:
> Looks good, and thanks for adding an extra test as well!
> 
> Does it work to have snapshotting disabled, install opensearch, and then turn on snapshotting? I assume that configurer and securityadmin will just rerun when you apply again, is that correct? For dev clusters I think this might happen, so that's why I want to make sure it works.

Good point, I just tried this it it actually seems to have broken some things. I'm going to investigate this before moving on.

rarescosma (approved):
Tested by cherry-picking onto a different work branch where I ran into the exact same issue and works like a charm, thank you! 🙏🏼

lunkan93:
> @viktor-f Update: The intial problems I ran into where an effect of an unrelated issue, once I sorted that one out I could do `.opensearch.snapshots.enabled: false` -> `.opensearch.snapshots.enabled: true` without issue. The resources where created as expected.
> 
> However, out of curiosity I also tried `.opensearch.snapshots.enabled: true` -> `.opensearch.snapshots.enabled: false`. The jobs will run to completion but you'll see a few errors in the UI in the snapshot configuration section. This likely happens because this maneuver removes the s3 configuration while the repository configuration from the `configurer.sh` script isn't cleaned up:
> 
> ```yaml
> # helmfile.d/values/opensearch/common.yaml.gotmpl
>     {{ if .Values.opensearch.snapshot.enabled -}}
>     # Object storage configuration
>     {{ if eq .Values.objectStorage.type "s3" -}}
>     s3:
>       client:
>         default:
>           endpoint: {{ .Values.objectStorage.s3.regionEndpoint }}
>           path_style_access: {{ .Values.objectStorage.s3.forcePathStyle }}
>           region: {{ .Values.objectStorage.s3.region }}
>     {{ else if eq .Values.objectStorage.type "gcs" -}}
>     # TODO: Add config for GCS if any.
>     {{ else if eq .Values.objectStorage.type "azure" -}}
>     azure:
>       client:
>         default:
>           account: {{ .Values.objectStorage.azure.storageAccountName }}
>     {{- end }}
>     {{- end }}
> ```
> 
> This only seems to happen in this particular scenario so I'm not sure it motivates the extra work in this context. What do you think? Should I try to fix it here or create another ticket for it?

If I understand this correctly, you mean that you get some errors in OpenSearch Dashboards when trying to access the Snapshot Management page after disabling snapshots? In that case I'd say that it's fine™

viktor-f:
> > @viktor-f Update: The intial problems I ran into where an effect of an unrelated issue, once I sorted that one out I could do `.opensearch.snapshots.enabled: false` -> `.opensearch.snapshots.enabled: true` without issue. The resources where created as expected.
> > However, out of curiosity I also tried `.opensearch.snapshots.enabled: true` -> `.opensearch.snapshots.enabled: false`. The jobs will run to completion but you'll see a few errors in the UI in the snapshot configuration section. This likely happens because this maneuver removes the s3 configuration while the repository configuration from the `configurer.sh` script isn't cleaned up:
>> ...
> > This only seems to happen in this particular scenario so I'm not sure it motivates the extra work in this context. What do you think? Should I try to fix it here or create another ticket for it?
> 
> If I understand this correctly, you mean that you get some errors in OpenSearch Dashboards when trying to access the Snapshot Management page after disabling snapshots? In that case I'd say that it's fine™

Thanks for testing enabled:false -> enabled:true.
Honestly I sort of think that it is ok if it is not possible to do the opposite, I.e. disabling snapshots after they had been enabled. If it can work then that is nice, but otherwise maybe a warning somewhere that it is not supported to disable it.

PR #2647, opened 2025-08-04
Reliable socat popups for localhost:8000 during end-to-end-test runs

rarescosma:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Reminding people to go open "localhost:8000" in their browser if/when a test suite gets stuck is not exactly ergonomic. Additionally, if we want to run proxy tests for the MC, it will require authentication with admin users through the central Dex and we can't even inform the user about it because we run cypress in silent more and `console.log` statements are dropped. (see [this discussion](https://github.com/elastisys/compliantkubernetes-apps/pull/2641#discussion_r2247937818) )

So I thought it would be nice if we had a reliable mechanism of auto-opening `localhost:8000` in the host browser when needed during *any* end-to-end test runs. The tool of choice is a statically linked `socat` binary that should work nicely regardless of the hosting OS (modulo glibc, of course). Inspiration taken from [this gist](https://gist.github.com/cristiklein/a7cc73e30a0d944e31e1b511a54e9091)

PR also removes the dbus sharing and all formerly installed xdg packages, and it further simplifies our `npm` setup so that `jsonschema2md` is now also managed through `package.json`. Getting rid of `npm` in further stages reduced the end-to-end docker image size by about 400Mb.

#### Information to reviewers

Run any suite that uses both MC & WC proxies and watch the popups pop:

```
make -C tests run-end-to-end/netpol/netpol.gen.bats
```

Additionally, you should _never_ have to authenticate towards Dex with any other users than your Google account.

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/scripts/run-from-container.sh
+++ b/scripts/run-from-container.sh
@@ -169,6 +169,15 @@ if [[ "${FORWARD_ENVIRONMENT:-false}" == "true" ]]; then
   args+=("--env" "CK8S_CONFIG_PATH")
   args+=("--mount" "type=bind,src=${CK8S_CONFIG_PATH},dst=${CK8S_CONFIG_PATH}")
   args+=("--env" "CK8S_HEADED_CYPRESS")
+
+  # Auto-open localhost:8000 for auth
+  socat_run="${root}/tests/end-to-end/.run"
+  socat_bin="${root}/tests/end-to-end/.bin"
+  if [[ -f "${socat_run}/socat.pid" ]] && kill -0 "$(cat "${socat_run}/socat.pid")" 2>/dev/null; then
+    args+=("-v" "${socat_run}/open-browser.sock:/run/user/$(id -u)/open-browser.sock")
+    args+=("-v" "${socat_bin}/socat-static:/usr/bin/socat:ro")
+    args+=("-v" "${socat_bin}/x-www-browser:/usr/bin/x-www-browser:ro")
+  fi
 fi
 
 # Check if we are in a work tree
--- a/scripts/run-from-container.sh
+++ b/scripts/run-from-container.sh
@@ -175,7 +175,6 @@ if [[ "${FORWARD_ENVIRONMENT:-false}" == "true" ]]; then
   socat_bin="${root}/tests/end-to-end/.bin"
   if [[ -f "${socat_run}/socat.pid" ]] && kill -0 "$(cat "${socat_run}/socat.pid")" 2>/dev/null; then
     args+=("-v" "${socat_run}/open-browser.sock:/run/user/$(id -u)/open-browser.sock")
-    args+=("-v" "${socat_bin}/socat-static:/usr/bin/socat:ro")
     args+=("-v" "${socat_bin}/x-www-browser:/usr/bin/x-www-browser:ro")
   fi
 fi
--- a/tests/common/exec.bash
+++ b/tests/common/exec.bash
@@ -157,7 +157,12 @@ postflight.integration() {
   return
 }
 postflight.end_to_end() {
-  return
+  local -r pid_file="${tests}/end-to-end/.run/socat.pid"
+
+  # we've got a pid file and there's a running socat with that pid => clean it up
+  if [[ -f "${pid_file}" ]] && kill -0 "$(cat "${pid_file}")" 2>/dev/null; then
+    kill "$(cat "${pid_file}")"
+  fi
 }
 
 # Runs the setup for a given test suite.
--- a/tests/common/exec.bash
+++ b/tests/common/exec.bash
@@ -142,7 +142,7 @@ end_to_end.socat_up() {
   if [[ ! -f "${pid_file}" ]] || ! kill -0 "$(cat "${pid_file}")" 2>/dev/null; then
     local -r sock_file="${tests}/end-to-end/.run/open-browser.sock"
     rm -f "${sock_file}"
-    socat unix-listen:"${sock_file}",fork system:'xargs xdg-open' &
+    socat -lf "${tests}/end-to-end/.run/socat.log" unix-listen:"${sock_file}",fork system:'xargs xdg-open' &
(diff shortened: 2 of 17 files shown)


rarescosma:
Ooops, forgot that some of our *bats* tests also expect `dev@example.com` authentication to have been performed, and so break the promise of this PR (that you only _ever_ have to use your google auth when a popup pops).

Will revert this to draft and attempt a fix.

tests/common/exec.bash
```diff
@@ -123,11 +123,36 @@ preflight.end_to_end() {
...
   fi
 }
+end_to_end.socat_up() {
+  local socat_sha="71c81cb46dd1903f12f3aef844b0fc559f31e2f613a8ae91ffb5630bc7011ef5"
+  local -r socat_path="${tests}/end-to-end/.bin/socat-static"
+  local -r pid_file="${tests}/end-to-end/.run/socat.pid"
+
+  if [[ ! -f "${socat_path}" ]]; then
+    mkdir -p "$(dirname "${socat_path}")"
+    curl -fsSL -o "${socat_path}" https://github.com/andrew-d/static-binaries/raw/0be803093b7d4b627b4d4eddd732e54ac4184b67/binaries/linux/x86_64/socat
```
Zash, line 140: Doesn't this run in a container? Any reason to not `apt install socat`? And how does it `xdg-open` from inside the container?

Or if it runs outside a container, why not `REQUIREMENTS` ?

rarescosma: > Doesn't this run in a container? Any reason to not `apt install socat`? And how does it `xdg-open` from inside the container?
> 
> Or if it runs outside a container, why not `REQUIREMENTS` ?

It runs on both host and container:
- For the host, I've added a commit to manage the `socat` binary through `REQUIREMENTS`: a4a9e5b
  The answer to 'why not `REQUIREMENTS` is simple: I don't develop on Ubuntu, so forgot about it 😆 
- Inside the container we mount a special `x-www-browser` executable that's called by `xdg-open` and uses socat to pass the URL through the socket back to the host

rarescosma:
> > Ooops, forgot that some of our _bats_ tests also expect `dev@example.com` authentication to have been performed, and so break the promise of this PR (that you only _ever_ have to use your google auth when a popup pops).
> > Will revert this to draft and attempt a fix.
> 
> Don't we have some tests that uses `admin@example.com` also?

Not outside of Cypress. 

I went through all of the `.bats` suites, and when they do request a kubeconfig that's not standard (`wc` or `sc`), they always go for the `dev@example.com` one.
 
The tests that are _in_ Cypress and use `admin@example.com` can handle authentication on their own.

rarescosma:
> So we will not automate the tests where we run as [admin@example.com](mailto:admin@example.com) different kubectl commands?

We can and we will. 

The goal of the epic is to get rid of most of that checklist. 

Plus we already have tests for delegation, but these authenticate as the static user: https://github.com/elastisys/compliantkubernetes-apps/blob/v0.48.0/tests/end-to-end/kubernetes/bin-access.bats

tests/common/exec.bash
Zash: I think my confusion came from no seeing where it launches the outside `socat`, looked like it jumps into the container fairly early. I am just now noticing the `tests/common/exec.bash preflight` calls.

scripts/run-from-container.sh
```diff
@@ -169,6 +169,14 @@ if [[ "${FORWARD_ENVIRONMENT:-false}" == "true" ]]; then
...
   args+=("--mount" "type=bind,src=${CK8S_CONFIG_PATH},dst=${CK8S_CONFIG_PATH}")
   args+=("--env" "CK8S_HEADED_CYPRESS")
+
+  # Auto-open localhost:8000 for auth
+  socat_run="${root}/tests/end-to-end/.run"
+  socat_bin="${root}/tests/end-to-end/.bin"
+  if [[ -f "${socat_run}/socat.pid" ]] && kill -0 "$(cat "${socat_run}/socat.pid")" 2>/dev/null; then
+    args+=("-v" "${socat_run}/open-browser.sock:/run/user/$(id -u)/open-browser.sock")
+    args+=("-v" "${socat_bin}/x-www-browser:/usr/bin/x-www-browser:ro")
+  fi
```
Zash, line 179: Is the `socat` process expected to stay running after the tests complete? If not I guess adding a `trap EXIT` around here may work.

Zash: Hm, yeah it gets complicated with parallelism. But parallelism inside a (single) container or multiple containers? Maybe easier if there's one `socat` process/pid file per invocation, but then I think you want to move the launching here too. Negligible cost compared to the rest of the test suite right? :)

I'd be fine with a notice somewhere and maybe a `make clean` that cleans up the `socat` and then we can figure out improving that later.

rarescosma: Let's worry about parallelism later 😆 

For now I've added "postflight" targets into the Makefile: https://github.com/elastisys/compliantkubernetes-apps/pull/2647/commits/169bc4583ab3bf49d5d9fc8fa1a64fd31be64db6

Along with some target changes to check the exit code after they ran: https://github.com/elastisys/compliantkubernetes-apps/pull/2647/commits/b900b7655be79b23d12e578aac6d1da4c0b59b25

Zash (approved):
Good work!

PR #2657, opened 2025-08-07
Add eslint for JS tests

rarescosma:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Adds `eslint` as the main linter for JS code in our tests codebase. Configures it with recommended styles from a couple of well-known suites, including the `eslint-cypress-plugin`.

Also extracts a `retryRequest` facility from the `falco` suite. Might be good for assertions over long chains of components interacting which might face substantial delay. (i.e. we deploy X, but then X gets polled by Y for metrics, Y aggregates them into alerts & sends the alerts to Z, we make assertions over Z's responses).

But most importantly: useless interpolation literals are now rejected by the linter, so this is no longer valid:

```
const foo=`just a normal string`
```

but this is valid:

```
const bar=`my backticks are fine because there's actually a ${templateVariable} in here`
```

#### Information to reviewers

Might be worthwhile going over this one commit by commit.

To test:

```shell
cd tests

npm install

eslint
```

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:


rarescosma:
> Good stuff 👍
> 
> Had to fight some local issues on my machine but eventually was able to test it out and it works well 😄 My npm/js ecosystem skills are clearly lacking..

What were the issues? Think we should be dockerizing the npm-requiring stuff in pre-commit sooner than later :)

lunkan93:
> > Good stuff 👍
> > Had to fight some local issues on my machine but eventually was able to test it out and it works well 😄 My npm/js ecosystem skills are clearly lacking..
> 
> What were the issues? Think we should be dockerizing the npm-requiring stuff in pre-commit sooner than later :)

First issue seemed to be that my npm/node installations were too old, then it looked like I had some old version of eslint installed globally that was getting used (probably installed while I was still on Ubuntu 22) which didn't work when running `eslint` manually in the terminal. Using the vscode extension was working out of the box.

I think I did `./node_modules/.bin/eslint --init` after removing the old binary to fix it? :sweat_smile:

tests/cypress.config.js
```diff
@@ -22,13 +23,11 @@ module.exports = defineConfig({
 
             child.unref()
 
-            resolve(null)
+            resolve(-1)
```
simonklb, line 26: Why not `resolve(true)`? `-1` makes it sound like a failure.

tests/cypress.config.js
```diff
@@ -43,15 +42,15 @@ module.exports = defineConfig({
               if (data.includes(PROXY_WAITING_FOR_DEX_MARKER)) {
                 resolve(data.toString().split(' ')[1])
```
simonklb, line 43: This could be something like:
```suggestion
                throw WaitingForDexAuthenticationException(data.toString().split(' ')[1])
```

tests/end-to-end/grafana/admin-promotion-ops.cy.js
```diff
@@ -32,7 +32,7 @@ describe('ops grafana user promotion', function () {
 
   afterEach(function () {
     cy.clearAllCookies()
-    cy.wrap(null).then(Cypress.session.clearAllSavedSessions)
+    cy.wrap(-1).then(Cypress.session.clearAllSavedSessions)
```
simonklb, line 35: Why is wrap needed here? Why not just call it like this:
```suggestion
    cy.then(Cypress.session.clearAllSavedSessions)
```

tests/common/check-js.bash
```diff
@@ -4,7 +4,7 @@
 #
 # This file is not meant to be run directly, but is called by a pre-commit hook.
 
-set -euo pipefail
+set -uo pipefail
```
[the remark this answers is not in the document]
rarescosma: Done cf23184

tests/cypress.config.js
simonklb: Just for the sake of it: in my book, while usually the case, an exception doesn't necessarily need to indicate an error or something failing. For example, `io.EOF` being returned as an error in Go or `KeyboardInterrupt` being thrown in Python. I think using exceptions in certain cases for signaling or control flow can actually be beautiful rather than trying to differentiate between different return values. I think retries could be one of those cases.

tests/end-to-end/grafana/admin-promotion-ops.cy.js
simonklb: But how would the `wrap` function impact any of that? I just looked up the [documentation for the wrap function](https://docs.cypress.io/api/commands/wrap) and couldn't understand how it would make any difference here, that's why I wondered if was needed or if it had any reason to be there previously.

PR #2666, opened 2025-08-12
apps: added group labels to alerts

### LINKED ISSUE #2634: [3] Group alerts in alertmanager
### Description

As a Welkin administrator, I am frustrated when I get multiple alerts from the same component for the same problem.
This is especially frustrating when I get multiple identical alerts but one for each pod, e.g. if there is a fluentd problem I get multiple of the same alert from every fluentd pod.

Use the alertmanager [grouping](https://prometheus.io/docs/alerting/latest/alertmanager/#grouping) feature to reduce spam.

It is part of this task to figure out which alerts should be grouped together. Ideally, alerts having the same underlying cause should be grouped together.

The alertmanager grouping default settings for intervals and delay can be used.

### Additional context

_No response_

### Definition of done

- [x] Welkin alerts are grouped to reduce alert spam.

davidumea:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [x] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

New alerts must now include a group label.

Added possibility and instructions on how to use grouped alerts
Added group label for all Welkin alerts

In order for grouping to work all alerts must have a group label, or all alerts with missing group label will be grouped together.
Grouping is disabled by default, keeping the previous behavior.

- Fixes #2634

#### Information to reviewers

Advise on grouping would be appreciated, a lot of alerts have the same group name as the alert name.

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/sc-config.yaml
+++ b/config/sc-config.yaml
@@ -1123,6 +1123,8 @@ alerts:
     #   {{ end }}
   opsGenie:
     apiUrl: https://api.eu.opsgenie.com
+    # Set updateAlerts to true if you intend to group alerts
+    updateAlerts: false
   # Configure custom alert receivers
   customReceivers: []
   customRoutes: []
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -4199,6 +4199,10 @@ properties:
             type: string
             format: uri
             default: https://api.eu.opsgenie.com
+          updateAlerts:
+            title: Whether to dynamically update existing alerts
+            type: boolean
+            default: false
       opsGenieHeartbeat:
         title: Alert OpsGenie Heartbeat
         description: Configure heartbeats to OpsGenie.
--- a/helmfile.d/charts/prometheus-alerts/README.md
+++ b/helmfile.d/charts/prometheus-alerts/README.md
@@ -1,3 +1,21 @@
-# Deviations from the upstream alerts
+# Prometheus-alerts
+
+All Welkin alerts must have a severity label to dictate which priority the alert should have.
+The different severity levels are:
+
+- critical
+- warning _or_ high
+- medium
+- low
+
+All Welkin alerts must also have a group label.
+If the alert does not fit into any specific group or should not be grouped together with other alerts, give the group label the same name as the alert.
+
+## Deviations from the upstream alerts
 
 1. In `fluentd.yaml` we set severity to warning and evaluation time to 10m for `FluentdRecordsCountsHigh`
+
+## OpsGenie
+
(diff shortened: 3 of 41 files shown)


Xartos (commented):
Question: For the AMSes we group much more fine grained, i.e. one group per alert almost. But here we do the opposite, group per application. Wouldn't it make sense to go the same route for both?
So either change to less grouping here or more grouping there

davidumea:
> > ... I fear that if we group more on AMS alerts, we might miss important stuff as one looks at the alert and sees something not urgent, and then next alert for something urgent comes and is added to existing alert for non urgent and is not handled in time
> 
> Could that not be the case for Apps as well?

(When using grouping) Alerts are grouped based on the severity label as well, so critical or "p1" alerts will never be grouped together with lower severity alerts.

Xartos:
> > > ... I fear that if we group more on AMS alerts, we might miss important stuff as one looks at the alert and sees something not urgent, and then next alert for something urgent comes and is added to existing alert for non urgent and is not handled in time
> > 
> > 
> > Could that not be the case for Apps as well?
> 
> (When using grouping) Alerts are grouped based on the severity label as well, so critical or "p1" alerts will never be grouped together with lower severity alerts.

Yea I was afraid of this as well. But I guess this is mainly something we need to adopt to, to check what alerts that has been grouped together. But I have a feeling that it can be easy to miss some stuff, especially in the beginning, when grouping so aggressively.

But I'm fine with doing this with apps to begin with and then maybe group more AMSes together later when we see that this works. As long as we easily can revert it if we see that this isn't working

davidumea:
> > > > ... I fear that if we group more on AMS alerts, we might miss important stuff as one looks at the alert and sees something not urgent, and then next alert for something urgent comes and is added to existing alert for non urgent and is not handled in time
> > > 
> > > 
> > > Could that not be the case for Apps as well?
> > 
> > 
> > (When using grouping) Alerts are grouped based on the severity label as well, so critical or "p1" alerts will never be grouped together with lower severity alerts.
> 
> Yea I was afraid of this as well. But I guess this is mainly something we need to adopt to, to check what alerts that has been grouped together. But I have a feeling that it can be easy to miss some stuff, especially in the beginning, when grouping so aggressively.
> 
> But I'm fine with doing this with apps to begin with and then maybe group more AMSes together later when we see that this works. As long as we easily can revert it if we see that this isn't working

It is easily reversable

PR #2672, opened 2025-08-14
scripts: added backup and restore script for QA

AlbinB97:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [x] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Adds a backup and restore script that can target harbor, velero, opensearch and rclone to automatically perform QA steps of taking and restoring backups.

- Fixes https://github.com/elastisys/ck8s-issue-tracker/issues/448

#### Information to reviewers

Tested all backups and restores on cluster running v0.48 with s3 object storage as both backup and restore target.

Have not tested the script with Azure or Swift.

Note that rclone requires user to manually apply sync or restore before running the script since I didn't want this script to interfere with peoples configs and apply changes to a cluster automatically. I am open to opinions on this though.

To run the script simply run `/backup-restore.sh <backup | restore> <opensearch | velero | rclone | harbor>`
It can be run from anywhere.

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/scripts/backup-restore.sh
+++ b/scripts/backup-restore.sh
@@ -1,0 +1,395 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ -z "$CK8S_CONFIG_PATH" ]; then
+  log "ERROR: CK8S_CONFIG_PATH is not set."
+  log "Please set and export the variable before running the script."
+  exit 1
+fi
+
+usage() {
+  echo "Usage: $0 <action> <target>"
+  echo "Actions: backup | restore"
+  echo "Targets: opensearch | harbor | rclone | velero"
+  exit 1
+}
+
+# Sets a timestamp log before every message, also colors it purple to make it easier to distinct from rest of the text because why not
+PURPLE='\033[0;35m'
+NOCOLOR='\033[0m'
+log() {
+  echo -e "[${PURPLE}$(date +'%Y-%m-%d %H:%M:%S')${NOCOLOR}]: $1"
+}
+
+# allows the script to be run from anywhere
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$(realpath "$SCRIPT_DIR/..")
+CK8S_CMD="$REPO_ROOT/bin/ck8s"
+
+if [ "$#" -ne 2 ]; then
+  usage
+fi
+
+ACTION="$1"
+TARGET="$2"
+
+run_backup() {
+  case "$1" in
+    opensearch)
+      log "Starting OpenSearch backup..."
+
+      local user="admin"
+      local password
+      password=$(sops -d "${CK8S_CONFIG_PATH}/secrets.yaml" | yq '.opensearch.adminPassword')
+      local os_url="https://opensearch.$(yq '.global.opsDomain' "${CK8S_CONFIG_PATH}/common-config.yaml")"
+      log "OpenSearch URL: ${os_url}"
+
+      log "Discovering snapshot repository name..."
+      local snapshot_repo
+      snapshot_repo=$(curl -kL -s -u "${user}:${password}" "${os_url}/_cat/repositories?v" | awk 'NR==2 {print $1}')
+      if [ -z "$snapshot_repo" ]; then
+        log "ERROR: Could not find OpenSearch snapshot repository."
+        exit 1
+      fi
+      log "Found snapshot repository: ${snapshot_repo}"
+
+      local snapshot_name="manual-snapshot-$(date +%Y%m%d%H%M%S)"
+      log "Taking snapshot: ${snapshot_name}"
+      curl -kL -u "${user}:${password}" -X PUT "${os_url}/_snapshot/${snapshot_repo}/${snapshot_name}" -H 'Content-Type: application/json' -d'
+      {
+        "indices": "*,-.opendistro_security",
+        "include_global_state": false
+      }
+      '
+
+      log "Waiting for snapshot to complete..."
+      local status=""
+      local retries=60 # 30 min timeout
+      while [ $retries -gt 0 ]; do
+        status=$(curl -kL -s -u "${user}:${password}" "${os_url}/_snapshot/${snapshot_repo}/${snapshot_name}" | jq -r '.snapshots[0].state')
+        log "Current snapshot status: ${status}"
+
+        if [ "$status" = "SUCCESS" ]; then
+          log "OpenSearch backup successfully completed."
+          return
+        elif [ "$status" = "FAILED" ] || [ "$status" = "PARTIAL" ]; then
+          log "ERROR: OpenSearch snapshot failed with status: ${status}"
+          exit 1
+        fi
+
+        sleep 30
+        retries=$((retries - 1))
+      done
+
+      log "ERROR: Timeout waiting for OpenSearch snapshot to complete."
+      exit 1
+      ;;
+    harbor)
+      log "Starting Harbor backup..."
+      local backup_job_name="harbor-backup-manual-$(date +%Y%m%d%H%M%S)"
+      log "Creating on-demand backup job: $backup_job_name"
+
+      "$CK8S_CMD" ops kubectl sc -n harbor create job "$backup_job_name" --from=cronjob/harbor-backup-cronjob
+
+      log "Waiting for backup job to complete..."
+      "$CK8S_CMD" ops kubectl sc -n harbor wait --for=condition=complete "job/$backup_job_name" --timeout=30m
+
+      log "Backup complete. Deleting job: $backup_job_name"
+      "$CK8S_CMD" ops kubectl sc -n harbor delete job "$backup_job_name"
+
+      log "Harbor backup successfully completed."
+      ;;
+    rclone)
+      log "Starting Rclone backup..."
+
+      # Check if rclone-sync is enabled before continuing
+      log "Checking for rclone-sync cronjobs"
+      local cronjob_list
+      cronjob_list=$("$CK8S_CMD" ops kubectl sc -n rclone get cronjobs -lapp.kubernetes.io/instance=rclone-sync -oname)
+
+      if [ -z "$cronjob_list" ]; then
+        log "ERROR: No rclone-sync cronjobs found."
+        log "Please ensure rclone sync is enabled in your configuration and has been applied."
+        exit 1
+      fi
+      local total_jobs
+      total_jobs=$(echo "$cronjob_list" | wc -w)
+      log "Found ${total_jobs} cronjobs."
+
+      # Create jobs from cronjobs
+      for cronjob in $cronjob_list; do
+        "$CK8S_CMD" ops kubectl sc -n rclone create job --from "${cronjob}" "${cronjob/#cronjob.batch\/}"
+        local job_name=${cronjob/#cronjob.batch\/}
+        log "Creating job '${job_name}' from '${cronjob}' cronjob..."
+      done
+
+      job_list=$("$CK8S_CMD" ops kubectl sc -n rclone get jobs -lapp.kubernetes.io/instance=rclone-sync -oname)
+
+      log "Waiting for all ${total_jobs} rclone jobs to complete..."
+      local retries=480 # these can take a really long time depending on content, so the timeout is set to 4 hours
+      while [ $retries -gt 0 ]; do
+        
+        local job_statuses
+        job_statuses=$("$CK8S_CMD" ops kubectl sc -n rclone get jobs -lapp.kubernetes.io/instance=rclone-sync -o json)
+        
+        local succeeded
+        succeeded=$(echo "$job_statuses" | jq '[.items[] | select(.status.succeeded == 1)] | length')
+        local failed
+        failed=$(echo "$job_statuses" | jq '[.items[] | select(.status.failed > 0)] | length')
+        local active
+        active=$(echo "$job_statuses" | jq '[.items[] | select(.status.active == 1)] | length')
+
+        log "Job Status -> Total: ${total_jobs} | Succeeded: ${succeeded} | Failed: ${failed} | Active: ${active}"
+
+        if [ "$active" -eq 0 ]; then
+          break
+        fi
+
+        sleep 30
+        retries=$((retries - 1))
+      done
+
+      # check if all jobs succeeded or not
+      if [ "$(echo "$job_statuses" | jq '[.items[] | select(.status.succeeded == 1)] | length')" -ne "$total_jobs" ]; then
+        log "ERROR: Not all rclone jobs succeeded. Please check the logs for failed jobs."
+        exit 1
+      fi
+
+      log "Cleaning up..."
+      for job in $job_list; do
+        "$CK8S_CMD" ops kubectl sc -n rclone delete "$job" --ignore-not-found=true
+      done
+
+      log "Rclone backup successfully completed."
+      ;;
+    velero)
+      log "Starting Velero backup..."
+
+      local backup_name="manual-backup-$(date +%Y%m%d%H%M%S)"
+      log "Creating Velero backup: ${backup_name}"
+
+      "$CK8S_CMD" ops velero wc backup create "$backup_name" --from-schedule velero-daily-backup
+
+      log "Waiting for backup to complete..."
+      local phase=""
+      local retries=90
+      while [ $retries -gt 0 ]; do
+        phase=$( "$CK8S_CMD" ops velero wc backup describe "$backup_name" -o json | jq -r .phase )
+        log "Current backup phase: ${phase}"
+
+        if [ "$phase" = "Completed" ]; then
+          log "Velero backup successfully completed."
+          return
+        elif [[ "$phase" =~ Failed|PartiallyFailed|Deleting ]]; then
+          log "ERROR: Velero backup failed or was deleted with phase: ${phase}"
+          exit 1
+        fi
+
+        sleep 30
+        retries=$((retries - 1))
+      done
+
+      log "ERROR: Timeout waiting for Velero backup to complete."
+      exit 1
+      ;;
+    *)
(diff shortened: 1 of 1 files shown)


scripts/backup-restore.sh
```diff
@@ -0,0 +1,401 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ -z "$CK8S_CONFIG_PATH" ]; then
+  log "ERROR: CK8S_CONFIG_PATH is not set."
+  log "Please set and export the variable before running the script."
+  exit 1
+fi
```
Xartos, line 9: NIT: I think we should try to use same as we typically have

```suggestion
: "${CK8S_CONFIG_PATH:?Missing CK8S_CONFIG_PATH}"
```

scripts/backup-restore.sh
```diff
@@ -0,0 +1,401 @@
...
+
+# Sets a timestamp log before every message, also colors it purple to make it easier to distinct from rest of the text because why not
+PURPLE='\033[0;35m'
+NOCOLOR='\033[0m'
+log() {
+  echo -e "[${PURPLE}$(date +'%Y-%m-%d %H:%M:%S')${NOCOLOR}]: $1"
+}
+
+# allows the script to be run from anywhere
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
```
Xartos, line 26: Question: Why change cwd? Why not just save the dir and use absolute paths?

```suggestion
SCRIPT_DIR="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
```

scripts/backup-restore.sh
```diff
@@ -0,0 +1,401 @@
...
+    local snapshot_repo
+    snapshot_repo=$(curl -kL -s -u "${user}:${password}" "${os_url}/_cat/repositories?v" | awk 'NR==2 {print $1}')
+    if [ -z "$snapshot_repo" ]; then
+      log "ERROR: Could not find OpenSearch snapshot repository."
+      exit 1
+    fi
+    log "Found snapshot repository: ${snapshot_repo}"
+
+    local snapshot_name
+    snapshot_name="manual-snapshot-$(date +%Y%m%d%H%M%S)"
```
Xartos, line 59: Suggestion: Do we want to enforce UTC to remove any timezone issues?

```suggestion
    snapshot_name="manual-snapshot-$(date --utc +%Y%m%d%H%M%S)Z"
```

Same further down

scripts/backup-restore.sh
```diff
@@ -0,0 +1,401 @@
...
+    while [ $retries -gt 0 ]; do
+      status=$(curl -kL -s -u "${user}:${password}" "${os_url}/_snapshot/${snapshot_repo}/${snapshot_name}" | jq -r '.snapshots[0].state')
+      log "Current snapshot status: ${status}"
+
+      if [ "$status" = "SUCCESS" ]; then
+        log "OpenSearch backup successfully completed."
+        return
+      elif [ "$status" = "FAILED" ] || [ "$status" = "PARTIAL" ]; then
+        log "ERROR: OpenSearch snapshot failed with status: ${status}"
+        exit 1
```
Xartos, line 80: Question: Do we want to catch other statuses? Or do we know that these are all

```suggestion
        exit 1
      else
        log "ERROR: Opensearch returned unknown status: ${status}"
        exit 1
```

scripts/backup-restore.sh
```diff
@@ -0,0 +1,401 @@
...
+  harbor)
+    log "Starting Harbor backup..."
+    local backup_job_name
+    backup_job_name="harbor-backup-manual-$(date +%Y%m%d%H%M%S)"
+    log "Creating on-demand backup job: $backup_job_name"
+
+    "$CK8S_CMD" ops kubectl sc -n harbor create job "$backup_job_name" --from=cronjob/harbor-backup-cronjob
+
+    log "Waiting for backup job to complete..."
+    "$CK8S_CMD" ops kubectl sc -n harbor wait --for=condition=complete "job/$backup_job_name" --timeout=30m
```
Xartos, line 99: Question: Will this not return error if timed out or if job failed? If so we should check status

```suggestion
    "$CK8S_CMD" ops kubectl sc -n harbor wait --for=condition=complete "job/$backup_job_name" --timeout=30m
    
    if [ $? -ne 0 ]; then
      log "ERROR: Harbor backup job failed"
      exit 1
    fi
```

scripts/backup-restore.sh
```diff
@@ -0,0 +1,401 @@
...
+    log "Finding latest successful snapshot..."
+    local latest_snapshot
+    latest_snapshot=$(curl -kL -s -u "${user}:${password}" -X GET "${os_url}/_snapshot/${snapshot_repo}/_all" | jq -r '.snapshots | map(select(.state == "SUCCESS")) | sort_by(.start_time_in_millis) | .[-1].snapshot')
+    if [ -z "$latest_snapshot" ] || [ "$latest_snapshot" = "null" ]; then
+      log "ERROR: No successful snapshots found to restore."
+      exit 1
+    fi
+    log "Found latest snapshot to restore: ${latest_snapshot}"
+
+    local indices_to_restore="kubernetes-*,kubeaudit-*,other-*,authlog-*"
```
Xartos, line 237: Suggestion: Maybe we should check if index per namespace is set and set something else in that case (maybe something like `[^\.].*` to not get indices starting with `.`)

scripts/backup-restore.sh
```diff
@@ -0,0 +1,401 @@
...
+    fi
+    local total_jobs
+    total_jobs=$(echo "$cronjob_list" | wc -w)
+    log "Found ${total_jobs} cronjobs."
+
+    # Create jobs from cronjobs
+    for cronjob in $cronjob_list; do
+      "$CK8S_CMD" ops kubectl sc -n rclone create job --from "${cronjob}" "${cronjob/#cronjob.batch\//}"
+      local job_name=${cronjob/#cronjob.batch\//}
+      log "Creating job '${job_name}' from '${cronjob}' cronjob..."
```
Xartos, line 297: Question: Would it make sense to have the log before running the command

```suggestion
      local job_name=${cronjob/#cronjob.batch\//}
      log "Creating job '${job_name}' from '${cronjob}' cronjob..."
      "$CK8S_CMD" ops kubectl sc -n rclone create job --from "${cronjob}" "${job_name}"
```

Alternatively change the wording to `Created job...`

scripts/backup-restore.sh
```diff
@@ -0,0 +1,401 @@
...
+
+if [ "$#" -ne 2 ]; then
+  usage
+fi
+
+ACTION="$1"
+TARGET="$2"
+
+run_backup() {
+  case "$1" in
```
Xartos, line 38: NIT: To make this a bit clear what argument this is for, maybe we should add

```suggestion
  local backup_target="${1}"
  case "${backup_target}" in
```

Mainly for the end of the function on the log message
```bash
  log "Error: Invalid backup target '$1'."
```

scripts/backup-restore.sh
AlbinB97: Only the subshell changes dir so it basically does the same thing no? but your version is way better since it actually works with symlinks as well I guess. Changing to that instead 👍

scripts/backup-restore.sh
AlbinB97: There's a set -e command at the top of the script that would exit the script if it failed. But you are 100% right I forgot to add a status check so it would just exit immediately without saying anything if failed.

Since there's a set -e, I believe your suggestion would fail as well since it would exit immediately and not do the if command? I added a `if !` check instead for the kubectl wait command, that should error out with a message if it fails I believe.

scripts/backup-restore.sh
AlbinB97: I made it easy and just changed the wording to created job. Since kubectl create job already spits out that it creates/created the job I am not sure if there's even a log necessary here. But I think it looks nice for the script itself to at least say that it created the job after the kubectl command finishes for logging/debugging purposes.

scripts/backup-restore.sh
```diff
@@ -0,0 +1,401 @@
...
+    local user="admin"
+    local password
+    password=$(sops -d "${CK8S_CONFIG_PATH}/secrets.yaml" | yq '.opensearch.adminPassword')
+    local os_url
+    os_url="https://opensearch.$(yq '.global.opsDomain' "${CK8S_CONFIG_PATH}/common-config.yaml")"
+    log "OpenSearch URL: ${os_url}"
+
+    log "Discovering snapshot repository name..."
+    local snapshot_repo
+    snapshot_repo=$(curl -kL -s -u "${user}:${password}" "${os_url}/_cat/repositories?v" | awk 'NR==2 {print $1}')
```
Zash, line 51: Nit: long flags improves readability

```suggestion
    snapshot_repo=$(curl --insecure --location --silent --user "${user}:${password}" "${os_url}/_cat/repositories?v" | awk 'NR==2 {print $1}')
```

Why is `-k,--insecure` needed?

Zash: These endpoints should normally have valid certs, right? I don't think `--insecure` is warranted here unless you know you're running with e.g. staging certs or so.

scripts/backup-restore.sh
AlbinB97: Added a check for this, using "-.*" instead, which we say to use in our documentation.

scripts/backup-restore.sh
```diff
@@ -0,0 +1,414 @@
...
+  echo "Actions: backup | restore"
+  echo "Targets: opensearch | harbor | rclone | velero"
+  exit 1
+}
+
+# Sets a timestamp log before every message, also colors it purple to make it easier to distinct from rest of the text because why not
+PURPLE='\033[0;35m'
+NOCOLOR='\033[0m'
+log() {
+  echo -e "[${PURPLE}$(date --utc +'%Y-%m-%d %H:%M:%S')${NOCOLOR}]: $1"
```
Xartos, line 18: Suggestion: Do we want to keep this local time? Since this will only be shown to the user running the script?

```suggestion
  echo -e "[${PURPLE}$(date +'%Y-%m-%d %H:%M:%S')${NOCOLOR}]: $1"
```

scripts/backup-restore.sh
```diff
@@ -0,0 +1,414 @@
...
+    local snapshot_repo
+    snapshot_repo=$(curl --insecure --location --silent --user "${user}:${password}" "${os_url}/_cat/repositories?v" | awk 'NR==2 {print $1}')
+    if [ -z "$snapshot_repo" ]; then
+      log "ERROR: Could not find OpenSearch snapshot repository."
+      exit 1
+    fi
+    log "Found snapshot repository: ${snapshot_repo}"
+
+    local snapshot_name
+    snapshot_name="manual-snapshot-$(date --utc +%Y%m%d%H%M%S)"
```
Xartos, line 56: Suggestion: To make it clear it's zulu time

```suggestion
    snapshot_name="manual-snapshot-$(date --utc +%Y%m%d%H%M%S)Z"
```

Same everywhere else we use UTC

scripts/backup-restore.sh
```diff
@@ -0,0 +1,414 @@
...
+      exit 1
+    fi
+    local total_jobs
+    total_jobs=$(echo "$cronjob_list" | wc -w)
+    log "Found ${total_jobs} cronjobs."
+
+    # Create jobs from cronjobs
+    for cronjob in $cronjob_list; do
+      "$CK8S_CMD" ops kubectl sc -n rclone create job --from "${cronjob}" "${cronjob/#cronjob.batch\//}"
+      local job_name=${cronjob/#cronjob.batch\//}
```
Xartos, line 126: Suggestion: Switch place on these and you don't need to do it twice

```suggestion
      local job_name=${cronjob/#cronjob.batch\//}
      "$CK8S_CMD" ops kubectl sc -n rclone create job --from "${cronjob}" "${job_name}"
```

Xartos (commented):
Just some minor suggestions otherwise I think it looks good

scripts/backup-restore.sh
Xartos: I'm pretty sure we had self-signed certificates at some point. But maybe we have valid certificates everywhere for these now?

AlbinB97: I just tested it without the --insecure flag on the sme-long-running, and it returns a 403 Forbidden. So it is definitely still needed 🤔

AlbinB97: Nevermind, I forgot to turn on VPN 🤦

lunkan93: Yeah, to my knowledge all certs should be valid these days

rarescosma (approved):
LGTM script-wise, great stuff!

Pre-approving but haven't tested 😄

aarnq (commented):
Overall look good!

scripts/backup-restore.sh
```diff
@@ -0,0 +1,415 @@
...
+    local -r snapshot_repo=$(curl --location --silent --user "${user}:${password}" "${os_url}/_cat/repositories?v" | awk 'NR==2 {print $1}')
+    if [ -z "$snapshot_repo" ]; then
+      log "ERROR: Could not find OpenSearch snapshot repository."
+      exit 1
+    fi
+    log "Found snapshot repository: ${snapshot_repo}"
+
+    local -r snapshot_name="manual-snapshot-$(date --utc +%Y%m%d%H%M%S)z"
+    log "Taking snapshot: ${snapshot_name}"
+    curl -kL -u "${user}:${password}" -X PUT "${os_url}/_snapshot/${snapshot_repo}/${snapshot_name}" -H 'Content-Type: application/json' -d'
```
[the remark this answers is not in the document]
AlbinB97: Ah, I forgot about those when removing the insecure flag from the other curl commands 🤦 will fix.

scripts/backup-restore.sh
```diff
@@ -0,0 +1,415 @@
...
+      indices_to_restore="-.*"
+    else
+      log "Index-per-namespace is disabled, using default indices..."
+      indices_to_restore="kubernetes-*,kubeaudit-*,other-*,authlog-*"
+    fi
+
+    log "Will restore indices matching pattern: ${indices_to_restore}"
+
+    log "Closing existing indices matching the restore pattern..."
+    curl --location --silent --user "${user}:${password}" -X POST "${os_url}/${indices_to_restore}/_close?pretty"
```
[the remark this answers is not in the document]
AlbinB97: Yeah, it works with wildcards etc. It closes the indices matching the pattern that we are planning to restore right after. OpenSearch will not let you restore a snapshot of an index if an index with the same name is already open and available on the cluster. I got errors before where it complained about indices still being open when trying to restore, so I added this to close them safely and then restore.

scripts/backup-restore.sh
```diff
@@ -0,0 +1,415 @@
...
+      {
+        "indices": "'"${indices_to_restore}"'"
+      }
+      '
+
+    log "Waiting for cluster health to become green..."
+    local status=""
+    local retries=60 # 30 min timeout
+    while [ $retries -gt 0 ]; do
+      status=$(curl --location --silent --user "${user}:${password}" "${os_url}/_cluster/health" | jq -r '.status')
```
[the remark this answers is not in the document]
AlbinB97: Not 100% sure, I think it's pretty reliable way to see if the opensearch is ready and healthy again after the restore no? It goes into yellow state for a while even though OpenSearch looks fine in Kubernetes, then when it's finally healthy again it becomes green and exits.

Do you have any other suggestion on the best way to track that the restore completed successfully and working again?

scripts/backup-restore.sh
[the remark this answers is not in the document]
aarnq: Right, with the way restores work now we expect the old write indices to take over essentially.
Maybe we should change that, but maybe I should bring that to the logging gotos, my idea is to not restore alias data and add a slug to the names of restored indices.
Then we wouldn't have to impact any current logs in the new environment, and we wouldn't need any special config for restored environments. (to skip setting up indices)

Marking this as resolved in the meantime, we can setup that in a separate one if needed.

scripts/backup-restore.sh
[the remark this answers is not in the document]
aarnq: I had a look at this and while you can check the status of recoveries, they are per shard, and not aggregated for a snapshot restore, so this is probably the simplest way to do it. Maybe you should check that the cluster is healthy before you begin though, else it might get forever stuck due to the initial state of it?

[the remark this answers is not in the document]
AlbinB97: I can do that 👍 Good call

scripts/backup-restore.sh
```diff
@@ -0,0 +1,415 @@
...
+
+    local -r backup_name="manual-backup-$(date --utc +%Y%m%d%H%M%S)z"
+    log "Creating Velero backup: ${backup_name}"
+
+    "$CK8S_CMD" ops velero wc backup create "$backup_name" --from-schedule velero-daily-backup
+
+    log "Waiting for backup to complete..."
+    local phase=""
+    local retries=90
+    while [ $retries -gt 0 ]; do
```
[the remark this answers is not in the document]
AlbinB97: I added a retries counter to the job status log. Were you thinking something different like writing out the timeout window as well before the loop starts?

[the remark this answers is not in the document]
aarnq: I think that looks good, less daunting than writing "this can take four hours", but it still conveys that it can take a while :sweat_smile:

Issue #2684, opened 2025-08-20
Add alert for OpenSearch max shards limit


lunkan93:
### Proposed feature

We currently have no alert for when the OpenSearch cluster is about to hit the maximum shards limit. We should add an alert for this, so that we get notified and can respond before the limit is hit.

### Proposed alternatives

There's a metric for total number of shards `elasticsearch_node_shards_total` which can be used in combination with this [config value](https://github.com/elastisys/compliantkubernetes-apps/blob/main/config/sc-config.yaml#L738) to determine what the alert threshold should be.

### Definition of done

- [ ] We have an alert for when OpenSearch is close to reaching the max shards limit

PR #2704, opened 2025-08-29
Adding OpenSearch alert for total shards limit

elastisys-staffan:
> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Adds an alert for when the combined total shards is greater than 90% of the `maxShardsPerNode` value.

- Fixes https://github.com/elastisys/compliantkubernetes-apps/issues/2684

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml
+++ b/helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml
@@ -17,6 +17,24 @@ spec:
   groups:
   - name: opensearch-alerts
     rules:
+    - alert: OpenSearchTotalShardsWarning
+      expr: sum(elasticsearch_node_shards_total{namespace="opensearch-system"}) > {{ mulf (mul .Values.osMaxShardsPerNode .Values.osDataNodeCount) 0.85 }}
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        description: Cluster total shards near hard limit
+        summary: The total number of shards across the nodes is {{`{{ $value }}`}} which is above 85% of the combined hard limit {{ mul .Values.osDataNodeCount .Values.osMaxShardsPerNode }}
+        runbook_url: {{ .Values.runbookUrls.opensearch.OpenSearchTotalShardsTooHigh }}
+    - alert: OpenSearchTotalShardsCritical
+      expr: sum(elasticsearch_node_shards_total{namespace="opensearch-system"}) > {{ mulf (mul .Values.osMaxShardsPerNode .Values.osDataNodeCount) 0.95 }}
+      for: 15m
+      labels:
+        severity: critical
+      annotations:
+        description: Cluster total shards near hard limit
+        summary: The total number of shards across the nodes is {{`{{ $value }}`}} which is above 95% of the combined hard limit {{ mul .Values.osDataNodeCount .Values.osMaxShardsPerNode }}
+        runbook_url: {{ .Values.runbookUrls.opensearch.OpenSearchTotalShardsTooHigh }}
     - alert: OpenSearchTooFewNodesRunning
       expr: elasticsearch_cluster_health_number_of_nodes{namespace="opensearch-system"} < {{ .Values.osNodeCount }}
       for: 15m
--- a/helmfile.d/values/prometheus-user-alerts-wc.yaml.gotmpl
+++ b/helmfile.d/values/prometheus-user-alerts-wc.yaml.gotmpl
@@ -8,6 +8,8 @@
 {{- $groupLabelValueArray := .Values.prometheus.autoscaledNodeGroupAlerts.groupLabelValues -}}
 {{- $groupLabelValueArrayRegex := join "|" $groupLabelValueArray -}}
 
+osMaxShardsPerNode: 0
+osDataNodeCount: 0
 osNodeCount: 0
 alertmanagerJob: alertmanager-operated
 alertmanagerNamespace: alertmanager
--- a/helmfile.d/charts/prometheus-alerts/CHANGELOG.md
+++ b/helmfile.d/charts/prometheus-alerts/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2025.08.29
+
+1. helmfile/charts/prometheus-alerts/templates/alerts/opensearch.yaml
+    - ADDED - OpenSearchTotalSHardsTooHigh alert
+
 ## 2025.01.03
 
 1. helmfile.d/charts/prometheus-alerts/templates/alerts/alertmanager.rules.yaml
--- a/helmfile.d/charts/prometheus-alerts/values.yaml
+++ b/helmfile.d/charts/prometheus-alerts/values.yaml
@@ -2,6 +2,8 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 osNodeCount: 3
+osDataNodeCount: 3
+osMaxShardsPerNode: 1000
 alertmanagerJob: kube-prometheus-stack-alertmanager
 alertmanagerNamespace: monitoring
 prometheusJob: kube-prometheus-stack-prometheus
--- a/helmfile.d/values/prometheus-alerts-runbook-urls.yaml.gotmpl
+++ b/helmfile.d/values/prometheus-alerts-runbook-urls.yaml.gotmpl
@@ -193,6 +193,7 @@ runbookUrls:
   nodeNetwork:
     {{- dict "name" "NodeNetworkInterfaceFlapping" "subpath" "general" "config" .nodeNetwork | include "tpl.prometheus-operator" }}
   opensearch:
+    {{- dict "name" "OpenSearchTotalShardsTooHigh" "config" .opensearch | include "tpl.missing-runbook" }}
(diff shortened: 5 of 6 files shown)


lunkan93:
Does changing the masterNode/dataNode count update the alert for you? I was trying to test things out but it did not, so I'm wondering if I messed something up on my end

Zash:
> Does changing the masterNode/dataNode count update the alert for you? I was trying to test things out but it did not, so I'm wondering if I messed something up on my end

@lunkan93 Can you describe how you tested that?

Zash:
> Updating either of [these values](https://github.com/elastisys/compliantkubernetes-apps/pull/2704/files#diff-6f1509887e27cc018de766e5f8028e9da9101cf6d7e1c7ff7edb846352bb5b91R15) in my env `sc-config.yaml` with no change for the alerts when diffing sc. What I was initially going to verify was that this gets added correctly when using dedicated data nodes vs all purpose masters.

Hm, I sorta expect `osNodeCount` to show up in `helmfile.d/state.yaml-*/sc-alerts.yaml` produced by `helmfile write-values`  but I don't, am I missing something?

lunkan93:
> > > Does changing the masterNode/dataNode count update the alert for you? I was trying to test things out but it did not, so I'm wondering if I messed something up on my end
> > 
> > 
> > Yeah, I can re-produce it. Changing those values does not update the alert.
> > Setting `maxShardsPerNode` does work as expected however, which is odd because I used `osNodeCount` as a reference when I did the wiring for that one. Will investigate.
> 
> This only took me two hours to track down, yay! 😍 It was a problem with the line above `osNodeCount`:
> 
> ```diff
> --- a/helmfile.d/values/prometheus-alerts-sc.yaml.gotmpl
> +++ b/helmfile.d/values/prometheus-alerts-sc.yaml.gotmpl
> @@ -10,7 +10,7 @@
>  {{- $groupLabel := printf "label_%s" $transformedGroupLabel -}}
>  
>  {{- $groupLabelValueArray := .Values.prometheus.autoscaledNodeGroupAlerts.groupLabelValues -}}
> -{{- $groupLabelValueArrayRegex := join "|" $groupLabelValueArray -}}
> +{{- $groupLabelValueArrayRegex := join "|" $groupLabelValueArray }}
>  
>  osNodeCount: {{ add .Values.opensearch.dataNode.count .Values.opensearch.clientNode.count .Values.opensearch.masterNode.count }}
>  osIndexAlerts: {{ toYaml .Values.opensearch.promIndexAlerts | nindent 2 }}
> ```
> 
> Fix here: [d45c1e4](https://github.com/elastisys/compliantkubernetes-apps/commit/d45c1e48c6e2276267eab02b8196a4e3f0759488)
> 
> I wonder if this ever worked as intended?
> 
> Anyway, I'll check if this fix messes with the labels somehow.

Sneaky :detective: Nice find! Changes are now being picked up. I confirmed that this nodeCount is not quite what we want for the purpose of these alerts, as the shards are only placed on data nodes. So what we would want is basically:
If dataNode.dedicatedPods: true then nodeCount = dataNode.count else nodeCount = masterNode.count

Does that make sense?

elastisys-staffan:
> Sneaky 🕵️ Nice find! Changes are now being picked up. I confirmed that this nodeCount is not quite what we want for the purpose of these alerts, as the shards are only placed on data nodes. So what we would want is basically: If dataNode.dedicatedPods: true then nodeCount = dataNode.count else nodeCount = masterNode.count
> 
> Does that make sense?

Some thing like this? https://github.com/elastisys/compliantkubernetes-apps/pull/2704/commits/0a7f4248a0203d2ad472a4d032183322c1d498f5

lunkan93:
> > Sneaky 🕵️ Nice find! Changes are now being picked up. I confirmed that this nodeCount is not quite what we want for the purpose of these alerts, as the shards are only placed on data nodes. So what we would want is basically: If dataNode.dedicatedPods: true then nodeCount = dataNode.count else nodeCount = masterNode.count
> > Does that make sense?
> 
> Some thing like this? [0a7f424](https://github.com/elastisys/compliantkubernetes-apps/commit/0a7f4248a0203d2ad472a4d032183322c1d498f5)

That looks good! We might want to use a separate nodeCount for these alerts, maybe something like dataNodeCount, as to not alter the behavior of [this alert](https://github.com/elastisys/compliantkubernetes-apps/pull/2704/files#diff-aec35538a623c22fc0038f077b02b77a6be68a053431454929d49f6092942511R29). But looking at it, it might be a good idea to separate that alert into one alert for each node type when using dedicated pods (client, data, master..). But that's out of scope for this task so I can create a new one for that :sweat_smile:

elastisys-staffan:
Tests are passing now! Approved? :smile:  @lunkan93 @Zash @vomba

lunkan93 (approved):
Nice :+1:

helmfile.d/values/prometheus-user-alerts-wc.yaml.gotmpl
```diff
@@ -8,6 +8,7 @@
 {{- $groupLabelValueArray := .Values.prometheus.autoscaledNodeGroupAlerts.groupLabelValues -}}
 {{- $groupLabelValueArrayRegex := join "|" $groupLabelValueArray -}}
 
+osMaxShardsPerNode: 0
 osNodeCount: 0
```
viktor-f, line 12: I think you might need to add `osDataNodeCount` here as well.

helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml
```diff
@@ -17,6 +17,15 @@ spec:
   groups:
   - name: opensearch-alerts
     rules:
+    - alert: OpenSearchTotalShardsTooHigh
+      expr: sum(elasticsearch_node_shards_total{namespace="opensearch-system"}) > {{ mulf (mul .Values.osMaxShardsPerNode .Values.osDataNodeCount) 0.9 }}
+      for: 15m
+      labels:
+        severity: critical
```
viktor-f, line 24: I'm wondering if we should have a "warning" severity version of this alert as well. Like a copy of this alert but change the severity and lower the threshold (maybe to 80%). The idea here is that we would hopefully get that alert during daytime a day or two before the number of shards are so high that we trigger the critical alert, because the critical alert might ping someone at night and this would hopefully give us time to fix the issue before that happens.

What do you think of that? I don't really know what the threshold should be for that, because I don't really know how fast the indices would increase. Maybe goto-logging can help with that.

viktor-f (commented):
I think the alert looks really good. Just have some questions before I approve.

helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml
lunkan93: That's a very good idea. Just looking at the numbers that these percentages result in with defaults, I'm thinking we could do 90% for the warning and 99% for the critical, which would be 2700/3000 and 2970/3000 with defaults. I don't think these should change super rapidly.

elastisys-staffan: How should we name them? Is `OpensearchTotalShardsWarning` vs `OpensearchTotalShardsCritical` fine?

viktor-f: Then perhaps do like 85% for warning and 95% for critical (99% felt like it was so high that if it ever reached that before we had a chance to act on the warning, then I fear that it would reach 100% before we could act on the critical alert. But I might be totally wrong about that, so feel free to take another number)?
Have we looked at some environments to see what percentage we are usually at? Preferably some envs with and some without index per namespace, since I assume that makes a significant difference.

I think those names are ok.

helmfile.d/values/prometheus-user-alerts-wc.yaml.gotmpl
elastisys-staffan: Fixed in latest commit

helmfile.d/charts/prometheus-alerts/templates/alerts/opensearch.yaml
```diff
@@ -17,6 +17,24 @@ spec:
...
   - name: opensearch-alerts
     rules:
+    - alert: OpenSearchTotalShardsWarning
+      expr: sum(elasticsearch_node_shards_total{namespace="opensearch-system"}) > {{ mulf (mul .Values.osMaxShardsPerNode .Values.osDataNodeCount) 0.85 }}
+      for: 15m
+      labels:
+        severity: warning
+      annotations:
+        description: Cluster total shards near hard limit
+        summary: The total number of shards across the nodes is {{`{{ $value }}`}} which is above 90% of the combined hard limit {{ mul .Values.osDataNodeCount .Values.osMaxShardsPerNode }}
```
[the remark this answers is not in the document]
elastisys-staffan: Fixed!

Issue #2311, opened 2024-10-15
Falco's least privileged mode for modern BPF driver does not work


aarnq:
### Experienced behaviour

Should be the mode enabled once modern BPF is selected.

However the templating sets the wrong value, and the podsecuritypolicies are not adapting to the least privileged mode.

Which I guess is lucky, as it works, just in full privileged mode.

### Expected behaviour

Least privileged mode should work as expected.

### Steps to reproduce

Enable `modern-bpf` driver in falco, see the pods apply with privileged.

### Compliant Kubernetes Apps Version

v0.40

Ajarmar:
@aarnq Can you add a bit more context and a definition of done to this issue?

- Why do we need `modern-bpf` driver/what problem does it solve?
- What needs to be fixed for this to work? Just templates?

aarnq:
@Ajarmar The `modern-bpf` driver solve the issue that Falco during runtime needs to download a kernel module driver, or fail if there are no pre built driver for the specific kernel version.

We already expose the config use it, and it currently works, but that is because of a bug.

But another benefit of the `modern-bpf` driver is that it has a "least privilege" mode, so the Falco container does not need `privileged: true` and we can constrain its security context by quite a lot compared to the regular `kmod` driver.

The bug is that the Falco templates tries to set least privilege mode, but on the wrong key so it isn't enabled, and the podsecuritypolicy isn't updated so it wouldn't allow the capabilities it would need, nor constrain the privileged it then doesn't need.

So currently it just set and accept `privileged: true`.

PR #2721, opened 2025-09-09
falco: switch over to the 'modern_ebpf' driver by default

rarescosma:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

The migration introduced in this PR will have two effects:
- environments that already use the `modern-bpf` driver will be updated to the correct `modern_ebpf` value
- environments using the `kmod` driver will be be updated to `modern_ebpf`

As the title suggests, it's high-time we switched to the `modern-bpf` driver as the default for Falco. Security GoTo task.

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/common-config.yaml
+++ b/config/common-config.yaml
@@ -182,7 +182,7 @@ falco:
   ## configure syscall source
   ## ref: https://falco.org/docs/concepts/event-sources/kernel/
   driver:
-    kind: kmod
+    kind: "modern-bpf"
 
     ebpf:
       # -- Path where the eBPF probe is located. It comes handy when the probe have been installed in the nodes using tools other than the init
--- a/config/common-config.yaml
+++ b/config/common-config.yaml
@@ -182,7 +182,7 @@ falco:
   ## configure syscall source
   ## ref: https://falco.org/docs/concepts/event-sources/kernel/
   driver:
-    kind: "modern_ebpf"
+    kind: modern_ebpf
 
     ebpf:
       # -- Path where the eBPF probe is located. It comes handy when the probe have been installed in the nodes using tools other than the init
--- a/config/common-config.yaml
+++ b/config/common-config.yaml
@@ -182,7 +182,7 @@ falco:
   ## configure syscall source
   ## ref: https://falco.org/docs/concepts/event-sources/kernel/
   driver:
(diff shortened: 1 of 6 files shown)


config/common-config.yaml
```diff
@@ -182,7 +182,7 @@ falco:
   ## configure syscall source
   ## ref: https://falco.org/docs/concepts/event-sources/kernel/
   driver:
-    kind: kmod
+    kind: "modern-bpf"
```
elastisys-staffan, line 185: The new name is `modern_ebpf`, do we want to use that instead? https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile.d/upstream/falcosecurity/falco/BREAKING-CHANGES.md#400 

If we do the schema needs to be updated as well.

elastisys-staffan (approved):
Overall I think it is a good idea to switch so I approve. I only have a single suggestion but it's not blocking.

aarnq:
Could you gotos then prioritise this issue so that it runs with the least set of privileges it needs? https://github.com/elastisys/compliantkubernetes-apps/issues/2311

config/common-config.yaml
aarnq: I would be fine with that, but maybe check with the rest of the security goto too.
What we also could do is to let that be an optional step, as we do not really remove the kmod driver support.
Also you might want to add a platform admin note in the PR desc.

aarnq: Just write it in the migration doc instead of a snippet.

rarescosma: Decided to force update from `kmod` to `modern_ebpf` and added a platform admin notice.

PR #2809, opened 2025-10-20
scripts: added image drift script

### LINKED ISSUE #2760: [3] Add script for checking image versions when upgrading upstream charts
### Description

We recently made all container image versions configurable, where the default versions are retrieved from the following list file: https://github.com/elastisys/compliantkubernetes-apps/blob/main/helmfile.d/lists/images.yaml
However, this makes it easy to miss to update the image versions in this list when upgrading upstream charts, as this is currently a manual process.
This especially tricky when upgrading bigger components like operators in which several images are updated (e.g. [kube-prometheus-stack](https://github.com/elastisys/compliantkubernetes-apps/pull/2759/files#diff-12f900cc2de446b951bd04f11129ad9f2f18700420594d9a509cacebd78de6a1)).
We should add some script that you can run that can check if the images in the `images.yaml` file should be updated when upgrading a chart.
The script needs to take our Welkin configuration in consideration when comparing, as some of the images used in Welkin may not be enabled by default in the upstream charts.
Another consideration are upstream images that we have moved to the Elastisys GitHub registry, as we would want to be able to catch version drifts here as well if possible.

### Additional context

_No response_

### Definition of done

- [ ] There is a way to check or update the `list/images.yaml` when upgrading upstream charts

shafi-elastisys:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

This PR introduces the `image_drift.sh` script, which helps detect version drift between upstream chart images and the pinned image tags defined in our application configuration `./helmfile.d/lists/images.yaml`.
Because we use custom naming for image keys in `images.yaml`, the script relies on a mapping file `image_mapping.sh` to connect each custom key to the corresponding upstream chart and image path.
After pulling the latest version of an upstream chart, the script can be executed to verify that our pinned image versions are still aligned with the upstream defaults. 
... 

- Fixes #2760 

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/helmfile.d/upstream/README.md
+++ b/helmfile.d/upstream/README.md
@@ -60,7 +60,7 @@ Pull the chart:
 ./scripts/charts.sh pull <chart-name>
 ```
 
-Check image drift and optionally update image tags if necessary (../lists/images.yaml):
+Check image drift and optionally update image tags if necessary ([../lists/images.yaml](../lists/images.yaml)):
 
 ```bash
 ./scripts/image_drift.sh <UPSTREAM_CHART_PATH>


Zash (commented):
Good work.

Some bash style nits.

shafi-elastisys:
I’ve updated the script to use the existing image mapping from the test resources (images-parametric-tests.json) for drift detection. Instead of using helmfile or helm show values, which could include overridden values, the script now compares the default values.yaml from the upstream chart directly with our images.yaml.

scripts/image_drift.sh
```diff
@@ -0,0 +1,161 @@
...
+LAST_PATH_COMPONENT=$(basename "$INPUT_PATH")
+CHART_FILTER=""
+UPSTREAM_BASE_DIR=""
+
+if [ -f "$INPUT_PATH/values.yaml" ]; then
+  # Single Chart Mode: Calculate UPSTREAM_BASE_DIR
+  CHECK_SCOPE="Single Chart: $LAST_PATH_COMPONENT"
+  CHART_FILTER="$LAST_PATH_COMPONENT"
+
+  for MAP_OBJECT in $(jq -c '.parameters[]' "$IMAGE_MAP_FILE"); do
```
Zash: Consider using `readarray` to guard against accidental word splitting in the middle of lines.

Zash (approved):
Seems to work!

Not sure if it should be blocking but some [style](https://google.github.io/styleguide/shellguide.html) nits, e.g.

> prefer `"${var}"` over `"$var"`.

shafi-elastisys:
Right but the script doesn't update the images.yaml. If user wish to update then it does the update to  file.

anders-elastisys (approved):
Tested it with the new changes works well, nice job!
I like the new change where we differentiate between newer and older images than upstream.

Issue #2733, opened 2025-09-17
Add dry-run support for migration scripts


Ajarmar:
### Description

We want to be able to dry-run individual migration scripts in order to see what changes a migration script would make. The purpose of this is so that we can know which apps are affected by an upgrade.

This only applies to [`apply` migration scripts](https://github.com/elastisys/compliantkubernetes-apps/blob/26a57057659ed1c75df0fb12934b26b4850baa1e/migration/template/apply/00-template.sh), not `prepare` migration scripts.

### Additional context

Preferably this would be implemented via changes and/or additions to the [migration helper functions](https://github.com/elastisys/compliantkubernetes-apps/tree/26a57057659ed1c75df0fb12934b26b4850baa1e/scripts/migration).

### Definition of done

- [ ] Migration helper functions exist that allow us to dry-run migration scripts.
- [ ] [Template](https://github.com/elastisys/compliantkubernetes-apps/blob/26a57057659ed1c75df0fb12934b26b4850baa1e/migration/template/apply/00-template.sh) is updated to match new changes/additions.

PR #2838, opened 2025-10-31
migration: Default CK8S_DRY_RUN_INSTALL to true

simonklb:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [x] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Without this the script fails when it's unset.

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/migration/template/prepare/00-template.sh
+++ b/migration/template/prepare/00-template.sh
@@ -6,6 +6,8 @@ ROOT="$(readlink -f "${HERE}/../../../")"
 # shellcheck source=scripts/migration/lib.sh
 source "${ROOT}/scripts/migration/lib.sh"
 
+: "${CK8S_DRY_RUN_INSTALL:=true}"
+
 # functions currently available in the library:
 #   - logging:
 #     - log_info(_no_newline) <message>
--- a/scripts/migration/helm.sh
+++ b/scripts/migration/helm.sh
@@ -13,7 +13,7 @@ helm_do() {
     log_fatal "Second argument must be a helm operation, not an option"
   fi
 
-  if [[ "${CK8S_DRY_RUN_INSTALL}" == "true" ]]; then
+  if [[ "${CK8S_DRY_RUN_INSTALL:-"true"}" == "true" ]]; then
     # List of common Helm subcommands that generally DO NOT support --dry-run
     case "${command}" in
     get | list | ls | history | diff | repo | search | show | status | test | version) ;;
--- a/scripts/migration/helmfile.sh
+++ b/scripts/migration/helmfile.sh
@@ -41,7 +41,7 @@ helmfile_apply() {
   prefix="${1}"
   shift
 
-  if [[ "${CK8S_DRY_RUN_INSTALL}" == "true" ]]; then
+  if [[ "${CK8S_DRY_RUN_INSTALL:-"true"}" == "true" ]]; then
     log_info "Dry-running on ${*}"
     helmfile_diff "${prefix}" "${@}"
     return
@@ -70,7 +70,7 @@ helmfile_destroy() {
   prefix="${1}"
   shift
 
-  if [[ "${CK8S_DRY_RUN_INSTALL}" == "true" ]]; then
+  if [[ "${CK8S_DRY_RUN_INSTALL:-"true"}" == "true" ]]; then
     log_info "Dry-run: Destroy on ${*} ${prefix}"
     return
   fi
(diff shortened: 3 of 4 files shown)


elastisys-staffan (commented):
Nice catch! How did you find it? It is set [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/upgrade.bash#L20) and [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration/template/apply/00-template.sh#L8) which I assumed would cover all cases. There are identical checks in several places, should we change them as well?

rarescosma (approved):

elastisys-staffan:
> > Nice catch! How did you find it?
> 
> While developing and running a migration snippet directly instead of via `ck8s upgrade`.
> 
> > It is set [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/upgrade.bash#L20) and [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration/template/apply/00-template.sh#L8) which I assumed would cover all cases.
> 
> Any reason why you didn't add it to the prepare template? It was a prepare snippet in my case.
> 
> I've added it here: [dd51e0b](https://github.com/elastisys/compliantkubernetes-apps/commit/dd51e0ba1427d2c6bddeaf6db2c9b4d0d3827932) let me know if there is a reason for it to not be there and I'll revert it!
> 
> > There are identical checks in several places, should we change them as well?
> 
> Indeed we should! [1552d39](https://github.com/elastisys/compliantkubernetes-apps/commit/1552d39756ae4db60787f8e669a20fc1d31e9452)

The [original task](https://github.com/elastisys/compliantkubernetes-apps/issues/2733) left out the prepare step so that's why it's not there. My understanding is that the helper functions that implements dry-run should only be used for apply jobs. Prepare jobs is for local config changes using the `yq` helpers.

I'm not against adding it here as well, but I would like to hear @Ajarmar 's take on it.

simonklb:
> > > Nice catch! How did you find it?
> > 
> > 
> > While developing and running a migration snippet directly instead of via `ck8s upgrade`.
> > > It is set [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/bin/upgrade.bash#L20) and [here](https://github.com/elastisys/compliantkubernetes-apps/blob/main/migration/template/apply/00-template.sh#L8) which I assumed would cover all cases.
> > 
> > 
> > Any reason why you didn't add it to the prepare template? It was a prepare snippet in my case.
> > I've added it here: [dd51e0b](https://github.com/elastisys/compliantkubernetes-apps/commit/dd51e0ba1427d2c6bddeaf6db2c9b4d0d3827932) let me know if there is a reason for it to not be there and I'll revert it!
> > > There are identical checks in several places, should we change them as well?
> > 
> > 
> > Indeed we should! [1552d39](https://github.com/elastisys/compliantkubernetes-apps/commit/1552d39756ae4db60787f8e669a20fc1d31e9452)
> 
> The [original task](https://github.com/elastisys/compliantkubernetes-apps/issues/2733) left out the prepare step so that's why it's not there. My understanding is that the helper functions that implements dry-run should only be used for apply jobs. Prepare jobs is for local config changes using the `yq` helpers.
> 
> I'm not against adding it here as well, but I would like to hear @Ajarmar 's take on it.

cc @aarnq as well

In this specific case, I've added a "pre-flight" / acceptance test as a prepare snippet which uses (read-only) Helm commands.

simonklb:
Here: https://github.com/elastisys/compliantkubernetes-apps/blob/d02faf70eadf711a1d50427facc8e03588a0c075/scripts/migration/crossplane.sh#L56

And in use here: https://github.com/elastisys/compliantkubernetes-apps/blob/5c3a466f6fa364d1b987ba4e0636a658a4414d10/migration/v0.50/prepare/90-module-node-local-dns-acceptance.sh

aarnq:
Should the default really be to mutate things? 

Also shouldn't get / read operations be performed even if we dry-run, so why do you need to run it with false from prepare snippets given the current state of it being only implemented for things normally done in apply snippets? I think limiting it would make it safer for use in that way in prepare snippets, right? :thinking:

simonklb:
@elastisys-staffan @Ajarmar any last words?

PR #2847, opened 2025-11-10
Disable CpuThrottlingHigh alert by default

anders-elastisys:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [x] kind/other         

> [!important]

- [x] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

The CpuThrottlingHigh alert has been disabled by default and is instead left as an opt-in alert.

This PR disables the CPUThrottlingHigh alert by default, as it is already ignored with OpsGenie integration, but not when integrating with other receivers like Slack or Teams.

- Part of https://github.com/elastisys/ck8s-issue-tracker/issues/548

#### Information to reviewers

Planning to discuss the disabled option further with GoTos as we might want to have the option to disable other alerts as well. The config is pretty much copy pasted from the upstream prometheus chart in case we ever decide to move to using the upstream chart for some alerts instead of our own.

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/config/common-config.yaml
+++ b/config/common-config.yaml
@@ -218,8 +218,8 @@ alerts:
     thanos: {}
     webhook: {}
 
-  disabled:
-    CPUThrottlingHigh: true
+  enabled:
+    CPUThrottlingHigh: false
 
 ## Falco configuration.
 falco:
--- a/config/schemas/config.yaml
+++ b/config/schemas/config.yaml
@@ -4427,16 +4427,16 @@ properties:
                 type: string
           webhook:
             $ref: '#/properties/alerts/properties/runbookUrls/$defs/noUpstreamRunbook'
-      disabled:
+      enabled:
         type: object
-        title: Disabled alerts
-        description: Alerts that can be disabled
+        title: Enabled alerts
+        description: Enabled alerts
         properties:
           CPUThrottlingHigh:
-            title: CPUThrottlingHigh disabled
-            description: Configure whether or not CPUThrottlingHigh alert should be disabled
+            title: CPUThrottlingHigh enabled
+            description: Configure whether or not CPUThrottlingHigh alert should be enabled
             type: boolean
-            default: true
+            default: false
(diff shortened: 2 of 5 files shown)


config/common-config.yaml
```diff
@@ -218,6 +218,9 @@ alerts:
     thanos: {}
     webhook: {}
 
+  disabled:
```
Zash: Thoughts on `disabled: true` vs `enabled: false # true if omitted` ?

We have a lot more `enabled` than `disabled` in the config so far, so perhaps more consistent to go with `enabled`, which also feels like less of a double negative to me.

anders-elastisys: I am also not a fan of double negative, but I wanted it to the upstream, however, our helmfile config does not need to match this anyways. If people have strong opinions on this I can change it

Zash (approved):

config/common-config.yaml
anders-elastisys: I changed it to enabled now as you say we do this for most of our config, but good to note that this deviates from how it is upstream so if we ever switch to use upstream we need to address that either in our config, or, in our helmfile values. https://github.com/elastisys/compliantkubernetes-apps/pull/2847/commits/10ef91b814c19eecdc27b39b70f426a9fd7d7505

viktor-f: Thanks, I think it is fine to switch in the helmfile values if we ever switch to use the upstream chart.

PR #2859, opened 2025-11-13
Simplify log-manager e2e suite

rarescosma:
> [!warning]
>

### What kind of PR is this?

- [ ] kind/feature       
- [x] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

Attempts to remove the `setup_suite` [foxtrot](https://github.com/elastisys/compliantkubernetes-apps/blob/v0.49.0/tests/end-to-end/log-manager/setup_suite.bash) that was necessary to stop fluentd from uploading new blocks into object storage, itself a prerequisite of log-manager tests.

Based on an internal discussion between @aarnq and @simonklb that was summarized in slack:

> What we were talking about was to not test it exhaustively in the end to end, and make an exhaustive test for integration testing instead.
>
> So for end to end, instead of checking that all blocks are compacted, we check that all blocks before the start of the job are compacted. And similarly for retention.
>
> If there is nothing for them to do then that should still be considered an OK run of the tests, as that is expected. And we'd remove the test of it not needing to do anything. Then we don't need to block fluentd from uploading new blocks.
>
> For integration tests we create and seed separate prefixes in object storage, and test the different scenarios for both compaction and retention, as we then can generate the data as needed. And here we'd cover both regular runs, and runs without need for compaction or retention.

This PR covers the first part (for the end-to-end suite).

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/tests/end-to-end/log-manager/compaction-audit-sc.gen.bats
+++ b/tests/end-to-end/log-manager/compaction-audit-sc.gen.bats
@@ -90,14 +90,6 @@ setup() {
   object_storage.is_compacted "${BUCKET}" "${PREFIX}" "${cutoff}"
 }
 
-@test "audit-sc - log-manager compaction job runs fine when no logs need compaction" {
-  local -r cutoff="$(_get_latest_job_start_time)"
-
-  object_storage.is_compacted "${BUCKET}" "${PREFIX}" "${cutoff}"
-
-  test_run_cronjob "${CRONJOB_NAME}" 120
-}
-
 _create_uncompacted_objects() {
   local test_data_file_1
   local test_data_file_2
--- a/tests/end-to-end/log-manager/compaction-audit-wc.gen.bats
+++ b/tests/end-to-end/log-manager/compaction-audit-wc.gen.bats
@@ -90,14 +90,6 @@ setup() {
   object_storage.is_compacted "${BUCKET}" "${PREFIX}" "${cutoff}"
 }
 
-@test "audit-wc - log-manager compaction job runs fine when no logs need compaction" {
-  local -r cutoff="$(_get_latest_job_start_time)"
-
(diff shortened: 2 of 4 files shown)


tests/end-to-end/log-manager/compaction-audit-sc.gen.bats
```diff
@@ -85,11 +85,15 @@ setup() {
...
 
-@test "audit-sc - log-manager compaction job runs fine when no logs needs compaction" {
-  object_storage.is_compacted "${BUCKET}" "${PREFIX}"
+@test "audit-sc - log-manager compaction job runs fine when no logs need compaction" {
+  local -r cutoff="$(_get_latest_job_start_time)"
+
+  object_storage.is_compacted "${BUCKET}" "${PREFIX}" "${cutoff}"
 
   test_run_cronjob "${CRONJOB_NAME}" 120
 }
```
simonklb: I think what caused the setup initially was this test. How do we know that there is "no logs that need compaction" when it runs? I think this is no longer can't be guaranteed.

It was part of the initial issue description but maybe I should have questioned it rather than trying to fit it in the e2e suite.

I would argue that this is test too low level and should be moved to the integration test instead.

aarnq (approved):
LGTM

PR #2893, opened 2025-12-09
scripts: added temp gpg key script

### LINKED ISSUE #2766: Update documentation on GPG and create a temp solution for quick start
### Proposed feature

Apps has a hard requirement on being able to encrypt secrets with GPG, but to simplify the setup for new Platform Administrators we should:
- first have better documentation to support the creation of GPG keys, and setup of SOPS, and 
- second have a fallback to support setting up a temporary GPG home and key for local testing

### Proposed alternatives

The first should be done on the public docs, as then we can link to it from any repo.

The second should be done here in Apps, as the prime use-case will be for local-clusters and Apps combo.

### Additional context

_No response_

### Definition of done

- [ ] Public docs for setting up GPG and SOPS
- [ ] Scripts and helpers to run a temporary GPG home for local testing

shafi-elastisys:
> [!warning]
>

### What kind of PR is this?

- [x] kind/feature       
- [ ] kind/improvement   
- [ ] kind/deprecation   
- [ ] kind/documentation 
- [ ] kind/clean-up      
- [ ] kind/bug           
- [ ] kind/other         

> [!important]

- [ ] kind/admin-change   
- [ ] kind/dev-change     
- [ ] kind/security       
- [ ] [kind/adr](set-me\) 

This PR introduces script to which helps to setup a temporary GPG for local testing. 

* Added `bin/setup-local-gpg` which generaes a GPG Key non-interactively and exports necessary environment variables for SOPS to use. 
* Updated `bats.lib.bash` added helper functions. This allows test to invoke the script during the setup phase. 

- Fixes #2766 

#### Checklist

- Change checks:
    - [ ] The change updates CRDs
- Documentation checks:
- Metrics checks:
- Logs checks:
- PodSecurityPolicy checks:
- NetworkPolicy checks:
- Audit checks:
- Falco checks:
- Bug checks:

### DIFF
--- a/scripts/local-cluster.sh
+++ b/scripts/local-cluster.sh
@@ -310,7 +310,7 @@ config() {
       # shellcheck disable=SC1090
       source "${env_file}"
       log.info "CK8S_PGP_FP was unset. Reusing temporary GPG environment from ${env_file}"
-      log.info "To persist this in your current terminal, run:"
+      log.info "To load this in your current terminal, run:"
       log.info "  source ${env_file}"
     elif [[ -x "${ROOT}/bin/setup-local-gpg.bash" ]]; then
       source "${ROOT}/bin/setup-local-gpg.bash"


tests/bats.lib.bash
```diff
@@ -162,6 +162,22 @@ with_kubeconfig() {
...
+
+  eval "$(setup-local-gpg)"
+
+  if [[ -z "${CK8S_PGP_FP:-}" ]]; then
+    fail "Failed to generate temporary GPG key"
+  fi
+
+  log.trace "Temporary GPG initialized. Fingerprint: $CK8S_PGP_FP"
+}
+
```
aarnq: Is this needed? We already have a function for it already in `common/bats/gpg.sh` that also takes care of an edge case in which the GPG agent isn't available in the first tries.

scripts/local-cluster.sh
```diff
@@ -304,8 +304,27 @@ config() {
...
 
+  local env_file="${CK8S_CONFIG_PATH}/.temp_gpg_env"
   if [[ -z "${CK8S_PGP_FP:-}" ]]; then
-    log.fatal "CK8S_PGP_FP is unset"
+    if [[ -f "${env_file}" ]]; then
+      # shellcheck disable=SC1090
+      source "${env_file}"
+      log.info "CK8S_PGP_FP was unset. Reusing temporary GPG environment from ${env_file}"
+      log.info "To persist this in your current terminal, run:"
+      log.info "  source ${env_file}"
```
aarnq: Since it won't be too persistent if the temp gnupg home dir is cleared maybe we should make it seem less permanent 😄 same on the other print.
```suggestion
      log.info "To load this in your current terminal, run:"
      log.info "  source ${env_file}"
```

scripts/local-cluster.sh
```diff
@@ -304,8 +304,27 @@ config() {
     log.fatal "CK8S_CONFIG_PATH is unset"
   fi
 
+  local env_file="${CK8S_CONFIG_PATH}/.temp_gpg_env"
```
aarnq: We have a preference to use kebab case for files
```suggestion
  local env_file="${CK8S_CONFIG_PATH}/.temp-gpg-env"
```

scripts/local-cluster.sh
```diff
@@ -353,6 +372,7 @@ config() {
   fi
 
   "${ROOT}/bin/ck8s" init both
+
```
aarnq:
```suggestion
"${ROOT}/bin/ck8s" init both
```

scripts/local-cluster.sh
```diff
@@ -304,8 +304,27 @@ config() {
...
   if [[ -z "${CK8S_PGP_FP:-}" ]]; then
-    log.fatal "CK8S_PGP_FP is unset"
+    if [[ -f "${env_file}" ]]; then
+      # shellcheck disable=SC1090
+      source "${env_file}"
+      log.info "CK8S_PGP_FP was unset. Reusing temporary GPG environment from ${env_file}"
+      log.info "To persist this in your current terminal, run:"
+      log.info "  source ${env_file}"
+    elif [[ -x "${ROOT}/bin/setup-local-gpg" ]]; then
+      eval "$("${ROOT}/bin/setup-local-gpg")"
```
simonklb: 🚫 No eval!

shafi-elastisys: Moved to source instead of eval.

bin/setup-local-gpg.bash
```diff
@@ -0,0 +1,25 @@
...
+%commit
+%echo done
+EOF
+
+gpg --batch --generate-key "${GNUPGHOME}/gpg-batch" >/dev/null 2>&1
+
+# SOPS needs this fingerprint to know which key to use
+FINGERPRINT=$(gpg --list-secret-keys --keyid-format LONG | grep sec | awk '{print $2}' | cut -d'/' -f2)
+
+export GNUPGHOME="${GNUPGHOME}"
```
simonklb: Already exported earlier in the script.

```suggestion
```

bin/setup-local-gpg.bash
```diff
@@ -0,0 +1,25 @@
...
+export CK8S_PGP_FP="${FINGERPRINT}"
```
simonklb:
```suggestion
export GNUPGHOME="${GNUPGHOME}"
# SOPS needs this fingerprint to know which key to use
export CK8S_PGP_FP="$(gpg --list-secret-keys --keyid-format LONG | grep sec | awk '{print $2}' | cut -d'/' -f2)"
```

scripts/local-cluster.sh
simonklb:
```suggestion
```

bin/setup-local-gpg.bash
shafi-elastisys: This is giving shell check fail as it is not recommended to avoid masking return values.

simonklb: Ah right, my bad! The only problem I have with the current solution is that when you source it you will also get the `$FINGERPRINT` variable in your shell. So in that case, can you do something like this instead?
```
CK8S_PGP_FP=$(gpg --list-secret-keys --keyid-format LONG | grep sec | awk '{print $2}' | cut -d'/' -f2)
export CK8S_PGP_FP
```

shafi-elastisys: Done with latest commit.

scripts/local-cluster.sh
```diff
@@ -541,6 +541,147 @@ delete() {
   fi
 }
 
+status() {
+  local target_cluster="${1:-}"
+
+  local index_file="${CK8S_CONFIG_PATH}/cluster-index.yaml"
```
aarnq: Why is this included here? Can you rebase and sort your prs by merging them into each other based on how they depend so we can have some clarity?

shafi-elastisys: Sorry my bad. I wrongly commited into this branch. Working in many changes in different branches cause this confusion.

scripts/local-cluster.sh
aarnq: @shafi-elastisys you missed the first print.

scripts/local-cluster.sh
shafi-elastisys: Now it's fixed.

bin/setup-local-gpg.bash
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+GNUPGHOME="$(mktemp -d)"
+export GNUPGHOME
```
simonklb, line 5: Hang on, won't this stop working if you reboot your computer?

aarnq: Yes, hence the warnings in local-cluster.sh

simonklb: Alright if this is acceptable then don't mind me. :smile:

